PR #40, opened 2026-07-24
fix(bench): add --child-auth {api,subscription} for review children

### LINKED ISSUE #32: Bench child-auth option: run review children on subscription OAuth (setup-token) instead of an API key
## Problem / Goal

The bench harness authenticates child review sessions exclusively through an `ANTHROPIC_API_KEY` force-injected from `bench/.env`; the isolated claude-home carries no credentials of its own. Every bench run is therefore metered API spend — including the routine functional smoke (~$27), which under the ratcheted measurement policy (#28) is the recurring per-sub-release cost and whose purpose is mechanical pass/fail, not measurement. An existing Claude subscription (Pro/Max) can absorb this traffic at no marginal cost via the officially supported setup-token path.

Goal: an explicit child-auth mode that runs bench children against subscription capacity via the officially supported long-lived OAuth token (`claude setup-token` → `CLAUDE_CODE_OAUTH_TOKEN`), with auth provenance recorded per run and cost fields handled honestly. Scoring/judging stays API-keyed — subscription auth covers only the child review sessions.

## Requirements

1. A `--child-auth {api, subscription}` knob (default `api`), threaded like the existing `--child-model` precedent, recorded in the `run.json` manifest and env fingerprint; ledger rows carry `auth_mode`.
2. Subscription mode suppresses the API-key injection **and strips every env credential source that outranks the OAuth token** in the documented precedence chain — `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_API_KEY`, and the cloud-provider selectors (`CLAUDE_CODE_USE_BEDROCK`/`CLAUDE_CODE_USE_VERTEX`) — since the child inherits the full parent env today and any of them silently defeats the mode. Preflight additionally fails loud if the effective `CLAUDE_CONFIG_DIR` defines an `apiKeyHelper` (reachable via the `BENCH_CLAUDE_HOME` override), which also outranks the token. `CLAUDE_CODE_OAUTH_TOKEN` is required (ambient env or `bench/.env` via the shared dotenv parser) and its absence fails loud at preflight.
3. Token-only by design: no reliance on macOS Keychain or `~/.claude/.credentials.json` surviving the HOME/`CLAUDE_CONFIG_DIR` isolation (Keychain creds happen to survive on macOS; `.credentials.json` is stranded on Linux — implicit auth is a portability trap).
4. `check_prereqs` branches on mode: the child API-key requirement is relaxed in subscription mode; the judge/scorer key requirements are unchanged. The scoring path is untouched and remains API-keyed.
5. Ledger/cost honesty: the ledger row records the recomputed envelope cost unchanged and carries `auth_mode`; aggregates and any runbook/README cost figure must exclude or explicitly label `auth_mode=subscription` rows (official docs say the subscription figure "isn't relevant for billing" and its representation is unspecified — token counts still record either way). This contract is part of the requirement-9 unit-test surface.
6. Usage-limit behavior is documented in the bench README/runbook as observed, not assumed: official docs say a limit hit blocks further `-p` requests until the window resets (which in bench terms would surface as a 45-min watchdog timeout), but no machine-readable error shape is documented — the first subscription smoke confirms the actual failure mode (blocking-until-reset vs immediate error) and the docs record it; either way `--retry-failed` after the reset recovers. Recommended usage: functional smokes and dev iteration on subscription; of-record paired legs stay on API (clean cost accounting, no throttle-induced timeouts). Optional nice-to-have: classify a limit stall as a distinct invalid-reason if a reliable transcript signature exists.
7. Scope note (policy): this mode is for the owner's own subscription capacity — "ordinary, individual usage" per the usage policy, via the setup-token mechanism that is explicitly designed for CI/scripts. It is not a third-party credential-routing surface.
8. #28's runbook gains a "which tiers may run on subscription" line; cross-reference both directions.
9. Stdlib-only per repo rules; unit tests cover env assembly (injection suppressed, keys stripped, token required, manifest/ledger fields) with no live auth in CI.

## Evidence

Harness auth surface today (`bench/runner/invoke.py` unless noted):

```text
276-319 build_env — full parent env inherited; ANTHROPIC_API_KEY loaded from
  bench/.env and force-set (295-297), winning over ambient; HOME→claude-home
  (312), CLAUDE_CONFIG_DIR→claude-home/config (313); no credentials seeded
  into the isolated home (only _seed_trust's minimal .claude.json, 198-229).
  This injection point is the harness's single auth branch point.
232-273 _gh_auth_env — existing precedent for re-pointing a tool's credentials
  across the HOME isolation move (gh); 164-183 _load_dotenv_key — the shared
  dotenv parser (prereqs/build_env/judge) where a token path naturally lands.
run.py:225-229 check_prereqs hard-requires ANTHROPIC_API_KEY in bench/.env.
run.py:801-806 --child-model — the structural precedent for a per-run child
  knob (CLI → invoke_review → manifest, run.py:633-664).
Cost path: costs.py:52-80 parses envelope total_cost_usd/modelUsage; the
  ledger recomputes cost at score time from per-PR raw envelopes
  (score.py:666-699) → bench/experiments.jsonl.
Judge path (must stay API-keyed): score.py:399-416 sets MARTIAN_API_KEY →
  https://api.anthropic.com/v1/ for the vendored OpenAI-client scorer under
  uv (Bearer auth); key from BENCH_JUDGE_API_KEY | ANTHROPIC_API_KEY
  (score.py:110-127).
```

Platform facts (official docs, checked 2026-07-23):

```text
setup-token: 1-year OAuth token, Pro/Max/Team/Enterprise, "designed for CI
  pipelines and scripts" → CLAUDE_CODE_OAUTH_TOKEN
  (code.claude.com/docs/en/iam#generate-a-long-lived-token)
Auth precedence: cloud providers → ANTHROPIC_AUTH_TOKEN → ANTHROPIC_API_KEY →
  apiKeyHelper → CLAUDE_CODE_OAUTH_TOKEN → /login subscription
  (code.claude.com/docs/en/iam#authentication-precedence)
Cost under subscription: total_cost_usd "isn't relevant for billing purposes";
  representation (zero vs API-equivalent) unspecified
  (code.claude.com/docs/en/costs#track-your-costs)
Limits: 5-hour rolling window + weekly caps + Opus-specific cap; -p requests
  block until the reset shown in the message; no documented machine-readable
  error shape (support.claude.com/en/articles/11145838;
  code.claude.com/docs/en/errors)
Usage policy: subscription auth exists "to support ordinary use"; advertised
  limits "assume ordinary, individual usage"; third parties may not route
  requests through Free/Pro/Max credentials
  (code.claude.com/docs/en/legal-and-compliance#usage-policy)
```

Cost motivation (ledger): routine functional smoke ≈ $21–32/run (mean ~$27) — the recurring per-sub-release spend under the ratcheted policy; mini-subset legs ≈ $78–85 (recommended to remain on API per requirement 6).

- One-PR shape: knob + env-assembly branch + preflight branch + manifest/ledger fields + README/runbook docs + unit tests.
- Subscription-mode env assembly mirrors `_gh_auth_env`'s shape: a small `_claude_auth_env(mode)` helper owning injection/strip/require logic at the single branch point.

## Out of scope

- Scoring/judge auth (stays API-keyed; subscription cannot serve the judge path).
- Any review-pipeline behavior change.
- Bedrock/Vertex/cloud-provider auth *modes* (stripping/guarding their selector env vars in subscription mode is in scope per requirement 2; supporting them as auth modes is not).
- Auto-detecting available subscription headroom (no API for it; operator judgment).
- Scheduling or triggering any measurement spend.

## Verification

Tier: **suite-only.** Measurement policy (ratcheted, 2026-07-23): every item ships behind the always-on deterministic suites (pytest + `node --test`, parity and bundle-freshness included); this is bench-side tooling with no review-pipeline behavior surface, so no paired bench. Unit tests (under `bench/tests/`) prove the env-assembly contract: subscription mode suppresses injection, strips API-key vars, requires the token, and stamps `auth_mode` into manifest and ledger; api mode is byte-identical to today's behavior. The live proof is self-demonstrating: the first subscription-mode functional smoke run is the verification (exit-0 plus the #28 checker gates once that checker lands; a manual smoke suffices until then — soft dependency, not a blocker).

leehopper:
Closes #32.

> **Before squashing:** the pre-filled squash body will contain the literal skip-ci token, inherited from commit `10e7d9d`'s message (it explains semantic-release's convention and quotes the token). **Delete that line before confirming**, or the squashed commit carries it onto `main` and GitHub Actions skips the release run — no version bump, no changelog entry. `AGENTS.md` now warns about this; the token itself cannot be edited out of the branch's history without a force-push.

## Why?

Every bench run is metered API spend today, because `build_env` force-injects `ANTHROPIC_API_KEY` from `bench/.env` into the isolated claude-home, which carries no credentials of its own. That includes the routine functional smoke — the recurring per-sub-release cost under the ratcheted measurement policy, and one whose purpose is mechanical pass/fail rather than measurement. An existing Pro/Max subscription can absorb that traffic at no marginal cost through the officially supported `claude setup-token` → `CLAUDE_CODE_OAUTH_TOKEN` path.

This adds an explicit, opt-in child-auth mode for exactly that, with auth provenance recorded per run and cost fields handled honestly. Scoring/judging is untouched and stays API-keyed — subscription capacity covers only the child review sessions.

## What Changed?

**`--child-auth {api,subscription}`, default `api`** (`bench/run.py`) — threaded like the existing `--child-model` precedent: CLI sentinel, resolved default, `run.json` manifest and its `env_fingerprint`, then the ledger row. A resumed run keeps the mode it started with, and a flag contradicting it is refused rather than honoured: one run dir's per-PR costs are summed into a single `auth_mode`-labelled row, so mixing credentials within it would make that row a lie. The mode reaches **both** invocation paths — `--anchor naive` authenticates through the same `build_env` — which is why `child_auth` is recorded even for runs whose `tool`/`child_model` are null.

**Env assembly** (`bench/runner/invoke.py`) — a single `_claude_auth_env(base_env, mode)` helper owns the whole credential decision, mirroring `_gh_auth_env`'s shape at the one existing branch point. `api` is unchanged. `subscription` **removes** every env credential source the documented precedence chain places above the OAuth token — `CLAUDE_CODE_USE_BEDROCK`, `CLAUDE_CODE_USE_VERTEX`, `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_API_KEY` — because the child inherits the operator's full env and any survivor silently defeats the mode. The token comes from `bench/.env` (winning over ambient, the same precedence that file already has for the API key); its total absence raises rather than falling back to the metered key.

**Preflight** (`check_prereqs`) branches on the mode and nothing else. `subscription` drops the metered-key requirement and instead requires the token and refuses an `apiKeyHelper` in the isolated claude home — the one over-ranking source that lives in a settings file and so cannot be stripped from an env. `main()` preflights the mode the run will *actually* spend, and a resume of a run dir with no manifest records the mode it chose, so scoring cannot later default it to `api`. Because dropping the key requirement means a missing judge key would only surface hours later at score time, the runner prints a one-line startup `NOTE:` when a subscription run can see no judge key at all — a warning, not a prerequisite.

**Cost honesty** (`bench/runner/ledger.py`, `score.py`, `report.py`) — the ledger row carries `auth_mode`; `cost_usd`/`tokens_total` are recorded exactly as before, so the envelope figure is never rewritten. `ledger.cost_is_billable` is the single gate consumers read, because Anthropic documents a subscription-served run's `total_cost_usd` as not relevant for billing. The dashboard keeps the figure in the run-ledger table behind a `‡` marker but drops it from the run-cost tile, the release-card legs, and the derived per-golden aggregate; the footnote appears only when such a row exists, so a ledger of API-keyed runs renders byte-identically and the committed `report.html` does not churn. `auth_mode` is deliberately **not** in `REQUIRED_KEYS` — the ledger is append-only and every historical row must stay valid.

`ledger.py` is also the single home of the child-auth vocabulary (`AUTH_MODES` and friends) and of the manifest → `auth_mode` chain (`manifest_auth_mode`), read by the CLI, the env assembly, the scorer and the dashboard alike. It earns that role twice over: `auth_mode` is the field that outlives a run, and the module imports nothing from the package, so every consumer can depend on it without a cycle.

**Docs** — `bench/README.md` gains a *Child auth modes* section: the setup-token flow, the stripping rationale, token-only-by-design rather than leaning on Keychain or a `.credentials.json` that the `HOME` isolation strands on Linux, the guard's own limitations (worktree-scoped and enterprise-managed settings it cannot reach), scoring staying API-keyed, cost honesty, and the usage-policy scope. Usage-limit behaviour is recorded as **unconfirmed** with the documented expectation and the bench-side symptom to look for, for the first live smoke to settle rather than assert now. `bench/MEASUREMENT.md` gains a which-tiers-may-run-on-subscription table; both docs cross-reference each other.

## Verification

Tier: **`suites-only`** — see [`bench/MEASUREMENT.md`](bench/MEASUREMENT.md). Bench-side tooling with no review-pipeline behaviour surface, so the deterministic suites are the gate.

129 new unit tests prove the contract with no live auth. The load-bearing one crosses the subprocess boundary: the fake `claude` records which credential vars it actually *received* (presence only, never values), so "the metered key never reaches the child" is asserted rather than reasoned about. I verified that test is not vacuous by neutering the production strip list and watching it fail; likewise the new `report.py` script-invocation test, verified red with the `sys.path` shim removed.

The live proof is owner-triggered and self-demonstrating: the first subscription-mode functional smoke, best folded into the next per-sub-release smoke (which then costs ~$0 instead of ~$27). It is also what settles the usage-limit paragraph the README marks unconfirmed.

## Additional Notes

Two pre-review adversarial passes found three defects, each now regression-tested: a resume path that preflighted one credential and spent the other for a run dir with no `run.json`; an `apiKeyHelper` scan that missed a home-relative `.claude/` dir; and an unlabelled subscription cost still printed on the release cards.

CodeQL flagged clear-text logging on the prereq output. I reproduced it locally with the CodeQL CLI to get the real dataflow rather than guessing: its credential heuristic matches any `api?key` or `oauth` identifier, so interpolating those names into a printed message — even when only the *name* travels — reads as logging a credential. Fixed at the source: env var names are string literals in user-facing messages, and the helper lookup is named `api_key_helper_files` for the file paths it returns (`file` is the heuristic's own de-classifier for a filename). No sink suppression, no information dropped from the messages, and tests pin both the literal-vs-constant spelling and the naming constraint.

**Review rounds (all six findings fixed, threads answered inline and resolved):** the auth-mode strings and the manifest → mode chain were consolidated into `ledger.py` — I'd rate the duplicated chain above its LOW label, since two implementations could resume a run on one credential while labelling its row with the other, which is precisely the cost honesty this PR adds; `report.py`'s `sys.path` shim is now exercised by a real script run; the table's duplicated `fmt_money` call folded into one; an unused import removed after an AST sweep found no others; and Bugbot's orphan-run-dir finding closed by persisting the mode such a resume spends.

**Merged `main` twice** — the contribution-surface refresh and then the v3.1.3 release. The only conflict was `bench/MEASUREMENT.md`'s tier ladder: `main` added the `Slug` column and its explanatory paragraph while this branch qualified the Cost column as `auth_mode=api` spend. Kept both, with `main`'s table shape verbatim, since `tests/test_contribution_surface.py` parses the `Slug` column positionally; all three slugs still resolve. The release commit also reintroduced the documented `CHANGELOG.md` markdownlint nit, re-fixed here.

**Deferred work, captured in-repo rather than left in a closing issue:** the optional limit-stall invalid-reason classification is now an entry in `bench/MEASUREMENT.md`'s watch ledger, whose stated purpose is facts surviving issue deletion — it is evidence-blocked rather than actionable, so a work-queue issue would sit open with no trigger. The rationale for `--check`/`--score-only` accepting and ignoring the per-run knobs is now a comment at that guard, where a future contributor would otherwise "fix" it.

Not in scope (per the issue): judge/scorer auth, any review-pipeline behaviour, Bedrock/Vertex as auth *modes*, subscription-headroom detection, and triggering any measurement spend.

[bench/report.py output for the same run under both auth modes: with auth_mode=api the Run cost tile reads $27.00 and the ledger row](https://cursor.com/agents/bc-0f973935-105d-4880-a9d2-b34d6316f80a/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Freport_subscription_cost_honesty.png)

- [x] Pipeline tests pass: `python -m pytest tests/ -q` (562 passed, 245 subtests)
- [x] Bench self-tests pass: `python -m pytest bench/tests/ -q` (483 passed, was 354)
- [x] Workflow tests pass: `node --test workflows/test/*.test.js` (263 passed, Node 24.18.0)
- [x] Parity fixtures regenerated with `workflows/test/tools/record_parity.py` if a deterministic transform changed — n/a, no deterministic transform touched
- [x] Docs, skill references, and agent contracts updated if behavior changed

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-0f973935-105d-4880-a9d2-b34d6316f80a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-0f973935-105d-4880-a9d2-b34d6316f80a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


bench/runner/ledger.py
```diff
@@ -8,12 +8,23 @@
...
+run's ``total_cost_usd`` as not relevant for billing purposes, so
+``cost_is_billable`` below is the single gate every consumer (scoring, the
+dashboard) reads before letting a row's cost into a billable figure.
 """
 
 import json
 import os
 
+SUBSCRIPTION_AUTH_MODE = "subscription"
+DEFAULT_AUTH_MODE = "api"
```
leehopper: **💡 [LOW] Auth-mode string constants duplicated across three modules instead of one canonical source**

The child-auth domain concept ("api"/"subscription") is defined independently in three places -- invoke.CHILD_AUTH_MODES, ledger.DEFAULT_AUTH_MODE/SUBSCRIPTION_AUTH_MODE, and run.DEFAULT_CHILD_AUTH -- as bare string literals rather than one shared enum/constant, so nothing but unit tests keeps the three spellings from drifting.

bench/run.py
```diff
@@ -653,9 +728,74 @@ def _resolve_child_model(tool, child_model):
...
+        return None
+    try:
+        manifest = _load_json(path)
+    except (ValueError, OSError):
+        return None
+    if not isinstance(manifest, dict):
+        return None
+    fingerprint = manifest.get("env_fingerprint")
+    fingerprint = fingerprint if isinstance(fingerprint, dict) else {}
+    return manifest.get("child_auth") or fingerprint.get("child_auth") or DEFAULT_CHILD_AUTH
```
leehopper: **💡 [LOW] Manifest-to-auth-mode fallback chain is duplicated (and drift-risky) across run.py and score.py**

bench/run.py's _recorded_child_auth and bench/runner/score.py's _auth_mode both implement the identical three-step fallback -- top-level manifest.child_auth, else env_fingerprint.child_auth, else the api default -- as separate inline expressions with separately-defined default constants.

bench/report.py
```diff
@@ -18,19 +18,36 @@
...
 
+BENCH_DIR = Path(__file__).resolve().parent
+# This module runs as a script (``python3 bench/report.py``), so the repo root is
+# not importable by default; the ledger helpers below are the one definition of
+# which costs are billable and must not be re-derived here.
+REPO_ROOT = BENCH_DIR.parent
+if str(REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(REPO_ROOT))
+
+from bench.runner.ledger import cost_is_billable, row_auth_mode  # noqa: E402
```
leehopper, line 33: **💡 [LOW] New cross-package import + sys.path shim in report.py never exercised via actual script invocation**

This PR adds report.py's first-ever intra-repo import (bench.runner.ledger.cost_is_billable/row_auth_mode) plus a new sys.path.insert(REPO_ROOT) shim needed because report.py is documented as run standalone (python3 bench/report.py). Every existing test imports the module via `from bench import report` after the test file itself has already inserted REPO_ROOT onto sys.path, so the shim's actual job is never exercised. No test or CI step subprocess-invokes `python3 bench/report.py`.

bench/report.py
```diff
@@ -1329,6 +1359,21 @@ def build_table_html(groups, top_anchor, ceiling):
...
+        else:
+            cost_tip = html.escape(
+                f"auth_mode={row_auth_mode(r)} — served by Claude subscription "
+                "capacity, not metered API spend; excluded from every derived "
+                "cost figure."
+            )
+            cost_cell = (
+                f'<td class="num" title="{cost_tip}">'
+                f'{fmt_money(r.get("cost_usd"))} {NON_BILLABLE_MARK}</td>'
+            )
```
leehopper, line 1376: **💡 [LOW] cost_cell branch computes fmt_money(cost_usd) twice instead of once**

The new cost_cell if/else in build_table_html calls fmt_money(r.get("cost_usd")) separately in each branch even though the formatted value is identical in both -- only the wrapping markup differs.

PR #44, opened 2026-07-25
feat(bench): environment purity receipts for wrong-plugin rejection

leehopper:
## Why?

Fixes #23.

A same-named stale plugin install can be stochastically resolved by bench child sessions instead of `--plugin-dir`, silently running the wrong pipeline. Five bench runs were voided by this class before quarantine. Measurement validity (and gates for #22 / #29) depends on failing loud when the child did not run the intended plugin.

This PR also re-triggers release automation after `8c23613` landed on `main` without a Semantic Release run (its squash message carried the skip-ci token). Merging this PR is the intended recovery path — no extra empty commit needed. Do not put that token in the squash body.

## What Changed?

**Plugin (echo metadata only — no review-behavior change):**
- `Headless config:` three-copy contract now includes identity receipt lines:
  - `pipeline_version=<PIPELINE_VERSION> (bundle)`
  - `plugin_root=<absolute resolved path> (resolved)`

**Bench harness:**
- `invoke.py` parses the identity receipt after each child (stdout / envelope `.result` / report md), compares against repo `PIPELINE_VERSION` + `REPO_ROOT`, and checks new `wf_*.json` `scriptPath` values when present.
- On mismatch: loud stderr + `invalid` / `plugin_identity_mismatch` — never scored.
- Smoke checker G4 (`check.py`) prefers echo receipts (`PLUGIN_IDENTITY_STRATEGY = "echo_receipt"`); a complete valid receipt is sufficient without wf records; scriptPath remains defense-in-depth when records exist.
- Shared helpers: `scriptpath_from_record`, `script_path_matches_repo` (used by both invoke and check).
- Tolerant `raw.json` parsing for G4 identity (preamble/stderr-safe).
- `fake_claude` emits identity lines from `--plugin-dir`; modes cover wrong echo, wrong/nested scriptPath, and missing identity lines.

## Additional Notes

- Measurement tier: **smoke** (owner-triggered). Suites-only coverage is in this PR.
- Unit proof: mismatch ⇒ run invalidation (`PluginIdentityGuardTest`).
- Before squash-merge: scrub the pre-filled squash body of any skip-ci token inherited from branch history (there should be none).
- After merge: verify Semantic Release fired and a tag newer than `v3.1.3` exists.

- [x] Pipeline tests pass: `python -m pytest tests/ -q`
- [x] Bench self-tests pass: `python -m pytest bench/tests/ -q`
- [x] Workflow tests pass: `node --test workflows/test/*.test.js` (Node 24; the glob is required)
[...]

Issue #36, opened 2026-07-24
Establish a consistent visual identity across delivered output: brand mark, report formatting contract, and emoji conventions


leehopper:
## Problem / Goal

The tool has no visual identity in anything it delivers, and its formatting convention never reaches the component that renders reports.

**No brand mark where it matters.** Inline PR comments — the highest-volume surface, one per delivered finding — carry no attribution at all. The only brand strings in the product are three plain-text occurrences of the words "Code Gauntlet" (`report-format.md:220`, `:238`, `stages.js:1151`): no mark, no consistent placement, and none on inline comments.

**The report formatting contract never reaches the renderer.** `references/report-format.md` is still read by the *orchestrator* in Phase 8 (`SKILL.md:262`, `phase8-delivery.md:35`, `:60`) and governs PR-comment format. But in v3 the report itself is written by the `report-writer` agent, dispatched from `workflows/src/stages.js:1168-1178` — and neither that agent file nor its dispatch prompt references `report-format.md` or any formatting convention. Report *structure* is constrained by `agents/report-writer.md:28-37`, but the emoji convention, severity section headings, permalink format, and the `Generated by code-gauntlet` / `Reviewed up to:` footer specified in `report-format.md` reach neither the agent nor the prompt. Compounding this, `phase8-delivery.md:60` still instructs the orchestrator to output the report per `report-format.md` — a surface that no longer exists in v3. The deterministic fallback report drops emoji entirely, so a degraded run visually reads as a different product.

**The one convention that survives is stated four times, inconsistently.** The severity-emoji mapping appears in three reference files in three phrasings, plus a fourth hard-coded copy in Python — with no `<!-- Canonical source: … -->` marker, the convention this repo uses elsewhere for deliberate duplication.

Goal: a deliberate, consistent identity across every human-facing surface — a brand mark on delivered comments and reports, a formatting contract the renderer actually receives, and a single canonical source for emoji conventions — without touching any machine-parsed string.

## Requirements

1. A brand mark appears consistently on delivered output: inline PR/MR comments, the summary review-comment header, and the report title. Constraints on the mark: visually distinct from the four severity emoji (which carry semantics, not identity), legible at small sizes, and unambiguous in both light and dark GitHub themes. The specific mark is a design decision for the pickup session; whatever it chooses lands everywhere in one pass rather than accreting per-surface.
2. The deterministic fallback report carries the same identity as the agent-rendered report — a degraded run must not look like a different product.
[...]

PR #45, opened 2026-07-27
fix(detection): make prior-review detection read the signal the pipeline actually writes

### LINKED ISSUE #39: Prior-review detection is broken: reruns on an already-reviewed PR always start fresh
## Problem / Goal

Rerunning a review on a PR that was already reviewed always treats it as a first review — the incremental/resume path never triggers. This is a user-visible functional regression: the "reviewed up to SHA X, only look at what changed since" feature does not fire.

Root cause is a read/write signal mismatch. Phase 1 detects a prior review by looking for a **prose footer** (`Generated by code-gauntlet` / `Reviewed up to: {sha}`) in PR comments and parsing the last-reviewed SHA out of that prose. But the pipeline never mechanically writes that prose: the review summary body is emitted empty, and the templates the model follows to compose a summary omit the footer. The one signal that *is* always written durably — an HTML marker in the review summary — is not what the detector reads. So the detector matches a string that is usually absent, and every rerun looks fresh.

This is distinct from the marker payload-shape skew tracked in #36 (requirement 8). That skew degrades *diffing* (classifying findings as introduced/fixed/pre-existing) once a prior review is found; this issue is that a prior review is never found in the first place. Both live on the same marker surface and should be reconciled together, but they are different defects with different symptoms.

Goal: rerunning a review on a previously-reviewed PR reliably detects the prior review and resumes incrementally, with the detection signal being something the pipeline actually and deterministically writes, and with test coverage so it cannot silently regress again.

## Requirements

1. The prior-review signal the detector reads must be one the pipeline **deterministically writes on every completed review** — not model-composed prose that examples omit and that the empty review body drops. Whatever the durable signal is (marker, footer, or both), the write path and the read path must reference the *same* signal.
2. Detection must actually retrieve the PR/MR comments it inspects — a concrete comment-fetch step, matching where the signal is actually posted (the review summary body, not only inline comments or issue comments), on both GitHub and GitLab.
3. Detection must recover the last-reviewed commit SHA reliably enough to drive the incremental diff (full-vs-short SHA handling; behaviour defined when the PR head has advanced, and when it has not).
4. The legacy `Generated by deep-review` signal continues to be recognized (rename back-compat is currently correct and must stay).
5. Headless and interactive modes both detect correctly; the headless previously-reviewed policies (`CODE_GAUNTLET_REVIEWED_POLICY` — incremental/full/skip) depend on this detection firing, so they are exercised by the fix.
6. **Regression coverage.** A round-trip test proves the signal a review writes is the signal a rerun detects — today there is none, and the only marker test pins the write-side shape without any read/detection test. The detection→resume path gets explicit coverage.
7. Reconcile with #36: the marker's payload shape (`version`, and whether a `findings` array / `title_hash` set is present) is owned by #36 requirement 8. This issue owns *detecting the prior review*; #36 owns *what the marker carries for diffing*. Whichever lands second must leave a single coherent marker that satisfies both detection and diffing — coordinate rather than diverge.

## Evidence

```text
WRITE path — what is durably, mechanically written:
scripts/post_review.py:347-353 build_footer appends ONLY the HTML marker to the
  review summary body:
    <!-- code-gauntlet-findings: {"version":"3.0","findings_count":N,"sha":FULL_SHA} -->
  sha = full `git rev-parse HEAD` (post_review.py:356-358).
Posted in the review SUMMARY body — GitHub POST .../pulls/{pr}/reviews
  (post_review.py:406-419); GitLab MR summary note (:483-494).
  render_comment_body (:315-340, the inline comments) has NO footer.

The prose footer is NOT written mechanically:
  `Generated by code-gauntlet | Reviewed up to: {full_sha}` exists only in the
  report template (report-format.md:238) and is model-composed into review_body.
  The pipeline emits review_body as '' (workflows/src/stages.js:1286), and the
  examples the model follows omit the footer:
    delivery-guide.md:94   review_body = "Found 3 issues: 1 critical, 2 medium."
    phase8-delivery.md:99   review_body = "'''REVIEW_BODY_HERE'''"
```

```text
READ path — what a rerun matches:
phase1-preflight.md:82 (mirrored SKILL.md:60): "Check PR/MR comments for the
  `Generated by code-gauntlet` footer ... Parse `Reviewed up to: {sha}` to find
  the last reviewed commit SHA."
=> matches the PROSE substring and parses the SHA from prose; does NOT parse the
  HTML marker JSON that is the only guaranteed-written signal.
Break #2: that step names no gh/glab fetch, unlike every sibling step
  (e.g. `gh pr view --json` at phase1-preflight.md:32,:35) — the retrieval may
  never execute.
```

```text
Test gap:
No detection/incremental test in tests/ or bench/. tests/test_post_review.py
  (~:373-389) asserts only the write-side marker SHAPE and pins version:"3.0" /
  findings_count — no read/round-trip/detection test exists.
```

Symptom confirmation: reported by the owner — a rerun on an already-reviewed PR consistently starts fresh; the evidence above traces why (the matched signal is normally absent).

## Out of scope

- The marker payload shape and the introduced/fixed/pre-existing diffing logic (owned by #36 requirement 8) — beyond the requirement-7 coordination that the two paths converge on one marker.
- Any change to what findings a review produces.

## Verification

Tier: **smoke.** Every item ships behind the always-on deterministic suites (pytest + `node --test`, parity and bundle-freshness included), which now include the requirement-6 detection round-trip test. Because this is a delivery/detection change with no effect on which findings a review produces, correctness is proven by that round-trip test plus the per-sub-release functional smoke (mechanical pass/fail; asserts nothing about recall or noise) — the release manager's run. Once #28's checker exists the smoke runs through it; until then a manual smoke suffices (soft dependency, not a blocker). No paired benchmark: detection does not touch discovery/verify/challenge output.

Done means: a review followed by a rerun on the same PR detects the prior review and takes the incremental path; the round-trip test (write-signal == read-signal) passes; back-compat with `Generated by deep-review` holds; GitHub and GitLab both covered; and the marker is coherent with #36's payload requirements.

leehopper:
Closes #39.

Rerunning a review on an already-reviewed PR always started fresh. The detector matched a **prose footer** the pipeline never mechanically wrote (`review_body` is hardcoded `''` in `stages.js`), while the one signal always written — the HTML marker — was not what it read. The detection step also named no `gh`/`glab` fetch command, unlike every sibling step.

## Design

`scripts/review_marker.py` becomes the single source of truth: it builds what `post_review.py` writes and parses what `detect_prior_review.py` reads back, so requirement 1's "write path and read path reference the same signal" is structural rather than a convention. Detection is a deterministic script rather than prose — requirement 6 asks for a round-trip test, and prose is precisely what failed here.

**The marker is a versioned wire format on a public surface.** It lives forever in users' PR histories, written by whatever plugin version they ran. The deciding fact: pre-rename `deep-review-findings` markers *also* carry `"version":"3.0"`, so the field has already lied across a plugin generation. The reader is therefore duck-typed and **never branches on `version`** — `sha` is the only field detection requires, unknown keys are preserved, both token generations are recognized, and JSON is extracted by balanced-brace scanning (a regex terminated at `-->` breaks the moment #36's `findings` array contains that sequence). The written payload is unchanged; minting a fourth shape in the wild buys nothing.

**Boundary with #36:** this PR owns *that a machine writes the signal*; #36 owns *what it carries beyond `sha`*. Templates telling a model to hand-type the marker are removed — a model-composed machine signal is the defect class this issue exists to kill — while `build_marker(..., findings=None)` leaves #36's slot wired and empty, so filling it needs no parser, schema, or doc redesign.

The gate runs in **Phase 2 after checkout**, not Phase 1: it compares the last-reviewed commit against the PR head, and before `gh pr checkout` the tree is on whatever branch the session started on, often without the PR's objects fetched.

## Verified on real data

PR #44 and #40 in this repo were genuinely reviewed by code-gauntlet in the past — those markers were written by the shipped `post_review.py`, before any of this work:

```
PR #44 (real PR head): previously_reviewed: true, signal: marker, sha_is_ancestor: true,
                       head_advanced: true, new_commit_count: 1, incremental_safe: true
PR #18 (never reviewed): previously_reviewed: false, errors: []
```

Also proven end to end with no mocks: the real `post_review.py --dry-run` in a real git repo, HEAD advanced two commits, the posted body fed to the real detector CLI. Legacy `deep-review-findings` tokens and legacy `Generated by deep-review` footers both detect (requirement 4); a human writing "Reviewed up to: `<sha>`" does not.

## Review notes

Four adversarial audit rounds ran against this branch (70 confirmed defects fixed). Two results worth a reviewer's attention:

- **Every fix is mutation-pinned.** An audit showed four fixes could be reverted with the whole suite still green — requirement 6 was not actually met. Regression tests were added and re-verified by reverting all nine major fixes across every round: none survive.
- **A deliberate contract change:** a stale prose footer naming a *different* commit gets the real one appended rather than suppressed. Two visible lines is uglier than one, but the alternative advertises a commit the review never examined.

Two security decisions: the `issues/{n}/comments` surface was dropped (nothing ever wrote the signal there, while any read-access user could post a forged marker that outranks the real review by timestamp), and the echoed marker payload is allow-listed and size-bounded because the orchestrator consumes it.

**Residual risk, deliberately not closed:** both remaining surfaces are user-writable, so a forged signal can make a rerun offer an incremental scope. Closing it needs author filtering, which has real false-negative costs (a teammate's review would not match) and is out of scope here — worth its own issue.

**In scope but pre-existing:** `phase2-triage.md` told the model to append to the tracked `.gitignore`, which SKILL.md forbids as "silently dirties the reviewed repo with an undisclosed edit to a user file". Now consistent, with a test pinning it.

## Verification

Tier: **smoke**, per the ratcheted measurement policy. No paired benchmark — detection touches no discovery/verify/challenge output, and `bench/adapter/adapt.py` derives scorer candidates only from `comments[]`/`discussions[]`, so the changed summary body cannot move a benchmark number. The three regenerated adapter fixtures differ only in that body.

- `python -m pytest tests/ -q` — **683 passed** (was 443)
- `python -m pytest bench/tests/ -q` — 506 passed
- `node --test workflows/test/*.test.js` — 263 passed
- `node workflows/build.js && git diff --exit-code workflows/pipeline.js` — clean (no `workflows/src` change, so no bundle churn or parity risk)
- `pre-commit run --all-files` — clean

https://claude.ai/code/session_01TiGDWtfa4kPGWKbcKtRUUQ

---

> [!NOTE]
> **Medium Risk**
> Changes incremental review scope and forge-facing marker parsing (with deliberate security tradeoffs on user-writable PR surfaces); behavior is heavily tested but mis-detection could skip or mis-scope large reviews.
> 
> **Overview**
> Fixes reruns on already-reviewed PRs by making **prior-review detection read the same machine-written signal** `post_review.py` posts, instead of prose the pipeline never appended.
> 
> **`scripts/review_marker.py`** centralizes the wire format: a prose `Generated by code-gauntlet | Reviewed up to: {sha}` line plus the `code-gauntlet-findings` HTML comment (legacy `deep-review-findings` still parsed; readers never branch on `version`). **`post_review.py`** imports `build_footer`, pins the marker SHA via optional payload `sha` (`resolve_marker_sha`), and appends both halves idempotently.
> 
> **`scripts/detect_prior_review.py`** is the new deterministic detector: paginated `gh`/`glab` fetch of **only** surfaces reviews are posted to (GitHub PR reviews; GitLab MR notes — issue comments dropped as forgeable), `select_latest` from `review_marker`, git checks (`sha_resolvable`, `merge-base --is-ancestor`), and JSON output with **`incremental_safe`**; always exits 0 on degradation.
> 
> **Orchestration docs** move the “previously reviewed?” gate to **Phase 2 after checkout**, call the script with base-repo `owner`/`repo`, gate on `incremental_safe`, use file-bounded incremental diffs when chosen, require `sha` in post payloads, and switch artifact ignore to **`.git/info/exclude`** instead of editing tracked `.gitignore`.
> 
> Extensive tests (including real `post_github`/`post_gitlab` dry-run → `detect_signal` round-trip) and bench adapter fixtures updated for the new summary body shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 38220af85f9601a52254bba163785e21125b2d17. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/post_review.py
```diff
@@ -343,21 +358,41 @@ def render_comment_body(finding):
...
+    head = get_head_sha()
+    if not SHA_RE.fullmatch(head):
+        # get_head_sha() yields "unknown" when git fails. Writing that produces a
+        # marker the reader is guaranteed to reject, so the review posts but the
+        # next run cannot detect it. Say so rather than failing silently.
+        warn(
+            f"could not resolve a commit SHA for the review marker (git returned "
+            f"{head!r}); the posted review will not be detectable as a prior review. "
+            f"Set the 'sha' field in the findings JSON to avoid this."
+        )
```
leehopper, line 392: **💡 [LOW] resolve_marker_sha's degraded-fallback path (git rev-parse HEAD failing) is untested**

resolve_marker_sha falls back to get_head_sha() when the payload's sha field is absent or malformed, and additionally warns when that fallback value is itself not SHA-shaped. Every existing test mocks get_head_sha to return a SHA-shaped value, so the branch where the fallback itself fails (returning "unknown") is never exercised.

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: d57996d8d9a004b0ebc0d9efc66862116a12f621

scripts/review_marker.py
```diff
@@ -0,0 +1,409 @@
...
+        # normal case (a body whose final marker is the mechanical one) this
+        # costs exactly one scan. Forward-scanning every occurrence made a body
+        # full of unclosed `<!--code-gauntlet-findings:{` tokens cost
+        # O(tokens x length) — minutes of Phase 1 stall from one hostile comment.
+        matches = list(_MARKER_RE.finditer(text))
+        for match in reversed(matches[-_MAX_MARKER_SCANS:]):
+            start = match.end()
+            if start >= len(text) or text[start] != "{":
+                continue
+            payload, _ = _scan_json_at(text, start)
```
leehopper, line 259: **💡 [LOW] No test exercises the _MAX_MARKER_SCANS DoS-hardening bound in find_marker**

find_marker caps the number of marker candidates it will attempt to parse to the last 32 matches (_MAX_MARKER_SCANS), explicitly to defend against a hostile PR/MR comment body containing many unclosed marker tokens, which the module's own comments say previously caused O(tokens x length) parsing costing minutes of Phase 1 stall. This is new, security-relevant logic on an attacker-controlled input surface (PR/MR comment bodies scanned by detect_prior_review.py), but no test constructs a body with more than 32 marker candidates to verify either the bound is respected or that a valid marker still resolves correctly when many malformed candidates precede or follow it.

scripts/detect_prior_review.py
```diff
@@ -0,0 +1,626 @@
...
+        "sha_resolvable": sha_resolvable,
+        "sha_is_ancestor": is_ancestor,
+        "head_sha": head_sha,
+        "head_advanced": head_advanced,
+        "new_commit_count": git_facts.get("new_commit_count"),
+        "incremental_safe": bool(sha_resolvable and head_advanced),
+        "marker": sanitize_marker(signal.get("marker")),
+        "scanned": scanned,
+        "errors": errors,
+    }
```
leehopper: **💡 [LOW] build_result()'s two return branches hand-duplicate a 15-key schema with no shared constant enforcing parity**

build_result returns one of two separately-written dict literals that must independently agree on the same 15-key output contract. Nothing enforces both branches stay in sync, so a future field addition to one branch and not the other silently produces schema drift.

Issue #50, opened 2026-07-27
Persistence floor: report.md's double-typing and the verify slice-input size, deferred from #38


leehopper:
## Problem / Goal

Issue #38 cut the artifact-writer's re-transcription of already-held content: `findings.json` and `report.md` are now the only two files an agent still transcribes by value (`workflows/src/stages.js`'s `persistPrimaries` calls them "the only genuinely unique content the writer persists"); `post-review.json` and `checkpoint.json` are derived programmatically by the new `scripts/assemble_artifacts.py` from an FNV-1a-checksum-proved plan, with no agent re-typing them at all.

Two more output-preserving levers were identified during that work and deliberately left unlanded, because each carries a real structural cost beyond "just delete a duplicate write":

**(a) `report.md` is still typed twice.** `report-writer` (tools: `Read` only — "Reasoning only, no disk writes") produces the report's prose, and `artifact-writer` (tools: `Write`, `Read`) re-types that same ~6.4KB (measured: 6,406 bytes, 19.5s of generation in the profiled run) as one of its two remaining primaries. Letting `report-writer` write the file itself removes that re-typing, but forks report generation into a segmented path (agent Write, needs its own tool grant and its own write-proof) vs. the existing unsegmented/fallback minimal-report path, and moves a delivered artifact outside the single-writer checksum proof `persistPlan`/`assemble_artifacts.py` currently establishes for `report.md` (it's one of the plan's two `expect[]`-proved primaries today, proved as *artifact-writer's* output specifically).

**(b) The verify-slice INPUT remains unaddressed and un-shrinkable solo.** It's 14,470 bytes and costs 48.4s of `verify-input-writer` generation (unchanged by #38 — this is a different agent, a different direction, and #38 only touched pipeline→disk writes of already-*derived* content). `scripts/verify_findings.py`'s `run_verification()` mutates each finding dict **in place** and returns those same objects; `workflows/src/stages.js`'s verify stage takes the script's `--output` as the complete, authoritative finding set for every downstream stage (validate/filter/challenge). Dropping any field from `--input` to shrink it would make that field permanently absent from `--output` — and hence from the pipeline — for good. Safely shrinking the input requires the JS side to merge the script's per-id deltas back onto the findings it already holds by value after the call returns, which is exactly the delta-echo mechanism issue #25 is building, in the **disk→pipeline echo** direction (the direction #38 didn't touch).

[...]

Issue #49, opened 2026-07-27
Decide whether AGENTS.md (and QODO.md) content should reach the review agents


leehopper:
## Problem / Goal

`references/phase2-triage.md` used to instruct the orchestrator to read `AGENTS.md` and `QODO.md` if present during Phase 2's project-context gathering, but the "Shared Agent Context File" section — the thing that actually determines what a discovery/validate/challenge agent sees — only ever assembled CLAUDE.md/REVIEW.md rules, risk classification, and the diff. So the read, when it happened, was pure discarded work: nothing an agent read from AGENTS.md or QODO.md ever reached a single review agent. The #38 latency work has since removed that now-provably-dead read (output-preserving — deleting a read whose output reaches no agent doesn't change any agent's input), and left a pointer: "whether AGENTS.md content *should* inform review context is a separate, tracked question — this removal is not that decision." This issue is that question.

AGENTS.md is increasingly the cross-tool convention file (the equivalent role CLAUDE.md plays for this tool specifically) — many users would reasonably expect a review pipeline to honor project conventions documented there the same way it honors CLAUDE.md/REVIEW.md. Leaving it unread means agent findings can miss or contradict conventions the maintainers explicitly wrote down for exactly this purpose, in a file this repo itself ships at its own root.

Goal: decide, deliberately, whether AGENTS.md content should be folded into the shared context file alongside CLAUDE.md/REVIEW.md rules — and if so, land it; if not, record why not so the question doesn't resurface as an accidental re-add of the same dead read.

## Requirements

1. Decide whether AGENTS.md content becomes part of the shared context file's project-rules bullet, and if so, its precedence relative to CLAUDE.md/REVIEW.md when they conflict.
2. If adopted, the read (`references/phase2-triage.md`'s 2d step) and the carry-forward (the "Shared Agent Context File" section) move together in the same change — the defect being fixed here is precisely a read whose output never reached the assembly step, so a re-add that repeats that split is not a fix.
3. Decide QODO.md's fate explicitly (include on the same terms as AGENTS.md, exclude, or defer) rather than let it ride passively on whatever AGENTS.md decides. No file named `QODO.md` exists anywhere in this repo today, so this is a forward-looking, lower-priority decision, not a live gap.
4. If adopted, this changes what a discovery/validate/challenge agent can see and therefore what it can find — analogous to issue #47's field-plumbing gap, it is not output-preserving and does not ride issue #38's tier.

## Evidence

```
[...]

Issue #48, opened 2026-07-27
security-reviewer reviewed roughly half the diff


leehopper:
## Problem / Goal

In the profiled run `wf_cef39739-577` (2026-07-24), all 7 discovery agents were dispatched against the same shared context file Phase 2 writes once per run (`code-gauntlet-context-479c1b2b.md`, 95,057 bytes / 2,028 lines — the diff, project rules, and risk classification). Every one of the 7 discovery agents' FIRST `Read` of that file returned a truncated 58,145-character result ending at line 1083, with no truncation notice anywhere in the tool result. Six of the seven agents (bug-detector, cross-file-impact, test-analyzer, conventions-and-intent, type-design-analyzer, code-simplifier) issued one or more follow-up `Read` calls with an `offset` and read on to the file's actual end (line 2027/2028). `security-reviewer` did not: its entire tool-call sequence was one `Read` of the context file, two `Read`s of unrelated source files, and `StructuredOutput` — it never issued a second `Read` against the context file and so never saw roughly the second half of the diff and project context.

security-reviewer also happened to finish fastest of the seven (73.6s wall-clock, 5 tool calls total) and runs on Opus per `registry.js`'s `modelOverride: 'opus'` for the `security` dimension. That speed has already been used as supporting evidence in two other open issues' reasoning about model choice: issue #38's Out-of-scope section and issue #29's premise both read security-reviewer finishing fastest as a signal about Opus's relative speed on comparable work. It was not comparable work — six of the seven agents read roughly double the context security-reviewer did, on the identical input file.

Cost of leaving it alone: any review whose diff spans beyond roughly the first half of the shared context file gets a security pass over only the visible half, with no signal anywhere (report, transcript, or tool result) that anything was cut off. "Fixed" looks like: security-reviewer (and any future discovery agent) reliably reads the whole shared context file before reporting findings, and the model-speed reasoning already recorded in #38/#29 is corrected against equal-context evidence.

## Requirements

1. Every discovery agent contract that instructs reading the shared context file must make explicit that the agent must continue reading (paginate) until it reaches the file's actual end, regardless of how much a single tool call returns — not rely on the agent noticing an unmarked truncation on its own.
2. Whatever satisfies requirement 1 must not depend on the Read tool emitting an explicit truncation notice — the profiled evidence below shows none of the 7 agents' first reads carried one, yet 6 of 7 self-corrected by other means.
[...]

Issue #47, opened 2026-07-27
Discovery findings silently lose `suggestion` and `claude_md_rule`


leehopper:
## Problem / Goal

All 7 discovery agent contracts (`agents/bug-detector.md`, `agents/security-reviewer.md`, `agents/cross-file-impact.md`, `agents/test-analyzer.md`, `agents/conventions-and-intent.md`, `agents/type-design-analyzer.md`, `agents/code-simplifier.md`) instruct every discovery agent to emit a `suggestion` field and a `claude_md_rule` field on every finding — `agents/conventions-and-intent.md:256` even states a convention finding without a cited `claude_md_rule` "will be rejected", and `:258` says the same for `spec_text` on intent findings. `agents/test-analyzer.md` additionally instructs `criticality` and `failure_scenario`. None of these five field names appear anywhere in `workflows/src/registry.js`'s `DIMENSIONS[].schemaExtra` or in `workflows/src/stages.js`'s `FINDING_PROP_TYPES`. `findingItemSchema()` (stages.js:200-206) builds every discovery dispatch's structured-output item schema strictly from `FINDING_PROP_TYPES ∪ {cross_file_refs} ∪ schemaExtra`, and the platform enforces that schema at the StructuredOutput boundary — so every one of these five fields is silently discarded the moment an agent transcribes its findings back, before merge, verify, filter, challenge, or report ever see them.

This is not a documentation nit: `skills/code-gauntlet/references/report-format.md`'s own "Finding Fields Reference" table lists `suggestion` as **required** ("yes"), and its Full Report Template renders `{finding.suggestion}` directly — a promise the emission boundary cannot keep, since the field never survives that far. It is the same class of defect CLAUDE.md and `stages.js`'s own comments name explicitly — the "empty-properties trap" (stages.js:190-194) that once caused the verify executor to silently drop `description` for every finding via an under-declared schema — except here the drop isn't a transcription accident on one run, it is a permanent, by-design gap between what agent contracts instruct and what the schema at the very first hop can carry.

[...]

PR #51, opened 2026-07-27
perf: cut review wall-clock without changing review output (#38)

### LINKED ISSUE #38: Cut review wall-clock without cutting review work: validate profiled overhead, then explore for more
## Problem / Goal

A full measured review run takes ~30 minutes wall clock. Profiling one complete, non-degraded run end to end (PR #34 on this repo — 17 files, 1,882 changed lines, all 8 phases, 20/20 agents succeeded on first attempt, no retries) shows that in this run a large share of *wall-clock* is spent waiting and re-emitting content rather than analyzing code.

Of 1,280 s of machine time (the 1,810 s review run minus 530 s of human answer time at three interactive gates), the serial, output-preserving overhead is concentrated in a few places:

- **242 s — the artifact-writer agent re-typing content it was handed verbatim.** 241.5 of those 242 s are token generation, producing ~100 KB of findings/report/payload/checkpoint that already existed as strings. This one agent is 24.5% of the workflow.
- **113 s of orchestrator model latency gating 28 s of actual shell work** in Phase 2, across 12 strictly sequential Bash calls, at least 8 of which are mutually independent.
- **21 s lost to a failed dispatch** rejected solely because an optional field was present as `null` rather than absent.

Those three serial items alone are ~424 s of the 1,280 s (~33%) in this run, none of which is code analysis. Separately, the discovery stage leaves large parallel capacity idle — four of seven agents finished within 195 s and then waited 228–330 s for the slowest — which is a wall-clock lever of a different kind (overlap, not deletion).

Tool execution is not the lever: across all 20 agents, **2,329 s was model generation vs 24 s tool execution (98.9% / 1.0%)**. Parallel speedup was only 2.38× despite a concurrency cap of 7, because the trailing report + artifact-writer stages (274.5 s, 27.8% of the workflow) run one agent at a time.

Goal: cut wall clock by removing waiting and re-work, with **zero change to what any reviewer agent examines or concludes**. Work that would reduce review depth — or that could shift delivered findings — is explicitly excluded (see Out of scope), and any change that turns out not to be output-preserving leaves this issue.

**This issue is deliberately open-ended on discovery.** The overhead below was found by profiling a single run (N=1); it is a validated *starting point*, not the complete set of speedups. The pickup session runs in two stages: first **validate** the enumerated findings (re-measure, confirm each still reproduces and is output-preserving), then run an **aggressive multi-agent exploration** for further speed enhancements across the whole pipeline before committing to an implementation set. The known findings anchor the work; they do not bound it.

## Requirements

1. **A hard line between overhead and analysis.** Every change under this issue must be output-preserving: the same agents examine the same context and produce the same delivered findings, in the same order into any order-sensitive stage. Anything that reduces investigation depth or could alter delivered findings belongs elsewhere (see Out of scope), or escalates per the Verification clause.
2. **Persisting already-held content must not incur model-generation cost or scale with content size**, and the persisted bytes must be identical to the strings the pipeline already holds. (This is the single largest line item — ~290 s including the verify slice-input writer.)
3. **Stages overlap where correctness allows.** Where downstream work on an item is valid before all sibling items finish, it must not be forced to wait on a full barrier. Barriers that a stage genuinely needs — in particular any cross-item merge/dedup, which is order-sensitive — must be preserved.
4. **Phase 2's mechanical prep must not incur one model turn per independent shell call.** The measured ratio in this run was 113 s of model latency for 28 s of shell; that ratio is the metric to move.
5. **An optional field supplied as `null` rather than omitted must not cause dispatch rejection or a round trip** at the args waist.
6. **A repeatable profiling method** — the way this profile was produced (per-phase spans from transcript timestamps; per-agent durations, concurrency, and critical path from the workflow record; model-vs-tool time split) is written down and re-runnable, so latency claims are measured, not asserted, and regressions are detectable.
7. **Validate-then-explore.** The pickup session first re-validates each enumerated finding against a fresh profile (still reproduces, magnitude confirmed, output-preserving as claimed), correcting or dropping any that don't hold. It then conducts an **aggressive multi-agent exploration** for additional speed enhancements — spanning the workflow stages, the orchestrator's Phase-1/2/8 mechanical steps, the args-waist and dispatch path, per-agent tool-call and context-read patterns, model/effort allocation, and any other surface the profile method exposes. New candidates are triaged by the same overhead-vs-analysis line (requirement 1): output-preserving wins are in scope; anything that could shift delivered findings is filed separately. The exploration's output is a ranked candidate list with measured or estimated savings; the implementation set is chosen from validated findings plus exploration results, not fixed in advance by this issue.
8. Changes must preserve the existing degradation contracts: never-drop/never-fabricate, the write-proof gates, and the gap-reporting paths that keep a degraded run honest.

## Evidence

Run `wf_cef39739-577`: `durationMs=988242`, `agentCount=20`, all `state:done`, `attempt:1`, `totalTokens=1114057`. Session wall clock 41m 06s; review run 30m 10s (t=15.5→1825.7 s); human answer time at 3 gates 530.3 s; machine time 1279.9 s.

```text
Stage profile — spans, share of the 988.2 s workflow:
  summarize        28.9 s   2.9%   1 agent
  discover        423.2 s  42.8%   7 agents, concurrency 7
  verify           92.8 s   9.4%   2 agents, SERIAL
  validate         99.6 s  10.1%   1 agent
  challenge        64.9 s   6.6%   7 agents, concurrency 7
  report           30.3 s   3.1%   1 agent
  artifact-writer 244.2 s  24.7%   1 agent (span; 242.0 s compute + 2.2 s dispatch)
Sum of agent durations 2355.3 s vs 988.2 s wall = 2.38x speedup.
Model generation vs tool execution, all agents: 2329.2 s / 24.4 s (98.9% / 1.0%).
(Stage-table figures are spans; critical-path figures below are compute.)
```

```text
Largest overhead — artifact-writer regenerates handed-over content:
  242.0 s compute, 241.5 s of it token generation across 4 Write calls
  (64.7 / 19.5 / 59.9 / 93.1 s) emitting 22.7 KB findings, report.md,
  23.3 KB post-review payload, 36.0 KB checkpoint.
  verify-input-writer-0 adds 48 s to emit a 14.5 KB slice file.
  Together ~290 s of model time re-typing strings the pipeline already held.

Discovery barrier idle (parallel-capacity lever, not serial time):
  slowest agent 418.9 s x 7 slots = 2932.3 agent-s capacity; 1591.9 agent-s
  used => 1340.4 agent-s idle (45.7%). Four agents done by t=194.8 s, then
  idle 228-330 s each. Stage median 136.7 s. The two stragglers gate the stage:
    conventions-and-intent 418.9 s (3.06x median, 40 tool calls)
    bug-detector           412.5 s (3.02x median, 36 tool calls)

Phase 2 orchestrator overhead (serial):
  143.5 s span ~= 112.7 s model latency + 27.7 s shell + ~3 s inter-call gaps,
  over 12 sequential Bash calls each <7 s (SHA resolve, stale-artifact truncate,
  diff write, changed-files JSON, numstat, authorship grep, test discovery,
  ls-files, nonce, 94,784-char context assembly). >=8 are independent.
  Two calls failed on bad flags (`ls --icons`, `gh pr diff --stat`).

Failed dispatch on an optional field (serial):
  {"ok":false,"error":"invalid args: reviewConfig must be an object
   (the parseReviewMd output shape) when present","phaseReached":"args"}
  Only difference between rejected and accepted payloads: presence of
  `reviewConfig: null` (1785 B vs 1763 B). Cost 21.3 s + a model round trip.

Critical path (988.2 s = 945.4 s agent compute + 38.0 s dispatch latency + 4.8 s gaps):
  summarize 23.7 -> conventions-and-intent 418.9 -> verify-input-writer 48.4
  -> verify-slice 36.5 -> validate-batch 94.3 -> challenge-2 53.7
  -> report-writer 27.9 -> artifact-writer 242.0
  Per-hop dispatch latency 2.3-14.0 s across 9 hops.
```

```text
Measured non-finding — the Phase 3 wait was NOT a latency cost in this run:
  Zero sleep/Read poll iterations executed. The orchestrator issued
  ScheduleWakeup{delaySeconds:1200} and ended its turn; the workflow-completion
  notification arrived ~10 ms after the workflow ended. The poll loop #26
  replaces did not execute at all here.
```

All figures are from a single profiled run (N=1); they establish where the time went in one representative review, not a distribution. Requirement 6 exists so any claimed reduction is re-measured rather than assumed.

- Persist artifacts from the main session off the `Workflow` return (the repo's existing `python3 -c` JSON-writing idiom), or hand a pinned executor a path rather than content.
- Overlap discovery with downstream work *after* the merge/dedup barrier — e.g. begin verify slicing as soon as the merged set for a region is final — without reordering inputs into merge/dedup.
- Collapse Phase 2's independent shell steps into one or two composite calls.
- Merge adjacent single-agent stages to reduce per-hop dispatch latency (~38 s across 9 hops).
- For the requirement-7 exploration: fan out parallel agents each owning one surface (a workflow stage, the orchestrator prep, the dispatch/args path, per-agent tool-call/context-read patterns, model/effort allocation), each returning measured or estimated candidate savings with an output-preserving vs quality-risk classification; dedupe and rank before choosing the implementation set. Profiling additional runs (beyond the single N=1 trace) would strengthen any distribution-level claim.

## Out of scope — deliberately excluded

Each of these is a real lever but changes or risks review output, so none rides this issue:

- **Capping discovery tool calls or effort.** The two straggler agents (419 s / 40 calls, 413 s / 36 calls) gate the pipeline and are exactly the agents that produced the delivered bugs. Truncating them trades findings for seconds.
- **Model mix.** ~~`security-reviewer` on `claude-opus-4-8` finished in 73.6 s / 5 tool calls while sonnet peers on the same context ran 92.6–418.9 s.~~ **Corrected 2026-07-27 per #48: it was not the same context.** `security-reviewer` made a single `Read` of the 95,057-byte / 2,028-line shared context file, got back 58,145 characters ending at line 1083 with no truncation notice, and never issued a follow-up `Read` — so it reviewed roughly the first half of the diff and project rules. The other six discovery agents all paginated to the file's actual end on the identical input, which is most of why they took longer. The 73.6 s figure is therefore not an Opus-vs-Sonnet speed signal and must not be used as one; any such comparison has to be re-derived on equal-context runs after #48's fix. The underlying point stands unchanged: a model-policy change is owned by #29 and gated on paired measurement.
- **Delivery-yield fix.** One of seven surviving findings was discarded at post time (`Skipping finding ... line not found in diff`), wasting ~42 s of challenge compute. Pre-filtering findings against diff line ranges before challenge would recover it — but it changes what is delivered, so it needs its own issue and tier.
- **Collapsing the three interactive gates** (530 s of human wall clock, 29.3% of the run) — owned by #35.

**Interaction with #25:** requirement 2 may remove or repurpose the artifact-writer agent, whose write-proof echo #25 hardens (including its requirement-5 assessment of generalizing content proof to `writeArtifacts`, stages.js:1207–1214). Whichever lands second adapts; the write-proof guarantee must survive either way — persisting without an agent must still prove what was written.

**Re #26:** in this run the poll loop #26 replaces did not execute — the orchestrator took the event-driven route (10 ms overhead). #26 is already scoped as a determinism/reliability fix and remains valid because that route is not guaranteed; this profile simply provides no latency argument for it.

## Verification

Tier: **smoke.** Every item ships behind the always-on deterministic suites (pytest + `node --test`, parity and bundle-freshness included). Because requirement 1 constrains this work to output-preserving changes, correctness is argued by unit tests and — for requirement 2 specifically — an in-run comparison that the persisted bytes equal the strings handed in (a within-run identity check; cross-run artifact identity is not achievable, since the reviewer agents are non-deterministic). The per-sub-release functional smoke (mechanical pass/fail: pipeline completes, schema-valid findings, no verify/writer degrade, intended plugin ran; asserts nothing about recall or noise) is the regression gate — the release manager's run. Once #28's checker exists the smoke runs through it; until then a manual smoke suffices (soft dependency, not a blocker).

**Escalation:** any change that reorders inputs into an order-sensitive stage (notably merge/dedup — CLAUDE.md relies on stable sort, and #25's evidence ties dedup ordering to a measured recall swing) or otherwise alters delivered findings is **not** output-preserving; it becomes escalation-eligible for the owner-triggered paired mini-subset run and leaves this issue's smoke tier. Smoke cannot detect a recall/noise shift, so the burden is on the pickup session to show output-equivalence per change.

Done means: the enumerated findings are validated (or corrected/dropped) against a fresh profile and the multi-agent exploration (requirement 7) has produced a ranked candidate list; the implementation set chosen from that combined pool has landed; a re-profile using the requirement-6 method shows the targeted reductions; requirement-2 persistence is proven byte-identical to the held content within a run; the deferred/quality-affecting levers are filed separately; and the suites plus one functional smoke are green.

leehopper:
Closes #38.

## What this is

Issue #38 asked to cut review wall-clock **without changing what the review examines or concludes**. The pickup ran in two stages as the issue's requirement 7 specifies: validate the enumerated findings against a fresh profile, then explore the whole pipeline for more. 13 agents did the validate-and-explore pass; three adversarial review rounds then ran over the resulting diff.

## Validation: four of the issue's own numbers were wrong

Corrections posted to #38. Summary:

| Issue said | Measured |
|---|---|
| discover 423.2 s / 42.8 %, median 136.7 s | 420.7 s / 42.6 %, median **162.8 s** (136.7 s is code-simplifier's own duration, 3rd of 7) |
| Phase 2: 12 Bash calls, 113 s model / 28 s shell | **13** calls, **99.9 s / 31.1 s** — and both flag failures are in **Phase 1** |
| dispatch 38.0 s / 9 hops / 2.3–14.0 s | **33.9 s / 8 hops / 2.3–11.2 s** (38.0 folded in a flagged unaccounted tail; the 14.0 s hop isn't on the critical path) |
| artifact-writer ~290 s | **confirmed exactly** |

## The persistence finding, verified against the artifacts on disk

Of the 88,389 bytes the artifact-writer generated, only ~30 KB was unique:

- `post-review.json`'s findings array is **canonically byte-identical** to `findings.json` (same 7 ids, same order)
- `checkpoint.phases.challenge.findings` (20,385 B) is the **alias-stripped twin** of the same array
- `checkpoint.phases.filter` (12,879 B) cached a **pure, agent-free function**
- the genuine residual is **383 B**

So ~66 % of that agent's output was re-transcription of one 21.4 KB array. The writer now emits only unique content; a pinned executor runs the new `scripts/assemble_artifacts.py` to derive the rest.

## The write-proof got stronger

Previously the writer echoed back the paths it claimed to write — which `stages.js`'s own comment admits is "a consistency/liveness check, not proof-of-write". Now `assemble_artifacts.py` recomputes an FNV-1a checksum over what is actually on disk and reports it; the pipeline compares against what it computed over the exact string it handed over. Both primaries **and both derived documents** are covered, which makes requirement 2's byte-identity claim provable for the payload users actually receive. This serves #25's requirement 5 (generalising content proof beyond verify slices) — noted on that issue.

To be precise about what it is not: the plan is on disk, so a Byzantine executor could echo any value it names. This is a check against a stale, confused or hallucinating executor and against serializer divergence — the same bound `trustSlice` already documents. The code says so where a maintainer will read it.

## Requirement 2, honestly

Its literal wording — "must not incur model-generation cost or scale with content size" — **is not achievable**, and the reason is physical rather than a design failure: the sandbox has no disk, so every persisted byte must be emitted as some agent's tool-call argument at least once. The floor is one generation pass per unique byte, not zero. This change drives the multiplicity to 1×. #50 tracks the remaining floor.

## Review rounds found 26 verified defects, 5 of them regressions from an earlier fix round

Two crossed the issue's own hard line and would have changed delivered findings:

1. **`changedLines` could silently read 0.** First via `git apply --numstat | awk` (git-apply refuses valid-but-unappliable patches and exits with empty stdout, so awk prints a plausible `0`); then, after the fix, via a hand-rolled scanner that decided "header" by line prefix — a removed line whose content starts `-- ` appears in the patch as `--- `. Same failure class, relocated. Now uses hunk-state tracking, as git does. This feeds the light-scope gate, so an undercount turns a 7-agent review into a 2-agent one.
2. **The persist plan had no content proof** — and it is the sole authority for which finding ids reach the post-review artifact.

Also fixed: the empty-report guard was defeated on resume (twice — the second time because the union ignored the `unverified` bucket); a lone UTF-16 surrogate crashed the assembler leaving a zero-byte artifact; `writeArtifacts` lost its documented try/catch; `persist: null` hard-rejected a run; `checkpoints: null` emitted a false degradation gap.

## Verification

Tier: **smoke**, per the ratcheted measurement policy.

- `python -m pytest tests/ -q` → **763 passed** (was 688)
- `node --test workflows/test/*.test.js` → **379 pass, 0 fail** (was 263)
- `python -m pytest bench/tests/ -q` → **522 passed**
- bundle freshness and parity fixtures included

An in-run byte-identity check (requirement 2's verification clause) asserts the derived artifacts equal the strings the pipeline holds. The headless functional smoke has now run (see **Live smoke** below) — but a comparative re-profile via `bench/profile_run.py` against `wf_cef39739-577` has **not**. So every latency figure above (the ~84.2 s requirement-2 estimate, "drives the multiplicity to 1×") remains a **projection** applied to the pre-existing baseline profile, not a measured delta of a run on this PR's code. Settling it needs a profile of a post-change run compared per-stage against the baseline — not by total wall clock, since different PRs have different diffs and agent behavior and only per-stage/per-item comparisons are meaningful.

## Live smoke

Run `smoke-20260727-205454-f99d948`, 3 PRs (grafana/grafana#80329, calcom/cal.com#22345, ai-code-review-evaluation/discourse-graphite#6), commit `f99d948` — this PR's tip *before* the follow-up commit below — 2026-07-27. First time this boundary has been measured with a content proof.

**Transcription result: 3 of 3 diverged.** The artifact-writer never transcribed `findings.json` faithfully:

- grafana#80329 — 16 chars off (tolerated, artifacts persisted)
- cal.com#22345 — 8 chars off (tolerated, artifacts persisted)
- discourse-graphite#6 — invalid JSON (`Expecting ',' delimiter: line 99 column 182`); `assemble_artifacts.py` structurally refused to derive, and the run lost `post-review.json`/`checkpoint-all.json`

Not new — the pre-existing by-value writer had the same failure mode with nothing checking it; only the detection is new. Full evidence posted on #50, #25, #28, and #47.

**Checker verdict.** All 3 PRs reached invocation status `ok`; the mechanical checker (`bench/runner/check.py`) returned `ok: false` — 5 `origin=unknown` findings plus a writer partial-artifacts degrade on discourse-graphite#6 (a verify-slice receipt-nonce-mismatch degrade, unrelated to this PR's change), plus 2 false-positive G3 hits on grafana/cal.com from a checker defect (raw-text scan matching bundled pipeline-source literals, posted on #28). 21 comments delivered across the 3 PRs; plugin identity clean on all three.

**Follow-up.** One commit, `3849e88` (`fix(persist): unwrap JSON-wrapped reports and retry a refused derived persist once`), landed both fixes this smoke prompted: the intermittent JSON-wrapped-report bug (long-standing, ~15/25 dated runs) and one deterministic retry on a structural assemble refusal. It does not fix the byte-level content divergence grafana/cal.com show (that's tolerated, not corrected) — that stays open as #50's durable-fix work.

## Filed separately (out of scope — each would change delivered findings)

- **#47** — every finding silently loses `suggestion` and `claude_md_rule`: 7/7 agent contracts instruct them, 0 declared in the schema
- **#48** — security-reviewer reviewed ~53 % of the diff: its one Read of the 95,057-byte context file returned 58,145 chars and it never paginated, where 6 of 7 peers did. This also undercuts #29's model-mix premise
- **#49** — should AGENTS.md/QODO.md reach review agents (read and discarded today; this PR removes the dead read)
- **#50** — the remaining persistence floor

https://claude.ai/code/session_018Wf8iDU1zEJPUuEKDhsCCw

---

> [!NOTE]
> **Medium Risk**
> Changes artifact persistence, delivery projection via plan id lists, and Phase 2 `changedLines`/scope gating—bugs could drop artifacts, mis-rank PR comments, or run the wrong agent scope; mitigated by extensive tests, legacy fallback when derivability fails, and explicit degradation gaps.
> 
> **Overview**
> Targets issue **#38** wall-clock without changing review conclusions by shrinking what the **artifact-writer** must emit and by reducing orchestrator round trips.
> 
> **Derived persistence (D3).** When `args.persist` is set, the writer persists only **findings**, **report**, and a **persist-plan** JSON; a pinned **executor** runs new **`scripts/assemble_artifacts.py`** to build `post-review.json` and the checkpoint from on-disk `findings.json` plus the plan. The plan carries **`planChecksum`** and per-file **`chars`/`checksum`** (`fnv1a32` over UTF-16 code units, parity-tested with JS). Primary checksum mismatches are tolerated with gaps; plan/structural failures hard-fail with **one retry** on derived persist (no legacy by-value fallback on that path). **`writeArtifacts`** stays non-throwing; report JSON wrappers are unwrapped before persist.
> 
> **Phase 1/2 latency.** **SKILL.md** replaces many separate Bash calls with **Phase 1** and **Phase 2 Composite A/B** scripts (checkout → prior-review → conditional stale truncate; files → diff → hunk-based **`changedLines`** scan). Docs stress **`git apply --numstat`** and bare `---`/`+++` prefix tests were wrong for scope gates.
> 
> **Observability.** Adds **`bench/profile_run.py`** and **`bench/PROFILING.md`** to profile recorded workflow runs (stages, critical path, write-byte accounting).
> 
> **Agent/contracts.** Updates **artifact-writer**, **executor**, and **report-writer** prompts; documents persistence in **CLAUDE.md**; drops unused **AGENTS.md/QODO.md** reads from triage guidance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4161653cd08cf9de131418811cf1cd99bc1685ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


workflows/pipeline.js
```diff
@@ -1847,17 +1850,134 @@ const NONCE_RE = /^[A-Za-z0-9._-]+$/;
...
+    const delivery = { ...out.delivery };
+    if (delivery.prIdentity === null) { delete delivery.prIdentity; dropped.push('delivery.prIdentity'); }
+    if (delivery.tier === null) { delete delivery.tier; dropped.push('delivery.tier'); }
+    out.delivery = delivery;
+  }
+  return { args: out, dropped };
+}
+
+// The value-only form, kept for callers that do not surface gaps.
+function stripNullOptionals(args) {
```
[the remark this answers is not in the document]
leehopper: Fixed in a515d7c: removed the unused `stripNullOptionals` wrapper from workflows/src/args.js (rebuilt workflows/pipeline.js), and repointed its dedicated test cases in workflows/test/args.test.js at `stripNullOptionalsReport(...).args` so the same behavior stays covered.

scripts/assemble_artifacts.py
```diff
@@ -0,0 +1,743 @@
...
+        # mkstemp creates 0600; os.replace would carry that onto the artifact, which
+        # a later CI step running as another user could no longer read. Restore the
+        # mode a plain open() would have produced.
+        umask = os.umask(0)
+        os.umask(umask)
+        os.chmod(tmp, 0o666 & ~umask)
+    except BaseException:
+        try:
+            os.unlink(tmp)
+        except OSError:
```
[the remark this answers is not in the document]
leehopper: Fixed in a515d7c: added an explanatory comment to the swallowed OSError in the pre-write cleanup path.

scripts/assemble_artifacts.py
```diff
@@ -0,0 +1,743 @@
...
+            pass
+        raise
+    try:
+        os.replace(tmp, path)
+    except BaseException:
+        try:
+            os.unlink(tmp)
+        except OSError:
```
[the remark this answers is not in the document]
leehopper: Fixed in a515d7c: added an explanatory comment to the swallowed OSError in the os.replace cleanup path.

bench/profile_run.py
```diff
@@ -0,0 +1,990 @@
...
+        # Default: most recently completed code-gauntlet run, by the record's own
+        # 'timestamp' field (falls back to file mtime if a record is unparseable).
+        def sort_key(p):
+            try:
+                d = json.loads(p.read_text(encoding="utf-8"))
+                if d.get("workflowName") == "code-gauntlet" and d.get("status") == "completed":
+                    ts = d.get("timestamp")
+                    if ts:
+                        return (1, iso_to_ms(ts))
+            except (json.JSONDecodeError, OSError, ValueError):
```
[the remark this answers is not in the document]
leehopper: Fixed in a515d7c: added an explanatory comment noting the unreadable/invalid record falls through to the mtime-based sort key.

workflows/pipeline.js
```diff
@@ -3172,40 +3346,255 @@ function writerEchoCoversPaths(echoed, paths) {
...
+  const partial = (reason, extraGaps) => ({
     artifactPaths: { findings: null, report: null, postReview: null, checkpoints: null },
-    gaps: [`writeArtifacts: ${reason} — artifacts not persisted (partial-artifacts)`],
+    gaps: (extraGaps || []).concat([`writeArtifacts: ${reason} — artifacts not persisted (partial-artifacts)`]),
     partial: true,
   });
+
+  try {
+    const c = ctx || defaultCtx();
+    const inp = typeof input === 'string' ? JSON.parse(input) : (input || {});
```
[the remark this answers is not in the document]
leehopper: Leaving this one as-is: `typeof input === 'string' ? JSON.parse(input) : (input || {})` is the same defensive args-normalization idiom used at the top of every stage function in workflows/src/stages.js (writeArtifacts is just one of ~8 identical instances). It's flagged here only because writeArtifacts currently has a single in-repo caller that always passes an object literal, so the analyzer can prove `input` is truthy for that call site — but the function is a public stage entry point dispatched by the workflow runtime, where args can arrive as a JSON string. Stripping the `|| {}` here would make this one stage inconsistent with its siblings without fixing a real bug.

workflows/src/stages.js
```diff
@@ -1315,6 +1666,328 @@ function writeArtifactsPrompt(inp, paths) {
...
+    if (Array.isArray(node)) {
+      for (let i = node.length - 1; i >= 0; i -= 1) stack.push([node[i], `${where}[${i}]`]);
+      continue;
+    }
+    if (node && typeof node === 'object') {
+      for (const [k, v] of Object.entries(node)) stack.push([v, `${where}.${k}`]);
+    }
+  }
+  return null;
+}
```
leehopper, line 1875: **💡 [LOW] firstUnsafeNumber does not actually return the FIRST unsafe number for object-keyed values**

firstUnsafeNumber is documented as returning the path of the first number the Python twin could not spell identically and its array branch deliberately pushes indices in reverse (length-1 downto 0) so the LIFO stack pops them in forward order, preserving array-index ordering. The object branch does not apply the same reversal: Object.entries(node) is pushed in forward key-insertion order, so the LAST key ends up on top of the stack and is visited (and returned) FIRST. When an object literal has more than one sibling field that is independently unsafe, the reported path names the wrong field.

bench/profile_run.py
```diff
@@ -0,0 +1,990 @@
...
+
+def build_critical_path(agents, workflow_start_ms, workflow_end_ms):
+    """Longest dependency chain through the (mostly linear) stage graph.
+
+    Within a fan-out stage the critical member is the one that finishes last
+    (its completion gates the next stage's dispatch). merge/filter transform
+    phases are inserted as zero-agent "gap" hops between discover->verify and
+    validate-batch->challenge.
+    """
+    by_stage = {}
```
leehopper: **💡 [LOW] Stage-grouping loop duplicated identically three times in the new profiler**

bench/profile_run.py builds a dict of agents keyed by stage using the identical three-line pattern in build_stage_profile, build_capacity_accounting, and build_critical_path.

bench/profile_run.py
```diff
@@ -0,0 +1,990 @@
...
+        # Phase 2 Bash-call accounting: sequential Bash tool_use/tool_result pairs between
+        # phase2_start_ms and launch_ts.
+        bash_calls = []
+        prev_result_ts = phase2_start_ms
+        for e in events:
+            if e["kind"] != "tool_use" or e["name"] != "Bash" or not (phase2_start_ms <= e["ts_ms"] <= launch_ts):
+                continue
+            tr = next((r for r in events if r["kind"] == "tool_result" and r.get("tool_use_id") == e.get("id")), None)
+            model_latency_ms = max(e["ts_ms"] - prev_result_ts, 0.0)
+            shell_ms = (tr["ts_ms"] - e["ts_ms"]) if tr else None
```
leehopper, line 569: **💡 [LOW] tool_use -> tool_result lookup pattern repeated three times in analyze_orchestrator_phases**

The same linear-scan lookup for matching a tool_result to a tool_use id is written out separately at three call sites in analyze_orchestrator_phases.

leehopper (commented):
Code Gauntlet automated review (v3 pipeline). 4 high-confidence findings survived full pipeline verification and blind challenge (1 bug, 1 doc-accuracy, 2 simplification suggestions); all severity low. See inline comments for details.

---
Generated by code-gauntlet | Reviewed up to: a515d7c3470b2381f9501498ac4ff03e5aa55336

workflows/src/stages.js
leehopper: Fixed in 1d52fb3: the object branch now pushes entries in reverse insertion order (mirroring the array branch's reverse-index push), so the LIFO stack pops the first-inserted key first. Added a regression test.

bench/profile_run.py
leehopper: Fixed in 1d52fb3: extracted the shared `_group_by_stage` helper and used it in all three call sites.

bench/profile_run.py
leehopper: Fixed in 1d52fb3: extracted a `_find_tool_result` helper and used it in all three lookup sites.

PR #56, opened 2026-07-28
fix(bench): decide G3 writer degrades from parsed gaps, not raw bytes

### LINKED ISSUE #52: G3 functional-smoke gate false-flags writer degrades on clean runs (raw-text scan matches pipeline.js's own source literals)
## Problem / Goal

`bench/runner/check.py`'s G3 gate decides writer-degradation by regex-matching the **raw text** of every collected artifact before it ever tries anything structured. `_scan_degrade_text` (`bench/runner/check.py:302-327`) does:

```python
matched = bool(_DEGRADE_RE.search(text))          # line 316 — raw text, always runs first
if not matched and path.suffix == ".json":        # line 317 — only reached if the raw scan missed
    ...
    matched = any(_DEGRADE_RE.search(g) for g in _gaps_strings(data))
```

`_DEGRADE_RE` is `r"(no write proof|partial-artifacts|partial.artifacts)"` (case-insensitive). The scanned artifact set (`_DEGRADE_SCAN_PATTERNS`, `check.py:60-61`) includes `workflows/wf_*.json`, and that collected record embeds the **entire pipeline bundle** in its `script` field. The bundle's own source contains those literals as normal gap-text strings it emits at runtime (`materializeVerifySlices` writes `"...(no write proof)"`, `writeArtifacts` writes `"partial-artifacts"`). So the raw-text scan matches on **every** run that collects a `wf_*.json` record, regardless of whether that run's `result.gaps` actually contains a degrade entry — the structured check at line 317 never gets a chance to run once the raw scan already matched.

Cost of leaving this alone: most currently-open work-queue issues name the per-sub-release functional smoke as their verification gate (see Requirements section of issues #22, #24, #25, #26, #27, #47, #50). Until this is fixed, "the smoke passed" carries near-zero information about writer-degradation specifically — as demonstrated below, 2 of the 3 PRs in today's smoke were false-flagged by G3 while their actual `gaps` were clean.

Goal: G3 determines writer-degrade status from each artifact's **parsed structure** (`result.gaps` / `gaps`) wherever a structured representation exists, and falls back to raw-text scanning only for carriers that have none.

**Important:** G3 has a second, independent half — the `origin=unknown` scan over `code-gauntlet-findings-*.json` (`check.py:~497`) — that is sound and caught a real failure in today's same smoke run (discourse-graphite#6, 5 findings). A fix here must not touch or weaken that signal.

This needs its own issue rather than a comment on #28: #28 (which codified the measurement policy this checker implements) is **closed**, and a closed issue is not picked up as work. An evidence comment describing this same defect was already posted on #28 today for the record, but the fix itself belongs here.

## Requirements

1. For an artifact with a structured JSON representation that carries a `gaps` array (or equivalent), G3 must determine writer-degrade status from that parsed structure — not from a raw-text regex scan of the file's bytes.
2. The `origin=unknown` detection (`check.py:~497`, scanning `code-gauntlet-findings-*.json`) is unmodified by this fix.
3. Raw-text scanning may remain the mechanism only for carriers that have no structured `gaps` representation to parse; the fix documents explicitly which carriers those are (e.g., `code-gauntlet-report-*.md`).
4. A new regression test asserts: a `workflows/wf_*.json` record whose `script` field contains the literal degrade substrings (as `pipeline.js`'s own source does — this is exactly today's false-positive shape) but whose `result.gaps` is clean, does **not** fail G3.
5. A new regression test asserts: a run whose `result.gaps` genuinely contains a `no write proof` / `partial-artifacts` entry **does** fail G3 (this is the existing sound behavior and must survive the fix — `bench/tests/test_check.py::test_workflow_return_partial_artifacts_gap_fails_g3` already covers one such case and must keep passing).

## Evidence

Literal counts in the committed bundle (2026-07-27):

```
$ grep -c "partial-artifacts" workflows/pipeline.js
6
$ grep -c "no write proof" workflows/pipeline.js
4
$ grep -n "no write proof" workflows/pipeline.js
2698:      return { ok: false, reason: 'slice-input writer echo did not cover all dispatched slice paths (no write proof)' };
3472:      return partial('writer echo did not account for all four planned artifact paths (no write proof)', extraGaps);
3569:    return fail('writer echo did not cover all three primary artifact paths (no write proof)');
3676:    if (!writtenPaths.has(p)) return { ok: false, reason: `assemble receipt did not write ${p} (no write proof)` };
$ wc -c workflows/pipeline.js
230311 workflows/pipeline.js
```

The scan-order bug, verbatim from the current source (2026-07-27):

```
bench/runner/check.py:302-327 (_scan_degrade_text):
        try:
            text = path.read_text(encoding="utf-8", errors="replace")
        except OSError:
            continue
        matched = bool(_DEGRADE_RE.search(text))
        if not matched and path.suffix == ".json":
            try:
                data = json.loads(text)
            except (json.JSONDecodeError, ValueError):
                data = None
            if data is not None:
                matched = any(_DEGRADE_RE.search(g) for g in _gaps_strings(data))
        if matched and label not in seen:
            seen.add(label)
            hits.append(label)
```

Reproduced against today's 3-PR functional smoke (`bench/workspace/runs/smoke-20260727-205454-f99d948`, commit `f99d948`), 2026-07-27:

```
$ python3 bench/run.py --check smoke-20260727-205454-f99d948
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6:code-gauntlet-findings-267d8be1.json[0..4]: origin=unknown (verify/slice degrade)
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6: writer degrade signal in workflows/wf_089793e7-ca8.json (no-write-proof / partial-artifacts)
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6: writer degrade signal in raw.json (no-write-proof / partial-artifacts)
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6: writer degrade signal in code-gauntlet-report-267d8be1.md (no-write-proof / partial-artifacts)
  FAIL: pr-calcom-cal.com-22345: writer degrade signal in workflows/wf_a6eaf1a7-e1f.json (no-write-proof / partial-artifacts)
  FAIL: pr-grafana-grafana-80329: writer degrade signal in workflows/wf_cb7cea68-9a4.json (no-write-proof / partial-artifacts)
Check smoke-20260727-205454-f99d948: ok=False pr_dirs=3 comments=21 findings_files=3 workflow_records=3 script_paths=3 unknown_origin=5
```

The grafana and cal.com flags are false positives — their parsed `result.gaps` is clean (only tolerated `artifact-content-proof` entries, a distinct and pre-existing content-divergence issue tracked separately) and `result.artifactPaths` is fully populated:

```
$ python3 -c "
import json
data = json.load(open('bench/workspace/runs/smoke-20260727-205454-f99d948/pr-grafana-grafana-80329/workflows/wf_cb7cea68-9a4.json'))
result = data['result']
print('gaps:', json.dumps(result['gaps'], indent=2))
print('artifactPaths:', json.dumps(result['artifactPaths'], indent=2))
"
gaps: [
  "artifact-content-proof: .../code-gauntlet-findings-04cfa3bf.json bytes on disk differ from the payload handed to the writer (expected 31595 chars/checksum fnv1a32:0xe43688ef, got 31579/fnv1a32:0x6554cf07)",
  "artifact-content-proof: derived document .../code-gauntlet-post-review-04cfa3bf.json does not match the pipeline's own derivation ...",
  "artifact-content-proof: derived document .../code-gauntlet-checkpoint-all-04cfa3bf.json does not match the pipeline's own derivation ..."
]
artifactPaths: {
  "findings": ".../code-gauntlet-findings-04cfa3bf.json",
  "report": ".../code-gauntlet-report-04cfa3bf.md",
  "postReview": ".../code-gauntlet-post-review-04cfa3bf.json",
  "checkpoints": ".../code-gauntlet-checkpoint-all-04cfa3bf.json"
}
```

None of those three `gaps` entries contain "no write proof" or "partial-artifacts" (verified: `_gaps_strings` over this exact structure returns no `_DEGRADE_RE` match), yet G3 still failed this PR — because `_scan_degrade_text` matched the raw `script` field of `wf_cb7cea68-9a4.json` (223,721 chars, containing the pipeline bundle) before ever reaching the structured branch. Same shape for cal.com#22345.

By contrast, discourse-graphite#6's G3 failures are **real**: its `result.gaps` for `writeArtifacts` reads `"assemble script reported failure (expected artifact is not valid JSON: ... Expecting ',' delimiter: line 99 column 182 ...) — artifacts not persisted (partial-artifacts)"`, and `artifactPaths` is entirely `null` — a genuine writer failure (tracked as Issue 2). Its `origin=unknown` gate hit is also real (verify nonce mismatch, tracked as Issue 3).

Corroborating evidence already on record: a same-day comment on the (closed) issue #28 thread from the repo owner independently reproduces this exact defect and its false-positive count (2 of 3 PRs).

- Invert the existing order for `.json` carriers: parse first and scan `result.gaps` via the already-present `_gaps_strings()` helper; fall back to the raw-text scan only when the file fails to parse as JSON at all (mirroring the JSON-decode-failure fallback path that already exists at line 318-321, just promoted to the primary path for JSON carriers).
- Keep the raw-text scan as the sole mechanism for non-JSON carriers (`code-gauntlet-report-*.md`) where there is no `gaps` structure to consult — `test_report_degrade_fails_g3` already covers this shape and should be unaffected.

## Out of scope

- The `origin=unknown` scan (G3's other half) — untouched, already sound, and validated against a real failure in today's smoke.
- Any change to what the pipeline itself (`workflows/src/stages.js`) treats as a degrade signal or writes into `gaps` — this is a bench-side detection bug only, not a pipeline behavior change.
- The underlying writer-transcription defect that discourse-graphite#6 genuinely surfaced (Issue 2) and the verify nonce-mismatch defect (Issue 3) — this issue is scoped to the checker's false-positive detection logic only.

## Verification

Tier: **suites-only** — this is a bench-harness detection-logic fix, not a pipeline behavior change. `bench/tests/test_check.py` (extended per Requirements 4–5) plus the existing suite must pass. As a live confirmation once landed, re-running `python3 bench/run.py --check smoke-20260727-205454-f99d948` against the already-collected smoke artifacts should show G3 passing for grafana/cal.com and still correctly failing for discourse-graphite (tracked separately by Issues 2 and 3).

leehopper:
Closes #52.

## Problem

G3's writer-degrade half raw-text-scanned each carrier's whole bytes **first**, and only consulted parsed structure if that scan missed. Every collected `workflows/wf_*.json` embeds the ~230 KB `workflows/pipeline.js` bundle in its `script` field, and the bundle's own source carries the sentinels as ordinary string constants (`no write proof` ×4, `partial-artifacts` ×6) — so every wf record matched, always.

In the 2026-07-27 smoke, 2 of 3 PRs were false-flagged while their `result.gaps` were clean. Since most open work-queue issues name that smoke as their verification gate, "the smoke passed" carried near-zero information about writer degradation specifically.

## Fix

`_DEGRADE_CARRIER_POLICY` classifies each carrier, and `_DEGRADE_SCAN_PATTERNS` is **derived from its keys** so the two cannot desync. Lookup is direct indexing, so an unclassified future pattern raises rather than guessing.

| Carrier | Policy | Why |
|---|---|---|
| `workflows/wf_*.json` | structured | owns `result.gaps`; embeds the bundle in `script` |
| `code-gauntlet-checkpoint-all-*.json` | structured | owns a real `gaps` array |
| `raw.json` | text | envelope; signal is prose in `.result`, no `gaps` array exists |
| `code-gauntlet-report-*.md` | text | not JSON |

A structured carrier is judged **solely** by its parsed `gaps` array. The load-bearing detail: presence of any gaps list — *including an empty one* — settles it. Treating "gaps present but clean" as "no gaps" would send the healthy `result.gaps == []` shape back to the byte scan and the bug would return.

A structured carrier that will not parse, or has no `gaps` at all, is still scanned rather than reported clean — silently passing an artifact we could not read would be a new way to lose a run — but with the bundle-bearing `script` value blanked first. That blanking is scoped to structured carriers only: its unterminated-field pattern necessarily runs to end-of-text, so applying it to a text carrier could swallow genuine trailing prose, and text carriers never embed the bundle (measured 0/131 `raw.json`, 0/27 report `.md`).

The `origin=unknown` half of G3 is untouched (requirement 2).

## Validation

Tier: **suites-only**, per the issue.

| Check | Result |
|---|---|
| `pytest bench/tests/ -q` | 529 passed (522 on main, +7 new) |
| `pytest tests/ -q` | 763 passed |
| `node --test workflows/test/*.test.js` | 392 passed |

**Full-corpus sweep** — old vs new `_scan_degrade_text` across all 137 pr-dirs in 29 run dirs: exactly **2** flags disappear (grafana#80329, cal.com#22345), both confirmed false positives (parsed gaps hold only `artifact-content-proof` checksum entries; a whole-structure walk excluding `script`/`gaps` found zero sentinels in any sibling field). **0 genuine-signal regressions, 0 new flags.**

**Live confirmation** — `python3 bench/run.py --check smoke-20260727-205454-f99d948`: grafana and cal.com clear; discourse-graphite's 3 genuine hits (wf record, `raw.json`, report `.md`) and its 5 `origin=unknown` findings all preserved. The remaining `pipeline_version '3.2.0' != '3.2.1'` lines are pre-existing G4 drift — HEAD has moved past the smoke commit — and are present on main too.

## Tests added

Requirements 4 and 5 are covered by `test_wf_script_field_bundle_literals_do_not_fail_g3` and `test_wf_script_field_bundle_literals_with_genuine_gap_fails_g3`; the pre-existing `test_workflow_return_partial_artifacts_gap_fails_g3` passes unmodified. Also added: unparseable-record handling in both directions, structured-carrier authority over stray bytes, text-carrier prose preservation, and a policy/pattern coverage guard.

Independently verified that 4 of the 7 new tests fail against `main` (real regression guards); 2 are the negative controls requirement 5 asks for; 1 guards the text-carrier scoping described above.

## Known, non-blocking

- A structured carrier with `gaps == []` but a genuine signal in a *sibling* field is not flagged. This is the design requirement 1 mandates; `raw.json` independently carries the same signal as prose, and CLAUDE.md documents that `writeArtifacts` reports gaps exclusively via `gaps[]`.
- Unrelated and pre-existing (identical on main): `_DEGRADE_RE`'s third alternative `partial.artifacts` has an unescaped `.`, so it also matches `partial artifacts` / `partialXartifacts` in prose — a false-positive vector for the two text carriers. Out of scope here; worth its own issue.

https://claude.ai/code/session_01PTKzwKm4YbznKaw9gteQJC

---

> [!NOTE]
> **Low Risk**
> Changes are limited to the mechanical smoke checker and its tests; behavior is regression-tested and removes known false positives without dropping real degrade signals on text carriers.
> 
> **Overview**
> Fixes **false G3 smoke failures** where every `workflows/wf_*.json` looked degraded because the checker regex-scanned full artifact bytes before consulting structure. Those records embed the whole `workflows/pipeline.js` bundle in `script`, and the bundle source contains the “no write proof” / “partial-artifacts” phrases as literals.
> 
> G3 now uses **`_DEGRADE_CARRIER_POLICY`**: `wf_*.json` and checkpoint JSON are **structured** and judged only from parsed `gaps` (including `gaps: []` as clean). Unparseable or gap-less structured files fall back to a text scan with the `script` value blanked via escape-aware regex. **`raw.json` and report markdown stay text-only** scans with no script stripping.
> 
> `bench/MEASUREMENT.md` gate 3 text matches this behavior. **Seven new tests** cover bundle literals, genuine gaps, truncated JSON, policy sync, and text-carrier scoping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a3c923ff0bd5e3a0c725ab5f88edc4a57631cda0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/bench/runner/check.py
+++ b/bench/runner/check.py
@@ -50,7 +50,7 @@ _DEGRADE_RE = re.compile(
 
 # Value of a JSON ``"script"`` field — a workflow record's echo of the whole
 # pipeline bundle, whose own source carries the degrade sentinels as ordinary
-# string constants (issue #52). The escape-aware body is required: a naive
+# substrings (issue #52). The escape-aware body is required: a naive
 # ``[^"]*`` stops at the first ``\"`` and leaves most of the bundle behind.
 _SCRIPT_FIELD_RE = re.compile(r'"script"\s*:\s*"(?:\\.|[^"\\])*"', re.DOTALL)
 # The same field truncated mid-write, with no closing quote.
@@ -72,8 +72,9 @@ PIPELINE_REL = Path("workflows") / "pipeline.js"
 #   workflows/wf_*.json  carries the compact return at ``result.gaps``. It also
 #       echoes the whole ~230 KB workflows/pipeline.js bundle into its
 #       ``script`` field, and that bundle's source contains the sentinels as
-#       ordinary string constants ("no write proof" x4, "partial-artifacts"
-#       x6) — so a raw scan matches on EVERY collected record, degraded or not.
+#       ordinary substrings ("no write proof" x4 string/template literals;
+#       "partial-artifacts" x6 — 1 string literal + 5 comments) — so a raw
+#       scan matches on EVERY collected record, degraded or not.
 #   code-gauntlet-checkpoint-all-*.json  the persisted checkpoint's ``gaps``.
 #
 # TEXT carriers have no ``gaps`` structure to parse, so a raw-text scan is the
@@ -337,20 +338,12 @@ def _gaps_lists(node):
             yield from _gaps_lists(value)
 
 
-def _gaps_strings(node):
-    """Yield string entries from any ``gaps`` array nested under ``node``."""
-    for gaps in _gaps_lists(node):
-        for item in gaps:
-            if isinstance(item, str):
-                yield item
-
-
 def _strip_script_field(text):
     """Blank out any JSON ``"script"`` field value ahead of a raw-text scan.
 
     ``workflows/wf_*.json`` echoes the entire pipeline bundle into that field,
     and the bundle's own source carries the degrade sentinels as ordinary
-    string constants (issue #52). Handles both a well-formed value and one
+    substrings (issue #52). Handles both a well-formed value and one
     truncated mid-write.
 
     Only STRUCTURED carriers get this treatment. The unterminated-field pattern
@@ -390,8 +383,14 @@ def _scan_degrade_text(pr_dir):
                 data = json.loads(text)
             except (json.JSONDecodeError, ValueError):
                 data = None
-            if data is not None and any(True for _ in _gaps_lists(data)):
-                matched = any(_DEGRADE_RE.search(g) for g in _gaps_strings(data))
+            gaps_lists = list(_gaps_lists(data)) if data is not None else []
+            if gaps_lists:
+                matched = any(
+                    _DEGRADE_RE.search(item)
+                    for gaps in gaps_lists
+                    for item in gaps
+                    if isinstance(item, str)
+                )
             else:
                 matched = bool(_DEGRADE_RE.search(_strip_script_field(text)))
         else:
--- a/bench/runner/check.py
+++ b/bench/runner/check.py
@@ -48,21 +48,57 @@ _DEGRADE_RE = re.compile(
     re.IGNORECASE,
 )
 
+# Value of a JSON ``"script"`` field — a workflow record's echo of the whole
+# pipeline bundle, whose own source carries the degrade sentinels as ordinary
+# string constants (issue #52). The escape-aware body is required: a naive
+# ``[^"]*`` stops at the first ``\"`` and leaves most of the bundle behind.
+_SCRIPT_FIELD_RE = re.compile(r'"script"\s*:\s*"(?:\\.|[^"\\])*"', re.DOTALL)
+# The same field truncated mid-write, with no closing quote.
+_SCRIPT_FIELD_OPEN_RE = re.compile(r'"script"\s*:\s*"(?:\\.|[^"\\])*\Z', re.DOTALL)
+
 # Expected pipeline entry relative to the plugin/repo root.
 PIPELINE_REL = Path("workflows") / "pipeline.js"
 
-# Where G3 looks for writer no-write-proof / partial-artifacts signals.
-# writeArtifacts puts that gap on the compact Workflow return (gaps[]), which
-# lands in collected workflows/wf_*.json and often in raw.json's .result text —
-# NOT in the persisted report/checkpoint (those are written before the echo
-# proof runs). Report/checkpoint remain scanned as secondary carriers.
+# Where G3 looks for writer no-write-proof / partial-artifacts signals, and how
+# each carrier is read (issue #52). writeArtifacts puts that gap on the compact
+# Workflow return (gaps[]), which lands in collected workflows/wf_*.json and
+# often in raw.json's .result text — NOT in the persisted report/checkpoint
+# (those are written before the echo proof runs). Report/checkpoint remain
+# scanned as secondary carriers.
+#
+# STRUCTURED carriers own a real ``gaps`` array, so that parsed array is the
+# authoritative — and only — signal consulted; their raw bytes are never
+# regex-scanned:
+#   workflows/wf_*.json  carries the compact return at ``result.gaps``. It also
+#       echoes the whole ~230 KB workflows/pipeline.js bundle into its
+#       ``script`` field, and that bundle's source contains the sentinels as
+#       ordinary string constants ("no write proof" x4, "partial-artifacts"
+#       x6) — so a raw scan matches on EVERY collected record, degraded or not.
+#   code-gauntlet-checkpoint-all-*.json  the persisted checkpoint's ``gaps``.
+#
+# TEXT carriers have no ``gaps`` structure to parse, so a raw-text scan is the
+# only mechanism that can ever see their signal:
+#   raw.json  the child CLI's result envelope, whose ``result`` is free prose
+#       (the model's final turn), never a nested ``gaps`` array. Measured over
+#       the retained corpus: 0/131 carry a structured ``gaps``, and 0/131 embed
+#       the bundle (max 9.5 KB — invoke.py never forwards Workflow tool-call
+#       inputs here), so scanning its bytes is both necessary and safe.
+#   code-gauntlet-report-*.md  markdown; there is no parse step to consult.
+#
 # Do not include bench-only fixture names such as deep-review-report.md.
-_DEGRADE_SCAN_PATTERNS = (
-    "workflows/wf_*.json",
-    "raw.json",
-    "code-gauntlet-report-*.md",
-    "code-gauntlet-checkpoint-all-*.json",
-)
+_DEGRADE_STRUCTURED = "structured"
+_DEGRADE_TEXT = "text"
+
+_DEGRADE_CARRIER_POLICY = {
+    "workflows/wf_*.json": _DEGRADE_STRUCTURED,
+    "raw.json": _DEGRADE_TEXT,
+    "code-gauntlet-report-*.md": _DEGRADE_TEXT,
+    "code-gauntlet-checkpoint-all-*.json": _DEGRADE_STRUCTURED,
+}
+
+# Single source of truth: the scanned patterns ARE the policy's keys, so the
+# two cannot desync. Scan order follows insertion order above.
+_DEGRADE_SCAN_PATTERNS = tuple(_DEGRADE_CARRIER_POLICY)
 
 
 def _load_json(path):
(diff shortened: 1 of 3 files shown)


leehopper (commented):
Code Gauntlet reviewed this PR (333 changed lines across bench/runner/check.py, bench/tests/test_check.py, bench/MEASUREMENT.md). 2 high-confidence findings survived validation and blind challenge (both LOW severity: a comment-accuracy nit and a simplification suggestion). No bugs or security issues found; the fix is well-scoped to the bench/ harness and covered by 5 new tests.

---
Generated by code-gauntlet | Reviewed up to: 899aeb58bfd3286246427cfec1fee5fd0c0b3957

bench/runner/check.py
```diff
@@ -48,21 +48,57 @@
...
+# scanned as secondary carriers.
+#
+# STRUCTURED carriers own a real ``gaps`` array, so that parsed array is the
+# authoritative — and only — signal consulted; their raw bytes are never
+# regex-scanned:
+#   workflows/wf_*.json  carries the compact return at ``result.gaps``. It also
+#       echoes the whole ~230 KB workflows/pipeline.js bundle into its
+#       ``script`` field, and that bundle's source contains the sentinels as
+#       ordinary string constants ("no write proof" x4, "partial-artifacts"
+#       x6) — so a raw scan matches on EVERY collected record, degraded or not.
```
leehopper: **💡 [LOW] "ordinary string constants" claim overcounts actual string literals for 'partial-artifacts'**

The new comment says the pipeline bundle contains the degrade sentinels as "ordinary string constants ('no write proof' x4, 'partial-artifacts' x6)". The x4 count for 'no write proof' is accurate (all four are genuine string/template literals), but only 1 of the 6 'partial-artifacts' occurrences in workflows/pipeline.js is an actual string constant (line 3418); the other five (lines 3393, 3408, 3476, 4185, 4385) are JS comments, not string constants. The substring-match reasoning for the fix still holds either way, but the specific factual claim about what kind of source construct carries the sentinel is inaccurate for 5/6 instances.

bench/runner/check.py
```diff
@@ -271,56 +307,95 @@ def _check_echo_identity(pr_dir, repo_root, label):
...
+        if _DEGRADE_CARRIER_POLICY[pattern] == _DEGRADE_STRUCTURED:
             try:
                 data = json.loads(text)
             except (json.JSONDecodeError, ValueError):
                 data = None
-            if data is not None:
+            if data is not None and any(True for _ in _gaps_lists(data)):
                 matched = any(_DEGRADE_RE.search(g) for g in _gaps_strings(data))
+            else:
+                matched = bool(_DEGRADE_RE.search(_strip_script_field(text)))
```
leehopper, line 395: **💡 [LOW] Non-idiomatic presence check plus redundant re-walk of the gaps tree**

Line 393's any(True for _ in _gaps_lists(data)) is an indirect way to ask 'does at least one gaps list exist', and immediately after, line 394's _gaps_strings(data) call internally re-invokes _gaps_lists(data), walking the same nested structure a second time. Materializing the generator once and reusing it removes both the unusual idiom and the duplicate traversal.

Issue #29, opened 2026-07-23
V3.2 fable mode: add alternate-model policy support, measured under the paired holdout protocol


leehopper:
## Problem / Goal

The shipped plugin has exactly one model policy — discovery agents on Sonnet, security-reviewer on Opus, stage agents on Sonnet — and `model_tier`'s only valid value is `optimized`. There is no supported way to run the pipeline under an alternate model policy ("fable mode"), and the dormant `frontier` plumbing that once gestured at this was removed entirely pre-3.0 (`901c31a`).

Goal: build alternate-model support fresh — model-mode support in `resolvePolicy` and the policy waist, a Phase-1 configuration surface, and docs — and measure it under the paired benchmark protocol against the orchestrator-choice baseline (mini-subset B), spending the reserved holdout. This is the designated consumer of that holdout; it is spent here and nowhere else.

Carried from #17 (closed; its closeout comment holds the V3.1 outcomes).

## Requirements

1. **Strict sequencing:** this is V3.2 and runs after the deterministic-hardening and consolidation items — #25 and #22. No earlier queued item may touch model modes.
2. **Build fresh:** no resurrection of the removed `frontier` plumbing (`901c31a`). Design model modes from the current policy waist.
3. **Scope:** model-mode support in `resolvePolicy` + the policy waist, a Phase-1 configuration surface, and docs. Contract preservation, precisely: the env knob `CODE_GAUNTLET_MODEL_TIER` keeps its fail-loud contract and REVIEW.md keeps its self-heal contract for **unrecognized** values; this issue expands the recognized-value set to include the new mode(s). An explicit env pin to an unrecognized value stays an error (never healed); an unrecognized REVIEW.md value still self-heals to `optimized` with the loud methodology warning.
4. **Benchmark constraint:** fable is measured under the same paired protocol against the mini-subset B baseline, which must exist first — owner-triggered per the measurement runbook in the #28 item (#28). Holdout `holdout-20260721-085348-eec15be` is spent here and nowhere else.
5. **Model-identity caveat:** request-layer pinning is the only controllable layer — platform routing may serve an alias (the `[1m]` observation below). Benchmark interpretation must account for this.
[...]

PR #59, opened 2026-07-28
fix(context): compute the shared-context read plan instead of asking agents to paginate (#48)

### LINKED ISSUE #48: security-reviewer reviewed roughly half the diff

leehopper:
Closes #48.

## What broke

On profiled run `wf_cef39739-577`, every one of the 7 discovery agents' **first** `Read` of the 95,057-byte / 2,028-line shared context file returned 58,145 characters, ending at line 1083 of 2028. **No tool result carried a truncation notice** — checked programmatically across all 7 transcripts. Six agents inferred the cutoff from the line numbering and paginated to the file's true end. `security-reviewer` did not: one `Read`, then `{"findings":[],"complete":true}` after seeing roughly the first half of the diff and project rules.

Nothing in the run distinguished that from a clean empty result. The report, the artifacts, and the transcript all looked normal.

## The fix, and why it is not a prompt change

The obvious move — tell the agent to "paginate with `offset` until you reach the end" — puts a judgment call where arithmetic belongs, and it is precisely that judgment which failed 1 time in 7. So the stop condition is no longer the agent's to work out:

1. **Phase 2 measures.** The skill computes `contextLines` / `contextChars` in the same `python3 -c` that writes the context file, and stamps them onto the args waist. The workflow has no disk and cannot measure the file itself, so this is the only path for it.
2. **`contextReadPlan(lines, chars)` computes** the exact `[{offset, limit}]` covering the file — pure arithmetic, chunked under **both** observed platform caps (the documented 2,000-line `Read` cap and the ~58,000-char return this run actually hit). For the profiled file: `Read(offset=1, limit=641), Read(offset=642, limit=641), Read(offset=1283, limit=641), Read(offset=1924, limit=105)`.
3. **`sharedContextLine` enumerates them** in the Summarize, Discover, and Validate prompts. The agent's job is to make the listed calls.

Read-to-end-by-`offset` survives only as the fallback wording when no measurement was stamped, so every pre-existing caller — bench included — keeps working unchanged.

## Requirement coverage

| # | Requirement | Where |
|---|---|---|
| 1 | Contracts make reading to the end explicit | `sharedContextLine` (all 3 stages) + the complete-read block in all 10 file-reading agents |
| 2 | Must not depend on a truncation notice | Every prompt variant states a partial `Read` is unannounced; pinned by test |
| 3 | Deterministic regression check | 4 guards, below |
| 4 | Correct #38 / #29 | #38's Out-of-scope bullet edited in place + comment; #29 follow-up comment |

## The guards, and how they were derived

An adversarial review of this change **defeated the first draft of guard 1** — which counted the literal `"Read the shared context at "` in `stages.js` — by adding a real dispatched stage saying `"Open the shared context at <path>"` with no plan and no warning. Whole suite stayed green. A phrase count cannot enforce a structural property. The guards were re-derived from that failure and now key on the context **path**, which any stage naming the file must necessarily embed:

- **behavioral (primary, wording-independent):** drive `runWith`; every dispatched prompt *containing the path* must carry the plan — or the count-free fallback when unmeasured.
- **source:** `.contextPath` is dereferenced in exactly one function; every stage input threading `contextPath` also threads `...contextSize`.
- **contract:** the complete-read block is byte-identical across all 10 agents (`tests/test_agent_contracts.py::TestCompleteReadContract`).

Re-running the same evasion against the new guards fires **3 of 4**.

The same review found `contextReadPlan` unbounded on `lines`: `contextLines: Number.MAX_SAFE_INTEGER` OOM-killed the node process — a V8 fatal error, so `runWith`'s top-level catch never runs, nothing dispatches, and no gap is recorded. Now bounded at both layers (plan-level chunk ceiling checked *before* the first allocation; a fail-loud waist cap where the bad value is still attributable to its producer). It also caught the test fixture using the file's **byte** count (95,057) where `contextChars` is **characters** (94,784) — corrected, along with the dependent plan expectations.

## Incidental

Removes a dead `contextPath` from the Challenge stage input, surfaced by the pairing guard. `challengeStage` never read it — the challenger is structurally blind by design — but a threaded-and-unused path invites a future edit to "use the context we already have" and quietly break the blindness the challenge round exists for.

## Verification

Tier **smoke**, per the issue. Deterministic suites only; no benchmark spend.

- `python -m pytest tests/ -q` → **769 passed**, 296 subtests
- `python -m pytest bench/tests/ -q` → **529 passed**, 22 subtests
- `node --test workflows/test/*.test.js` → **419 passed**
- `node workflows/build.js && git diff --exit-code workflows/pipeline.js` → clean
- `pre-commit run --all-files` → clean

The documented line-count formula was checked against the real profiled file still on disk: it yields **2028**, matching what the agents' `Read`s actually reached, where `wc -l` reports **2027** — the off-by-one the docs now warn against.

**Not claimed:** no live review ran on this branch, so nothing here is evidence about recall or noise. Per the issue's Verification section, this is escalation-eligible for an owner-triggered paired run if you judge that seeing the full diff plausibly changes `security-reviewer`'s output — the profiled run (`total_seen: 6, findings: []` on one PR) cannot answer that.

https://claude.ai/code/session_01EJs9TuugGEbMDrB4hzGb4h

---

> [!NOTE]
> **Medium Risk**
> Changes core review input plumbing (what agents see of the diff) and prompt assembly for summarize/discover/validate; regressions could miss diff content or mis-size reads, though extensive guards and tests target that failure mode.
> 
> **Overview**
> Fixes **issue #48**, where a single `Read` of the shared context file could return only part of the diff with **no truncation notice**, so agents (notably `security-reviewer`) could stop halfway and still report `complete: true`.
> 
> **Phase 2** now measures the context file when it is written and stamps optional `contextLines` / `contextChars` on the args waist (using `cat -n`-style line counting, not `wc -l`). **`contextReadPlan`** and **`sharedContextLine`** in `workflows/src/stages.js` turn that into explicit `Read(offset, limit)` instructions under line and character caps; **`runWith`** builds the sentence once and passes **`contextLine`** into summarize, discover, and validate instead of letting each stage hand-roll a path-only “read the context” line. Missing or unplannable measurements still fall back to stepped read-to-end wording but emit **`context_unmeasured`** / **`context_unplannable`** gaps so degradation is visible.
> 
> **Args validation** shape-checks the new fields (including ceilings to avoid OOM from absurd sizes). Ten file-reading agents gain a duplicated **complete-read contract** (`references/complete-read-contract.md`) with Python tests for byte-identical copies. The **challenger** no longer receives a dead `contextPath` on the challenge stage input. Docs, skill/triage prose, **`pipeline.js`**, and broad JS/Python tests (including behavioral guards keyed on the context path in prompts) accompany the change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 676f9c9db2d3a617a3ff17cc93551c5be42e354d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:
**Note on the `Run Linting` failure in the first CI run — it was not from this change.**

`pre-commit run --all-files` fails on `main` as of the 3.2.2 release commit `6430af8`: semantic-release's generated changelog entry leaves two consecutive blank lines before the `## v3.2.1` heading, markdownlint's MD012 auto-fixes it, and the hook exits non-zero. Every branch cut from that commit inherits the failure. This PR's own markdown was clean — the other 14 checks passed on the first run.

`main` is merged in here and the one stray blank line is deleted in `8616248`, purely so this PR's CI can go green. That is a stopgap: the generator reintroduces the pattern on the next release. The durable fix is filed as #60.

Re-verified against the merged tree (`main` @ `6430af8` + this branch):

- `python -m pytest tests/ -q` → 769 passed, 296 subtests
- `python -m pytest bench/tests/ -q` → 529 passed, 22 subtests
- `node --test workflows/test/*.test.js` → 419 passed
- `node workflows/build.js && git diff --exit-code workflows/pipeline.js` → clean (the merge brought a version bump into `pipeline_entry.js`; the bundle was rebuilt and matches)
- `pre-commit run --all-files` → clean

leehopper:
## Debt accounting — addressed in `2689972`, not deferred

Prompted by the question of whether this PR's "extensive guards and backward-compatible optional waist fields" were tracked debt. Reviewing my own change against that framing turned up one defect and two genuine debt items. All three are fixed here rather than filed.

**1. The optional field degraded silently — a defect, not debt.** I probed it: with the measurement absent, `runWith` returned `ok: true` with **zero gaps** while every prompt fell back to the weaker judgment-based wording. Phase 2 is model-executed and *can* skip the stamp, so a regression there would revert this entire fix with nothing anywhere saying so — the same silent-failure class as #48 itself, and a direct violation of the contract `args.js` states for tolerated nulls: *"TOLERATED IS NOT SILENT."* `runWith` now emits a `context_unmeasured` gap naming the field, the consequence, and the remedy. Deliberately still non-fatal — hard-failing would trade a partial read for a dead review.

**2. Two guards were text scans standing in for a structural property.** They existed only because three stages could each reach `inp.contextPath` and hand-roll a sentence. `runWith` now calls `sharedContextLine` **once** and threads the resulting *string*; no stage receives the path or the size, so no stage is *able* to build a context-read instruction. The capability is gone rather than policed, and the two brittle scans collapse into one structural assertion. Re-verified against the same evasion that defeated the original phrase-count guard — it still fails the suite.

**3. The fallback wording was itself a judgment call.** Without a measurement the terminus is unknowable, so end-detection is unavoidable — but the stepping was not. It now enumerates fixed 750-line steps, leaving only "did that call return anything" rather than "have I read enough yet".

### Correction to this PR's description

I justified the optional fields as back-compat for "bench and older callers." That was wrong and I've fixed it in `CLAUDE.md`, `args.js`, and `phase2-triage.md`. **Bench does not assemble the args waist** — it invokes the skill. `SKILL.md` is the only producer, and it ships in the same plugin as the bundle, so there is no meaningful version-skew window. The honest rationale for optionality is the live one: Phase 2 is model-executed and can skip a step, and degrading-with-disclosure beats killing the review.

### Re-verified on the merged tree

`pytest tests/` 769 passed · `pytest bench/tests/` 529 passed · `node --test` 421 passed · bundle rebuilt · `pre-commit` clean.

No remaining known debt from this change. #60 (the semantic-release changelog lint break) is unrelated and pre-existing.

workflows/src/stages.js
```diff
@@ -2132,9 +2223,34 @@ export async function runWith(ctx, rawArgs) {
...
+  // stamp, so this is a live path, not a theoretical one — and without a gap the whole fix
+  // reverts with nothing anywhere saying so. Same contract args.js states for a tolerated
+  // null: what was lost, and what to do about it.
+  const contextSizeGap = A.contextLines === undefined
+    ? ['context_unmeasured: args.contextLines was not stamped, so the shared-context read plan could not be computed — '
+      + 'every agent got the count-free "read until a Read returns no further content" wording instead of the exact '
+      + 'Read calls covering the file. That restores the failure mode of issue #48: a Read returns part of a large file '
+      + 'with no truncation notice, and an agent that stops there reviews only what it saw. Stamp contextLines '
+      + '(and contextChars) from the Phase 2 step that writes the context file.']
+    : [];
```
leehopper: **💡 [LOW] context_unmeasured gap is not emitted when a stamped contextLines produces an unbuildable read plan**

runWith decides whether to emit the 'context_unmeasured' disclosure gap solely from `A.contextLines === undefined`, but sharedContextLine/contextReadPlan can independently fall back to the count-free wording whenever the derived chunk count exceeds READ_PLAN_MAX_CHUNKS (2000). In that case the agent silently gets the exact same unmeasured/fallback wording as a truly-unstamped run, but no gap is recorded, violating the PR's own stated invariant that any degradation to the count-free wording must be disclosed.

leehopper: **💡 [LOW] No test covers the 'measured but too-large-to-plan' dead zone between the plan's chunk ceiling and the waist's size ceiling**

runWith only emits the context_unmeasured gap when contextLines is strictly undefined, but contextReadPlan/sharedContextLine silently fall back to count-free wording for validly-stamped contextLines values that still exceed READ_PLAN_MAX_CHUNKS (roughly 1.5M-5M lines), and no test exercises this combination.

workflows/src/stages.js
```diff
@@ -38,6 +38,98 @@ function modelFor(agentType, policy) {
...
+  if (plan.length === 1) {
+    return `${head} ${warn} It is ${total} lines: read it with Read(offset=1, limit=${plan[0].limit}), and if that call does not reach line ${total}, continue with offset set past the last line you received until it does. `;
+  }
+  const tail = `Make ALL of them — stopping early means you review only the part you saw. If any call does not land where the plan says, continue with offset set past the last line you received until you reach line ${total}.`;
+  if (plan.length <= READ_PLAN_MAX_ENUMERATED) {
+    const calls = plan.map((p) => `Read(offset=${p.offset}, limit=${p.limit})`).join(', ');
+    return `${head} ${warn} It is ${total} lines, which takes exactly ${plan.length} Read calls: ${calls}. ${tail} `;
+  }
+  const span = plan[0].limit;
+  return `${head} ${warn} It is ${total} lines, which takes exactly ${plan.length} Read calls of limit=${span}, at offsets 1, ${1 + span}, ${1 + 2 * span}, … stepping by ${span} through line ${total}. ${tail} `;
```
leehopper: **💡 [LOW] No boundary test at the enumerated-vs-rule-form threshold (READ_PLAN_MAX_ENUMERATED = 10) in sharedContextLine**

sharedContextLine switches wording form at plan.length <= READ_PLAN_MAX_ENUMERATED (10), but no test constructs a plan of exactly 10 or 11 chunks, so an off-by-one change to the comparison would go undetected.

workflows/src/stages.js
```diff
@@ -38,6 +38,98 @@ function modelFor(agentType, policy) {
...
+}
```
leehopper, line 132: **💡 [LOW] sharedContextLine repeats the `${head} ${warn}` prefix in all four return branches**

Every return statement in sharedContextLine begins with the identical `${head} ${warn}` prefix; extracting a `prefix` const would remove the repeated concatenation with byte-identical output.

leehopper (commented):
code-gauntlet found 6 low-severity issues: 1 bug (undisclosed degradation dead zone), 2 stale comments, 2 test-coverage gaps, 1 minor simplification. No critical/high/medium findings.

---
Generated by code-gauntlet | Reviewed up to: 2689972840c41d152195e6633c6ee803261d3097

Issue #25, opened 2026-07-23
Harden the verify executor/writer boundary: per-id delta echo, slice-input content proof, per-slice degradation, and a deterministic report degradation banner


leehopper:
## Problem / Goal

The verify boundary trusts a subagent to transcribe full finding arrays through StructuredOutput, and its failure handling is all-or-nothing and invisible. Four connected defects on this one surface: (1) the executor echoes complete verified/eliminated findings by value, so any transcription error (a live run observed 10 verified/0 eliminated on disk echoed as 7/3 with a valid receipt) corrupts or degrades output; (2) a corrupted verify-slice input file is only discovered after the executor has already spent (observed: tool-call markup appended after the slice JSON); (3) one untrusted slice degrades EVERY finding to `origin=unknown`, including findings from trusted slices; (4) the report renderer presented such a fully-degraded run as "0 unverified/pipeline-degraded findings" while all 19 delivered findings carried `origin=unknown`.

Goal: `verify_findings.py` echoes only a compact per-id delta joined deterministically in `verifyStage`; slice input is proven on disk before dispatch; degradation is per-slice; and a verify-degraded run always renders an unmissable banner. Residual degradation stays possible but becomes honest and visible.

## Requirements

1. **Delta echo.** `verify_findings.py` receipt mode additionally emits a compact per-finding delta map keyed by id; the executor echoes receipt + deltas only; `verifyStage` joins deltas onto the findings it already holds by value. The enriched set after the join must be equivalent to today's trusted-path output for every field downstream stages consume (`origin`, `confidence`, `severity`, `elimination_reason`, extras). **Constraint:** the join must NOT make merge-injected `agent` deterministically visible to the filter — withhold/strip `agent` at the verify boundary so today's effective semantics (in which `agent` survives the echo only stochastically) are preserved until the cross-dimension consolidation redesign issue (#22) lands. See the sequencing note under Out of scope and the measured evidence below.
2. **trustSlice adapts.** Receipt (nonce/sha/n_in) unchanged; the count guard becomes delta-id coverage; the `elimination_reason` content-fidelity stamp check is preserved or structurally subsumed.
3. **Per-slice degradation.** An untrusted/failed slice degrades only its own findings to `origin=unknown`; trusted slices keep verified output; remaining slices still dispatch. Never-drop/never-fabricate is unchanged. Optionally one deterministic re-dispatch of a failed slice before degrading.
[...]

PR #71, opened 2026-07-28
fix(verify): degrade only the failed slice, and retry it once first (#54, #25 PR1)

### LINKED ISSUE #54: Verify executor receipt nonce mismatch degrades an entire run's findings, not just the untrusted slice
## Problem / Goal

`verifyStage` (`workflows/src/stages.js`) dispatches one `executor` agent per verify slice, sequentially, and trusts each slice's receipt independently via `trustSlice`. On ai-code-review-evaluation/discourse-graphite#6 in today's 3-PR functional smoke (`bench/workspace/runs/smoke-20260727-205454-f99d948`, commit `f99d948`, 2026-07-27), the executor's receipt failed nonce validation (`receipt nonce mismatch (got missing, expected fb2b1ded5014dde8.0)`) on slice 0 — and because this pipeline has only one slice for this run, that single untrusted slice degraded **every finding in the run** (all 16) to `origin=unknown`, skipping surfaced-vs-introduced classification entirely for the whole PR.

The mechanism, read directly from `workflows/src/stages.js`: the per-slice nonce is derived as `` `${nonce}.${i}` `` (line 493, "so two equal-length slices can never satisfy each other's receipt"); `trustSlice` rejects the slice when the executor's echoed nonce doesn't match (line 651: `if (r.nonce !== nonce) return { ok: false, reason: ... }`); and the dispatch loop (lines 488-514) breaks out entirely on the first untrusted slice (`for (...; failureReason === null; ...)`, `if (!trust.ok) { failureReason = trust.reason; break; }`), after which `unverifiedResult(findings, failureReason)` re-emits the **original, full findings array** — not just the untrusted slice's share of it — with `origin='unknown'` on every entry. A run with N slices, where only the first is untrusted, loses verified classification for every finding from every slice, including ones that would have verified cleanly.

Cost of leaving this alone: any transient executor hiccup (a dropped nonce echo, a timeout, a retry that races the wrong receipt) on any single slice silently degrades an entire PR's verification, with no proportionality between the size of the fault and the size of the damage. Today's smoke shows this is not a hypothetical — it fired live in a routine 3-PR sample.

Goal: an untrusted or failed slice degrades only its own share of findings to `origin=unknown`; findings from slices that verified cleanly keep their verified classification regardless of another slice's fate.

## Requirements

1. When one slice's receipt fails trust (nonce mismatch, executor throw, or any other `trustSlice` rejection), only the findings dispatched in *that* slice are marked `origin=unknown` / `verified=false`. Findings from every other slice that passed `trustSlice` retain their verified classification and surfaced/introduced status.
2. Findings are never dropped and success is never fabricated for the failed slice's share — this is the existing never-drop/never-fabricate contract (`stages.js`'s `unverifiedResult` comment) and must hold identically at the per-slice granularity.
3. The rest of the slice loop continues dispatching remaining slices after one slice fails, rather than the current global `break` — a fault in slice 0 must not prevent slices 1..N from ever being attempted.
4. `bench/runner/check.py`'s G3 `origin=unknown` gate (sound, unaffected by Issue 1's fix) should show partial degradation proportional to the actual fault footprint on a future run of this kind, rather than 100% of a PR's findings.

## Evidence

The live failure, read from the collected `workflows/wf_089793e7-ca8.json` record's `result.gaps` (2026-07-27):

```
"verify: UNVERIFIED — slice 0: receipt nonce mismatch (got missing, expected
fb2b1ded5014dde8.0); all 16 finding(s) marked origin=unknown,
surfaced-classification skipped"
```

Confirmed independently via the checker's own findings-file scan on the same smoke run:

```
$ python3 bench/run.py --check smoke-20260727-205454-f99d948
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6:code-gauntlet-findings-267d8be1.json[0]: origin=unknown (verify/slice degrade)
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6:code-gauntlet-findings-267d8be1.json[1]: origin=unknown (verify/slice degrade)
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6:code-gauntlet-findings-267d8be1.json[2]: origin=unknown (verify/slice degrade)
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6:code-gauntlet-findings-267d8be1.json[3]: origin=unknown (verify/slice degrade)
  FAIL: pr-ai-code-review-evaluation-discourse-graphite-6:code-gauntlet-findings-267d8be1.json[4]: origin=unknown (verify/slice degrade)
```
(This particular persisted findings file has 5 entries, from a distinct write attempt in the same PR directory, all showing `origin=unknown` — consistent with the 16-finding gap message describing the run's actual final attempt.)

The mechanism, verbatim from `workflows/src/stages.js` (2026-07-27):

```
stages.js:490-493 (per-slice nonce derivation):
    // Per-slice nonce: derive `${nonce}.${i}` so two equal-length slices can never
    // satisfy each other's receipt (a slice-confusion / replay defense on top of the
    // base nonce). The base nonce is charset-validated at the args waist (args.js).
    const sliceNonce = `${nonce}.${i}`;

stages.js:651 (trustSlice rejection):
  if (r.nonce !== nonce) return { ok: false, reason: `receipt nonce mismatch (got ${r.nonce == null ? 'missing' : r.nonce}, expected ${nonce})` };

stages.js:487-514 (the loop, global break on first untrusted slice):
  for (let i = 0; i < slices.length && failureReason === null; i += 1) {
    ...
    const trust = trustSlice(env, { nonce: sliceNonce, headShaShort, n: slice.length });
    if (!trust.ok) {
      failureReason = `slice ${i}: ${trust.reason}`;
      break;
    }
    const verified = env.result && Array.isArray(env.result.verified) ? env.result.verified : [];
    verifiedOut.push(...verified);
  }
  if (failureReason !== null) return unverifiedResult(findings, failureReason);
```

`unverifiedResult(findings, ...)` is called with `findings` — the full, original array bound at the top of the function — not `verifiedOut` (the partial results already collected from trusted slices before the break). Every finding in the original array, regardless of which slice produced it, is re-emitted with `origin='unknown'`.

- This pairs naturally with issue #25's requirement 3 ("Per-slice degradation. An untrusted/failed slice degrades only its own findings to origin=unknown; trusted slices keep verified output; remaining slices still dispatch. Never-drop/never-fabricate is unchanged. Optionally one deterministic re-dispatch of a failed slice before degrading.") — the fix described there is the same fix this issue asks for, on the same code path. Recommend landing #25's PR1 (per issue #25's own decomposition) and this issue as one branch rather than two.

## Out of scope

- The delta-echo redesign (#25 requirement 1) that shrinks what the executor echoes — this issue is scoped to the *degradation blast radius* when a slice fails, not to what a trusted slice transmits.
- The report-banner visibility requirement (#25 requirement 7) for surfacing a degraded run — a real concern, but a separate piece of #25's scope.
- Issue 2's writer-transcription defect (also present on this same PR, on a different agent/boundary) — unrelated mechanism, filed separately.

## Verification

Tier: **suites-only**, gated by the same dual-runtime golden-fixture discipline #25 already specifies for this surface (`VERIFY_SCHEMA` / `trustSlice` parity fixtures, deterministic unit tests for per-slice degradation). A functional-smoke re-run against a multi-slice PR with one deliberately-faulted slice is the natural live confirmation once landed, but is not required to land the fix — the deterministic unit tests are the gate.

leehopper:
Closes #54. Lands **PR1 of #25** ("per-slice degradation + optional single retry — pure JS + tests"), per #54's own recommendation to land the two as one branch.

## The defect

`verifyStage` dispatched one `executor` per slice sequentially, and the **first** untrusted slice set `failureReason`, `break`ed the loop, and called `unverifiedResult(findings, …)` — with `findings`, the full original array, not `verifiedOut`. So one fault degraded:

- findings from slices that had already verified cleanly, and
- findings from slices that were never dispatched at all.

There was no proportionality between the size of the fault and the size of the damage.

Live on 2026-07-27 (`smoke-20260727-205454-f99d948`, `ai-code-review-evaluation/discourse-graphite#6`):

```
verify: UNVERIFIED — slice 0: receipt nonce mismatch (got missing, expected
fb2b1ded5014dde8.0); all 16 finding(s) marked origin=unknown,
surfaced-classification skipped
```

A dropped nonce echo cost all 16 findings their surfaced-vs-introduced classification.

## What changed

**Per-slice degradation** (#54 req 1–3, #25 req 3). An untrusted slice degrades only its own findings; the loop never breaks, so a fault in slice 0 no longer prevents slices 1..N from being attempted.

**Exactly one deterministic re-dispatch before degrading.** This is the part that would actually have saved the incident above — that PR had *one* slice, so per-slice degradation alone changes nothing for it. Same shape as `writeArtifactsDerived`'s single retry: the executor is a sampled agent, not a function, so a second dispatch is a fresh sample. It retries **uniformly** where the persist path classifies, because no verify failure class proves "our dispatch never reached the script" — a dropped receipt echo, a truncated result body, a fabricated elimination and a schema-retry throw are all consistent with an executor that ran and came back wrong.

**The retry carries a distinct nonce** (`{nonce}.{i}` → `{nonce}.{i}.r1`). Reusing the slice nonce would move the confusion `trustSlice` defends against from *space* (two equal-length slices satisfying each other's receipts) into *time* (attempt 2 satisfied by a replay of attempt 1's receipt) — and a fresh receipt is exactly what the re-dispatch exists to obtain. Surfaced by the recon pass, not in either issue.

**The nonce is computed once and threaded** into `verifyPrompt(inp, i, sliceNonce)` → `verifyCommand(inp, i, sliceNonce)`. It used to be derived independently at `stages.js:581` and `:772`, silently requiring two formulas to stay identical. An attempt-varying nonce turns that latent duplication into a live bug, so the duplication is removed rather than documented.

**A failed slice-input writer group degrades only the slices it carried.** `materializeVerifySlices` returns a `sliceIndex -> {group, reason}` Map instead of one first-failure reason; slices from healthy groups still reach the executor. Attribution is keyed on the slice-input *path*, not on a group's position, so it cannot be invalidated by a future chunker change.

**Gaps state their own blast radius.** One per degraded slice, in slice-index order: `verify: UNVERIFIED — slice 1: <reason>; 2 of 16 finding(s) marked origin=unknown, …`. A slice recovered by its retry emits a `verify-slice-retry:` disclosure gap carrying **no** `UNVERIFIED` token — nothing degraded. `verified` is true only when zero slices degraded.

**Never-drop/never-fabricate is unchanged**, now at slice granularity — against the failure classes `trustSlice` detects. Every finding leaves the stage either as its slice's trusted verified output or as itself with `origin='unknown'`. That qualifier is not boilerplate; see the disclosed gap below. Output stays in strict slice-index order because `applyChallenges`' stable sort breaks ties by array position.

**Guard math.** `worstCaseAgentCount` and `coarsenLimits` scale the verify term by `VERIFY_ATTEMPTS_PER_SLICE`, since every slice can independently take its retry. The constant is exported so the dispatch-count test asserts against it rather than a copied literal. The other terms stay nominal (report/writer segmentation, the persist retry) and the comment now says so rather than implying the whole function is an exact bound.

## #54 requirement 4 — no `check.py` change

G3 already counts `origin=unknown` **per finding**, so its stat becomes proportional to the actual fault footprint for free. Its verdict stays fail-on-any-occurrence, which is correct and unchanged.

## Scope calls

- **Per-slice degradation was extended to the writer boundary**, which #25's decomposition nominally assigns to PR3. PR3 owns the *content proof*; this is the *degradation blast radius*, the same defect on the same function. Leaving it would half-fix #54's stated goal. No conflict with PR3's echo-shape change.
- **Not in scope, unchanged:** the delta-echo redesign (#25 req 1), the report degradation banner (#25 req 7), and the bucket relabel (#25 req 8).

## Considered and declined

- **Capping the gap list.** A fully-degraded run now emits one gap per slice rather than one. Production `verifySliceSize` is 200 (`phase2-triage.md:356`), so realistic runs have 1 slice (≤200 findings) to 5 (1000 findings) — a handful of gaps. A cap would risk hiding the diagnostic that makes a degraded run debuggable, which is what this change is for.
- **A circuit breaker on systemic failure.** A wholly broken executor now costs up to 2N sequential dispatches where the old fail-fast cost 1. That is exactly what #54 req 3 mandates ("a fault in slice 0 must not prevent slices 1..N from ever being attempted"), so it is a deliberate tradeoff, not a regression to fix here. Filed as #72 for the adjacent fact that `limits.verifySliceSize` is unclamped at the args waist, which sets that N.

## Verification

Tier: **suites-only**, per #54's Verification section. All green:

- `node --test workflows/test/*.test.js` — **440 pass** (was 433)
- `python -m pytest tests/ -q` — **812 pass**, 301 subtests (includes `test_bundle_fresh.py`: the committed bundle is byte-identical to a fresh `node workflows/build.js`)
- `python -m pytest bench/tests/ -q` — **529 pass**, 22 subtests

New/changed coverage: per-slice degradation with mixed trusted/degraded slices; retry-recovers; retry fires exactly once; the retry's distinct nonce **and** rejection of a replayed attempt-1 receipt; a trusted slice never re-dispatched; the loop continuing past a failure; executor-throw isolation; per-writer-group isolation (with a genuine multi-group fixture, and a healthy group's slices proven to reach the executor); all-slices-degraded; and gap order + `k of n` proportionality. The two `stages_latency.test.js` D2.3 tests that encoded first-failure-wins were rewritten to pin what #38 actually cared about — deterministic ordering — rather than the superseded semantics.

An adversarial review pass (#25 process directive 4) ran four independent lenses — correctness/edge-cases, trust-boundary/replay, blast-radius/resource, and contract/test drift — each raised finding independently refuted by a separate agent. Seven findings; six refuted. Two of those six nonetheless pointed at real staleness in pre-existing text (`verifyStage`'s "dispatches ONE executor agent per slice" docstring and four "whole set UNVERIFIED" test titles); both are fixed in commit 2.

**One finding survived, and it is disclosed rather than fixed here.** See below.

## Disclosed gap: `trustSlice` binds a slice's shape, not its content

`trustSlice` validates a slice's nonce, sha, `n_in` and the `verified+eliminated` count sum — but never that the findings in the envelope are the ones that slice dispatched. `verifySliceSize` is a constant, so most slices in a run share a length, and an envelope carrying a **sibling slice's** findings clears every guard. The real slice's findings are then **dropped outright — not degraded to `origin='unknown'`** — and the run still reports `verified=true`.

Reproduced end-to-end against this branch: 4 findings in two size-2 slices; slice 0's first attempt fails honestly and its retry answers with slice 1's content. Output is `[F2,F3,F2,F3]`, `verified: true`, no `UNVERIFIED` gap.

**The hole predates per-slice degradation** and `trustSlice` is byte-for-byte unchanged by this PR. What this PR adds is exactly one dispatch per failed slice on which it can be hit: before the retry, a slice whose first attempt failed honestly could only reach the conservative degrade. Net exposure still falls, because the common case went from "one fault costs every slice its classification" to "one fault costs one slice".

**Why it is not fixed here.** The fix is an id-set binding — precisely **#25 requirement 2**, "the count guard becomes delta-id coverage" — and it changes acceptance on the **trusted** path: an echo that merely mangles an id character would begin degrading runs that pass today. That is a findings-content change, so it needs #25's measurement tier, and it would invalidate the claim that justifies this PR's suites-only tier (the error-path parts "can only prevent verify-coverage loss or surface it, never change findings content"). Escalated on #25 with the PoC.

## Second adversarial pass (full diff, including the disclosure commits)

The first pass reviewed commit 1 only. A second ran over the complete diff with three fresh lenses — the never-reviewed doc/disclosure commits, downstream interactions nobody had looked at, and whether the deferral decision itself holds up. Six findings, one survived.

**Survivor, filed as #73 rather than fixed here.** Per-slice degradation makes a single run's findings array mix script-verified and degraded findings for the first time — before this PR it was homogeneous, all-verified or all-degraded. `detectDisagreement` groups by `(file, line_bucket)` and `rankFindings` orders by `(severity, -confidence)`; neither inspects `origin`. So a finding whose slice failed **both** attempts can take a `+10` consensus boost from a verified neighbour and out-rank a genuinely verified finding, displacing it past `challengeCap` into the undelivered bucket. Reproduced against the real exports: ranked `B=80 > A=75 > C=74`, and `C` — the only one actually fact-checked — is the one dropped.

It is filed rather than fixed for the same reason as the `trustSlice` gap: the fix lands in `detectDisagreement`/`rankFindings`, parity-locked dual-runtime transforms whose change alters delivery selection and needs the measurement tier. And this PR remains a net improvement on that axis — before it, a partially-degraded run delivered *nothing* verified.

Of the five refuted findings, two produced real fixes: the never-drop qualifier said "against every failure class **trustSlice** detects" when two of the four classes listed never reach `trustSlice` (commit 5), and the `trustSlice` PoC existed only as prose, so it is now a runnable script on #25 along with the note that **no bench or CI gate can currently detect that failure** — G3 keys on `origin=unknown`, and a content-swapped slice looks clean.

**CLAUDE.md.** Commit 4 reverts the 13-line section this PR originally added — 4,879 bytes, ~1,220 tokens on every session. Anthropic's test is "would removing this cause Claude to make mistakes?"; all 13 lines fail it, since every claim was already stated in a comment at the site where it matters. `CLAUDE.md` is now byte-identical to `main`. The measurement (115 lines but 23,982 bytes / ~6k tokens, averaging 224 bytes per line — the "under 200 lines" heuristic reads as passing while the real cost is triple) and the `.claude/rules/` path-scoped decomposition design went to #46, the open context-engineering audit.

## Deferred, all captured

| What | Where |
| --- | --- |
| `trustSlice` binds shape, not content | #25 req 2, with runnable PoC + detection gap |
| Mixed verified/degraded consensus & ranking | #73 |
| `limits.verifySliceSize` unclamped at the args waist | #72 |
| `CLAUDE.md` rightsizing | #46 |

Commit 3 therefore changes no behavior: it strips the unqualified never-drop assertion from `verifyStage`'s docstring, CLAUDE.md and `validation-pipeline.md`, and adds a `KNOWN GAP` block at `trustSlice` naming the mechanism and its owner. Shipping a guarantee a PoC breaks is the silent failure this repo exists to avoid.

https://claude.ai/code/session_01MoJGv7nzXjQrw9jf7ZmmtZ

---

> [!NOTE]
> **Medium Risk**
> Changes core verify orchestration (degradation blast radius, retries, agent fan-out math) that directly affects finding `origin` and downstream filter/challenge; behavior is heavily tested but mixed verified/degraded arrays expose known ranking gaps (#73).
> 
> **Overview**
> **Verify no longer fails the whole run on the first bad slice.** `verifyStage` keeps processing slices in index order, marks only failed slices' findings `origin='unknown'`, and sets `verified=false` only when at least one slice degraded. Gaps are per slice with a `k of n` blast radius instead of one message for the entire set.
> 
> **Each slice gets one retry before degrading.** `VERIFY_ATTEMPTS_PER_SLICE` (2) drives `verifySliceWithRetry`, a second dispatch with nonce `{nonce}.{i}.r1`, and disclosure gaps when retry succeeds. `verifyCommand` / `verifyPrompt` take the slice nonce as an argument so argv and `trustSlice` stay aligned. `worstCaseAgentCount` / `coarsenLimits` multiply the verify term by that constant.
> 
> **Slice-input writer failures are scoped to affected slices.** `materializeVerifySlices` returns a per-slice failure map instead of aborting the whole stage; healthy writer groups still materialize inputs for their slices.
> 
> Docs (`phase2-triage`, `phase3-dispatch`, `validation-pipeline`) and tests/mocks are updated for per-slice behavior, retry labels, multi-slice receipts, and mixed-origin end-to-end runs. Comments document an unchanged `trustSlice` content-binding gap (deferred to #25).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2d9794a33fb33ed2682384e91352d0634c482b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


workflows/src/stages.js
```diff
@@ -565,64 +603,137 @@ export async function verifyStage(ctx, input) {
...
+// One slice's share of the UNVERIFIED degradation: its ORIGINAL findings re-emitted with
+// origin='unknown' (surfaced-classification skipped). Nothing is dropped and nothing is
+// upgraded. Numeric-string fields are pinned here for the same reason they are pinned on
+// the slice-input path: the trusted path returns the script's re-scored numbers, but this
+// path re-emits discovery-shaped findings whose confidence is the schema's numeric STRING
+// ("85") — leaked downstream, the filter's consensus `+` boost concatenates ("85" + 10 ->
+// "8510" -> clamped to 100).
+function degradedSlice(slice) {
+  return slice.map((f) => ({ ...pinNumericFields(f), origin: 'unknown' }));
 }
```
leehopper, line 660: **💡 [LOW] No full-pipeline (runWith) test exercises a partially-degraded verify stage**

verifyStage's new per-slice retry/degrade behavior is thoroughly unit-tested in isolation (workflows/test/stages_verify.test.js), but no test in workflows/test/pipeline_run.test.js drives runWith end-to-end with a slice that degrades (untrusted receipt, writer-group failure, or executor throw) to confirm the mixed origin='new'/origin='unknown' finding array flows correctly through Validate, Filter, Challenge, report generation, and persistence, and that stats.verified/out.gaps surface correctly at the top level.

**Suggested fix:**
Extend makeCtx with an opts hook (e.g. opts.verifySliceFailIndex or reusing agentThrowLabel semantics for 'verify-slice-N') so a pipeline_run.test.js test can force one slice to degrade, then assert: findings from the failed slice keep origin='unknown' in the final output, findings from healthy slices are unaffected, out.stats.verified===false, out.gaps contains the per-slice UNVERIFIED gap, and downstream stages (validate/filter/challenge/persist) complete without error on the mixed-origin array.

workflows/src/stages.js
```diff
@@ -693,14 +812,21 @@ async function materializeVerifySlices(c, inp, slices, policy) {
...
+  const failed = new Map();
   for (let g = 0; g < groups.length; g += 1) {
     const r = results[g];
     // A null member is unreachable while the thunks above swallow their own throws; it is
     // still handled so a platform-side null can never be mistaken for a successful write.
-    if (!r) return { ok: false, reason: `slice-input writer group ${g} produced no result` };
-    if (r.ok !== true) return { ok: false, reason: r.reason };
+    const reason = !r
+      ? `slice-input writer group ${g} produced no result`
+      : (r.ok === true ? null : r.reason);
```
leehopper: **💡 [LOW] Nested ternary computing a writer group's failure reason**

materializeVerifySlices computes `reason` with a ternary whose false-branch is itself a ternary (missing-result vs ok-but-failed vs succeeded), then immediately branches on the sentinel value null to decide whether to continue. An if/else-if chain assigning `reason` (defaulting to null) would read top-to-bottom instead of requiring the reader to unwind two nested conditions to find the success case.

**Suggested fix:**
Before: const reason = !r ? `...produced no result` : (r.ok === true ? null : r.reason); if (reason === null) continue; After: let reason = null; if (!r) reason = `slice-input writer group ${g} produced no result`; else if (r.ok !== true) reason = r.reason; if (reason === null) continue;

**Cited rule:** Avoid nested ternary operators (reviewer instructions); no direct CLAUDE.md project rule on this specific pattern.

leehopper (commented):
## Code Gauntlet Review

This PR moves the Verify stage from all-or-nothing degradation to per-slice degradation with a single deterministic retry, so a single dropped nonce/receipt no longer marks every finding UNVERIFIED. 2 high-confidence findings survived verification and blind challenge (0 unverified/pipeline-degraded). Both are low-severity: a missing end-to-end test for the new partially-degraded verify path, and a nested-ternary readability nit in `materializeVerifySlices`. See inline comments for details.

---
Generated by code-gauntlet | Reviewed up to: aec331138e79d9b239779bfd4c4612b4e29c482f

Issue #75, opened 2026-07-28
verifyCommand splits its argv on spaces — a repo path with a space breaks the verify stage


leehopper:
## Problem

`verifyCommand` (`workflows/src/stages.js:912-925`) assembles the verify executor's command by pushing argv tokens into an array and then joining them with a single space:

```js
if (v.diffPath) parts.push('--diff-file', v.diffPath);
return parts.join(' ');
```

No token is quoted. Any waist path containing a space therefore splits into two argv words when the executor agent runs the command. The exposed fields are the ones interpolated into that command — `diffPath`, and the `inputPathBase`/`outputPathBase` under `verify` (both derived from `{output_dir}`).

This fires for an ordinary checkout under a path with a space, e.g. `/Users/foo/My Documents/repo/.code-gauntlet/code-gauntlet-diff-abc1234.patch`. `verify_findings.py` then receives `--diff-file /Users/foo/My` and the slice degrades.

## Why it was not fixed in #27

#27 added waist type-checks for the path-bearing fields (`repoRoot`, `outputDir`, `headShaShort`, `diffPath`) and deliberately stopped at "non-empty string, no control characters", plus "no whitespace" for `headShaShort` alone — because `headShaShort` is `git rev-parse --short=8 HEAD` and can never legitimately contain a space, whereas a *path* can. Rejecting a space in `diffPath`/`outputDir` at the waist would newly reject a valid `argsVersion:1` waist, which #27's requirement 2 forbids.

The right fix is at the command-construction site, not the waist.

## Suggested fix

Quote each interpolated path token in `verifyCommand`. Note the constraint from CLAUDE.md's AST-safe emission rules — the command has to stay parseable by the tree-sitter-bash AST checker that gates sandbox auto-approval, so whatever quoting is chosen needs to be verified against that path, not just against a shell. `stages.js:2236` already references the same "no env prefix, or shell operators" constraint for a sibling command, so there is a precedent to follow.

Worth a test with a fixture path containing a space, asserting the resulting command still names one file.

## Provenance

Found while reviewing the waist path-field checks for #27. Pre-existing on `main`; not introduced or worsened by that change.

Issue #74, opened 2026-07-28
build.js ORDER can silently drop a new workflows/src module from the bundle


leehopper:
## Problem

`workflows/build.js` concatenates sources in a pinned literal `ORDER` array, and `present()` returns the **intersection** of `ORDER` and what is on disk:

```js
function present() {
  const found = new Set(readdirSync(SRC).filter((f) => f.endsWith('.js')));
  return ORDER.filter((f) => found.has(f));
}
```

A new file added to `workflows/src/` but not added to `ORDER` is therefore **silently excluded from the bundle**. Nothing catches it:

- `tests/test_bundle_fresh.py` only checks properties of the already-built bundle (byte-identity to a fresh build, import-freedom, single `export`, compilability, trailing `return await run(`). A bundle missing a module is byte-identical to a fresh build of the same broken `ORDER`, so the freshness test passes.
- `workflows/test/build.test.js` covers the collision detector, not `ORDER` completeness.

The failure mode is the nasty kind: unit tests import from `../src/<file>.js` directly and stay green, while the shipped bundle is missing the module. The break surfaces on the first live dispatch as `X is not defined` — the same class as the `structuredClone` crash already recorded in CLAUDE.md.

## Suggested fix

A test asserting `set(ORDER) == set(f for f in listdir(SRC) if f.endswith('.js'))` — equality, not superset, so a deleted module also fails loudly rather than lingering in `ORDER`. Either side (`tests/test_bundle_fresh.py` or `workflows/test/build.test.js`) works; the Python suite already shells out to `node workflows/build.js`, so it has both halves on hand.

## Provenance

Surfaced while mapping the build surface for #27 (which deliberately put its new code in the existing `args.js` to avoid exactly this trap). Not a defect in any shipped code today — every module currently in `src/` is listed — so this is a guard against a future addition, not a live bug.

Issue #24, opened 2026-07-23
Move scope decision and review config into the args waist: deterministic agentFlags derivation, in-pipeline REVIEW.md parse, canonical limits defaults


leehopper:
## Problem / Goal

Config and decisions the pipeline consumes should be produced by code the pipeline owns, fed through the args waist — not re-derived by orchestrator prose each run. Three prompt-adherence surfaces currently live in SKILL.md instructions: (a) the light/full scope decision — the orchestrator classifies risk *and* stamps `agentFlags`; (b) REVIEW.md parsing — the orchestrator hand-assembles `reviewConfig`/`exclusionPatterns` even though the pipeline already ships text-based parsers built for the no-disk runtime; (c) the four numeric limits — documented only as a prompt-copied table, with in-code absent-value fallbacks that silently differ from the benchmarked values.

Each surface has bitten or nearly bitten: the M4 `agentFlags: {}` stamp after a light decision, the PR-310 live crash from hand-assembled ignore entries, and the SKILL.md warning that a hand-pinned threshold default silently raises the non-security bar from 55 to 70. Prefer moving behavior into deterministic workflow-runtime code over prompt tuning.

## Requirements

1. The per-file risk classification (already produced in Phase 2e for the context file) travels through the args waist; `discover()` derives `agentFlags` from (riskTable, changedLines, trivialScope) in tested JS. The orchestrator's only scope job becomes producing the risk table.
2. The waist schema-validates the risk-table input: malformed or missing table fails loud, never silently resolves to full scope.
3. The risk table stays orchestrator-produced, so its shape is constrained at the waist; the resolved scope decision is echoed in the compact return (e.g. `resolvedPolicy`/`stats`) so adherence is verifiable post-hoc; the trivialScope env value enters via the waist (no `process.env` in workflow JS per CLAUDE.md).
4. Single authority: the orchestrator-stamped `agentFlags` path and SKILL.md's prompt-level pre-dispatch check are retired, not left as a second half-authority — fail loud on ambiguity.
5. The scope gate is behavior-preserving: the same decision, made deterministically; no recall/noise movement expected on non-trivial PRs.
6. The waist carries raw REVIEW.md text (and exclusions text); the pipeline calls its own `parseReviewMd`/`loadExclusions`. The orchestrator parse instructions, the threshold-default-pinning warning, and the V3.1 shape-guards for hand-assembled `reviewConfig` become removable.
[...]

PR #76, opened 2026-07-28
fix(args): make entry-guard recovery copy-paste, not inference (#27)

### LINKED ISSUE #27: Make entry-guard recovery copy-paste for naked direct Workflow invocations
## Problem / Goal

Twice now — including once after the PR-310 L1 guard (finding L1 from the PR-310 live run) shipped — a session's first move on a fresh review request was a direct `Workflow({name: 'code-gauntlet:code-gauntlet'})` call with the raw PR URL as args, skipping the skill entirely. The 3.1 entry guard fires and the session recovers, but recovery today requires *inference* (parse the prose, realize the skill must be loaded, reconstruct the invocation) and the failure surfaces as a thrown JS stack (`at parseEntryArgs (workflow.js:1862:11)`) — misleading noise implying a pipeline bug rather than a caller error.

Goal: recovery from the entry guard is copy-paste rather than inference, and the failure surfaces without stack-trace noise. Error-path ergonomics only; the guard's redirect behavior, which worked as designed, must not change.

## Requirements

1. When raw args look like a PR URL or PR number, the guard's message states the exact next invocation to run — e.g. `Skill("code-gauntlet:code-gauntlet", args="<the PR reference>")` — echoing back the caller's own PR reference. Classification must be cheap, deterministic, and unit-tested.
2. Error-path-only and inert to review behavior — same inertness class as the original L1 fix. The guard must still reject every non-waist input it rejects today; no valid `argsVersion:1` waist object may be newly rejected.
3. Prefer returning the structured `{ok:false, error, phaseReached:'args', failingPhase:'args'}` envelope over throwing on this path — **contingent** on first verifying that a top-level envelope return surfaces properly in the failed-workflow UI. If it does not, keep the throw and improve only the message.
4. Candidate direction 3 (below) depends on unverified platform rendering behavior — validate the rendering behavior against platform docs/issues before building on it.
5. Args-absent is distinguished from args-malformed. `parseEntryArgs(undefined)` currently returns undefined silently, so a dropped `args` surfaces only as a generic cascade (`invalid args: args is not an object; unsupported argsVersion undefined; missing required field: mode; ...`) that never says the args never arrived, nor how to recover. The absent case must say so explicitly and name the skill invocation.
6. `validateArgs` type-checks the path-bearing waist fields it currently only presence-checks (`repoRoot`, `outputDir`, `headShaShort`, `diffPath` — `args.js:12` REQUIRED tests `=== undefined` only; `nonce` already has a non-empty check). A present-but-garbage value today yields a junk interpolated context path (`stages.js:1432`, consumed by every discovery prompt at `stages.js:120,359,864,1169`) instead of failing at the waist before any paid dispatch.
7. Honest residual, stated in the closing PR: all directions improve the error path; none deterministically prevents the naked call. Per the D22 platform fact (a platform fact established during V3.1 verification, internal ref D22), name-based plugin workflow resolution plus the CLI-synthesized invocation surface make the direct call attractive before the skill loads — that root cause is platform-side and out of plugin control.

## Evidence

Field report 2026-07-23, external developer, first 3.1 deep review — `liatrio-forge/emerald-grove-pet-clinic-yujie-wu` PR 45. The session's first move was `Workflow({name: 'code-gauntlet:code-gauntlet'})` with the raw PR URL as args. The guard fired as designed:

```
Dynamic workflow "code-gauntlet v3 pipeline: ..." failed: Error: args must be the
assembled argsVersion:1 waist object — do not invoke this workflow directly; run the
code-gauntlet skill (Phases 1-2 build the args).
Got: https://github.com/liatrio-forge/emerald-grove-pet-clinic-yujie-wu/pull/45
    at parseEntryArgs (workflow.js:1862:11)
```

The session read it, loaded the skill, and recovered. This is the second independent occurrence of the pattern after the PR-310 L1 guard; pre-3.1, this exact move was an uncaught native `JSON.parse` throw with zero guidance.

Root cause (D22 platform fact): name-based plugin workflow resolution works, and the CLI-synthesized invocation surface makes the direct call attractive before the skill loads.

1. PR-reference classification in the guard + copy-paste `Skill(...)` recovery line in the message.
2. Structured envelope instead of thrown stack on this path. The `{ok:false, ...}` envelope machinery already exists in the pipeline; the only open question is whether a top-level envelope return surfaces properly in the failed-workflow UI (the verification requirement 3 makes this contingent on).
3. A "do not invoke directly" line in the CLI-synthesized invocation hint via workflow `meta.description` — the change is one string in the workflow meta in `workflows/src/pipeline_entry.js`. Before relying on the hint rendering, the pickup session should first investigate what the observed sessions actually saw that made the direct Workflow call attractive (skill description triggering? CLI invocation-hint wording?). Platform rendering behavior unverified; validate first.

## Out of scope

- Any change to how valid waist-object invocations are parsed or dispatched.
- Attempting to deterministically block direct invocation — platform-side, out of plugin control (see requirement 5).
- Direction 3 ships only as an optional follow-up PR after platform-behavior verification, never bundled into the main PR.
- Coordination, not scope: this item and the #24 item both touch `workflows/src/args.js` and the pipeline entry surface — sequence or rebase those PRs against each other to avoid file contention; the workstreams themselves remain independent.

PR shape: one PR for directions 1–2 (guard message + envelope-vs-throw); optional follow-up for the `meta.description` hint only after verification.

## Verification

Tier: **suites-only.** Measurement policy (ratcheted, 2026-07-23): every item ships behind the always-on deterministic suites (pytest + `node --test`, parity and bundle-freshness included). This item is error-path-only plus meta-string changes, inert to review behavior — unit suites cover it, and no paired measurement is needed beyond suites. New unit tests must cover: PR-URL and PR-number classification (positive and negative cases), exact recovery-message content, and unchanged accept/reject behavior for waist objects and other malformed inputs. Changes landing without paired measurement are covered by the next owner-triggered paired run, which measures the accumulated delta since the last baseline of record (`custom-20260723-102149-381e9ff`, recall 19/30 = 0.6333 / noise 0.2233).

Source: #21 item 7 + comment IC_kwDORvwhZ88AAAABLcZQPg (2026-07-23) — evidence copied inline above; this issue stands alone.

leehopper:
Closes #27.

## What changed

A naked `Workflow({name: 'code-gauntlet:code-gauntlet'})` with a raw PR reference hit the entry guard and got a generic message plus a JS stack. Recovery required *inference*: realize the skill exists, that Phases 1-2 build the args, and how to invoke it. Now the refusal names the caller's own reference back and the exact call to run instead.

The field-report input from the issue:

```
args was a GitHub PR URL (https://github.com/liatrio-forge/emerald-grove-pet-clinic-yujie-wu/pull/45),
not the assembled argsVersion:1 waist object — do not invoke this workflow directly; run the
code-gauntlet skill (Phases 1-2 build the args). Run this instead:
Skill("code-gauntlet:code-gauntlet", args="https://github.com/liatrio-forge/emerald-grove-pet-clinic-yujie-wu/pull/45")
```

Three refusal classes, one wording owner (`entryRefusalMessage`):

| class | trigger | says |
|---|---|---|
| absent | `undefined` / `null` | "args never arrived" + the skill call (requirement 5) |
| review target | a PR/MR URL, `owner/repo#n`, a bare number, `PR 310` / `#310` / `!89` | names the kind, echoes the ref, gives the copy-paste line |
| shape | anything else non-object | the generic form, input echoed |

`classifyReviewTarget` runs on the **raw** value before the parse — `'310'` is valid JSON, so a bare PR number would otherwise parse clean into a generic `args is not an object`.

## Requirement 3: verification failed the contingency — the throw stays

Requirement 3 preferred a returned envelope, *contingent* on verifying it surfaces properly in the failed-workflow UI, with an explicit fallback if not. It does not:

- A returned `{ok:false, ...}` reaches the caller as `<status>completed</status>` — identical to a successful run. Recorded bench runs in this repo show it, and `anthropics/claude-code#66745` reports the same platform behavior as a bug ("makes silent failure invisible"), still open.
- `anthropics/claude-code#66088` adds that a returned value can be truncated in the completion notification.
- A **throw** is what #27's own field report shows rendering as a visible failure with the message verbatim.

The caller most likely to make the naked call is the one least likely to parse a return value. Converting the throw would have traded stack noise for a review that reports success and did nothing. So `pipeline_entry.js` keeps throwing, and the message does the work.

**Two seams, two signals, one message.** `runWith` returns the same refusal in the standard `ok:false` / `failingPhase:'args'` envelope, because its own contract is throw-free — and empirically was not: `runWith(undefined, 'PR 310')` escaped as an uncaught `SyntaxError` from a `JSON.parse` sitting outside its try/catch, despite the function's doc comment promising it "NEVER lets a throw escape". Both signals derive from one `entryRefusalMessage`, and a test pins that they are byte-identical.

## Requirement 4: direction 3 is NOT shipped

No official source establishes where `meta.description` renders. The "shown in the permission dialog" phrasing traces only to an unofficial reconstructed system-prompt document; `meta.whenToUse` appears in no official source. Saved/named workflows do become namespaced slash commands, but this plugin invokes by `scriptPath`, so that surface does not apply. No documented mechanism injects a workflow description into a session's context before it decides to call `Workflow`. **`meta` is untouched.**

## Requirement 6 and two additions

- `validateArgs` now type-checks `repoRoot` / `outputDir` / `headShaShort` / `diffPath`: non-empty string, no control characters, plus no whitespace for `headShaShort` (it reaches the verify executor's `--head-sha` argv, the same hazard `NONCE_RE` exists for). A present-but-garbage value used to render a junk context path into every paid dispatch.
- **Beyond the literal requirements, flagged for your call:** the `validateArgs` cascade gains the referenceless recovery line, so a hand-built `args={}` learns *where* the fields come from rather than only which are missing. A plain object is still accepted by the entry on purpose — a near-miss waist deserves the field-by-field list.
- JSON unwrapping is now one bounded loop shared by both seams. Previously it happened by accident across two hops (entry parse + `runWith` parse), so refusing at the entry would have hard-rejected a valid **double-encoded** waist — a requirement-2 violation caught in adversarial review.

## Honest residual

Every direction here improves the error path; **none deterministically prevents the naked call.** Per the D22 platform fact, name-based plugin workflow resolution plus the CLI-synthesized invocation surface make the direct call attractive before the skill loads. That root cause is platform-side and out of plugin control.

The platform draws no status distinction for a returned refusal, so the payload text is the entire signal a caller gets — which is why the wording, not the mechanism, is the deliverable.

## Verification

Tier: **suites-only**, per the issue. Error-path-only and inert to review behavior.

- `node --test workflows/test/*.test.js` — **529 pass, 0 fail** (new `workflows/test/entry_guard.test.js`, 82 tests)
- `python -m pytest tests/ -q` — **812 passed, 301 subtests passed**, `test_bundle_fresh.py` included
- Bundle rebuilt; committed `pipeline.js` is byte-identical to a fresh build

Adversarial review ran three lenses (contract, classifier correctness, recovery ergonomics). Five findings were applied: an embedded-newline leak in the echo, a trailing-backslash corruption of the copy-paste line, the double-encoded-waist regression, an overpromising `omit args` placeholder (Phase 1 auto-detects the branch's PR first), and the missing recovery line on the `{}` cascade.

## Docs: CLAUDE.md is deliberately untouched

`CLAUDE.md` is byte-identical to `main` on this branch. A block explaining the two-seam design was added and then reverted (`9d138bc`): +2,109 bytes / ~525 tokens on a file loaded into **every** session, for zero new information. An independent claim-by-claim audit found all 11 propositions already stated, several near-verbatim, in comments at the site that owns them — `args.js` (`unwrapWaist`, `refusalFrom`, `entryArgs`, `parseEntryArgs`), `stages.js` (runWith's refusal arm), `pipeline_entry.js`, and `entry_guard.test.js`'s header.

This applies the rule recorded on #46 after the identical near-miss on #71: an addition to `CLAUDE.md` must fail to be derivable from the code **and** must not be adequately stated by a comment at the site. Both bullets failed both halves.

The duplicate had also already drifted — it claimed both signals are built from `entryRefusalMessage`, where both seams actually reach the wording through `entryArgs` → `refusalFrom`. The co-located comment is correct. A second copy does not just cost tokens; it decays.

The one-line `SKILL.md` correction stays: it fixed a statement that was wrong (an `ok:false` is a return, never a throw), rather than adding a standing instruction.

**Also carries the forward-looking half (#46), folded in at the owner's direction.** `tests/test_claude_md_budget.py` pins CLAUDE.md at its measured 23,982 bytes with **no headroom**, plus a pre-commit hook scoped `files: ^CLAUDE\.md$` — the existing local pytest hook is `types: [python]` and never fires on a CLAUDE.md-only commit, which is exactly how the file grows.

The test deliberately does *not* check the semantic rule: "is this derivable from the code, and already stated by a comment at the site" is a judgment call, not a computation, and this repo forbids a guard that is a phrase count or a model prompt doing code's job. It buys the narrower sufficient thing — the judgment can no longer be skipped *silently*. A second test asserts the ceiling never drifts above the real size, since one generous bump is how this kind of guard rots. Verified it fires: appending a single byte fails with the two-part test in the message.

Twice is the reason it exists — #71 and #76, the second in the same session that wrote the rule down, one PR later. A written rule asks an author mid-feature to re-derive a semantic test from memory; that is precisely what failed.

## Coordination

#24 also touches `workflows/src/args.js`; it is open with no PR, so there is no contention today. Whichever lands second should rebase.

Two adjacent issues surfaced and filed rather than fixed here: #74 (`build.js` `ORDER` silently drops an unlisted `src` module) and #75 (`verifyCommand` splits argv on spaces, so a repo path with a space breaks verify — not fixable at the waist without rejecting a valid waist).

https://claude.ai/code/session_0151T3dyGNaVzdv3Jj8Swj9a

---

> [!NOTE]
> **Medium Risk**
> Touches workflow args entry and path/argv validation on a security-sensitive boundary, but changes are error-path-only with broad test coverage and no intended review-behavior change.
> 
> **Overview**
> **Workflow entry guard (#27)** replaces the old JSON-parse failure with shared `entryArgs` / `refusalFrom` logic on both seams: `parseEntryArgs` still **throws** (so naked `Workflow` calls surface as failures), while `runWith` **returns** the same text in a standard `ok:false` envelope. Review-target inputs (PR/MR URLs, `owner/repo#n`, bare numbers, `PR 310`, etc.) get a labeled refusal that echoes the caller’s ref and a copy-paste `Skill("code-gauntlet:code-gauntlet", …)` line; absent args and other non-waist shapes get distinct messages. Bounded JSON unwrapping (up to four layers) keeps valid double-encoded waists working after entry-level refusal.
> 
> **`validateArgs`** adds checks on path-bearing waist fields (`repoRoot`, `outputDir`, `headShaShort`, `diffPath`) and applies `NONCE_RE`-style rules to `headShaShort` for verify argv safety. Hand-built near-miss objects still get field-level errors, now appended with `SKILL_RECOVERY_LINE`.
> 
> **#46 guard:** `tests/test_claude_md_budget.py` pins `CLAUDE.md` at 23,982 bytes (no slack), plus a pre-commit hook on `CLAUDE.md` edits using stdlib `unittest` so CI lint jobs without pytest still run it.
> 
> **Docs:** `skills/code-gauntlet/SKILL.md` narrows mid-run crash wording (thrown error with no return vs `ok:false`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d27d1ce54d661e83264fa64e2aa71d98f89a33b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper (commented):
Code Gauntlet review of this PR found 5 findings, all surfaced as improvement suggestions (no bugs or security issues survived challenge verification). Two cover untested boundary conditions in the entry-guard args parsing (MAX_JSON_UNWRAP depth, PR-shorthand regex alternatives), one flags a duplicated args-refusal envelope shape with no shared factory, one flags dead/unused code (entryRefusalMessage), and one flags a stale line-number citation in a comment. See inline comments for details.

---
Generated by code-gauntlet | Reviewed up to: e8ec380f014fc9aa2e57701cc8e864c26ac6e23c

workflows/src/args.js
```diff
@@ -135,21 +135,269 @@ export function normalizeArgs(raw) {
...
+  for (let depth = 0; depth <= MAX_JSON_UNWRAP; depth++) {
+    if (isPlainObject(value)) return { waist: value, terminal: value };
+    if (typeof value !== 'string' || depth === MAX_JSON_UNWRAP) break;
+    let next;
+    try { next = JSON.parse(value); } catch (e) { break; }
+    value = next;
+  }
+  return { waist: null, terminal: value };
+}
+
```
leehopper, line 243: **🟡 [MEDIUM] Exact MAX_JSON_UNWRAP boundary (4 layers accepted, 5 refused) is never directly tested**

unwrapWaist peels up to MAX_JSON_UNWRAP=4 JSON layers before giving up. Tracing the loop shows a 4-times-JSON.stringify-wrapped valid waist is accepted and a 5-times-wrapped one is refused, but the test suite only exercises 1, 2, and 3 layers as accepted and then jumps to 7 layers (1 initial + 6 more) as refused, leaving the exact accept/refuse boundary unverified.

**Suggested fix:**
Add two tests: parseEntryArgs on a 4-times-JSON.stringify-wrapped `good` waist resolves to `good` (accepted, at the bound), and on a 5-times-wrapped waist throws the refusal (just past the bound).

workflows/src/args.js
```diff
@@ -135,21 +135,269 @@ export function normalizeArgs(raw) {
...
+const MR_URL_RE = new RegExp(`^(?:https?://)?[\\w.-]+/\\S+?/-/merge_requests/(${REVIEW_TARGET_N})(?:/[^\\s?#]*)?(?:\\?[^\\s#]*)?(?:#\\S*)?$`);
+const REPO_REF_RE = new RegExp(`^[^\\s/]+/[^\\s/]+#(${REVIEW_TARGET_N})$`);
+const PR_NUMBER_RE = new RegExp(`^(${REVIEW_TARGET_N})$`);
+const PR_SHORTHAND_HASH_RE = new RegExp(`^[#!](${REVIEW_TARGET_N})$`);
+// Longest alternatives first (`pull request`/`merge request` before `pull`) purely to
+// avoid a redundant backtrack on the common phrase — anchoring makes the order
+// non-load-bearing for correctness. Case-insensitive: "pr 310" is as real a caller typo
+// as "PR 310".
+const PR_SHORTHAND_WORD_RE = new RegExp(`^(?:pull request|merge request|PR|MR|pull)[ ]?[#!]?(${REVIEW_TARGET_N})$`, 'i');
+
```
leehopper, line 175: **🟡 [MEDIUM] PR_SHORTHAND_WORD_RE alternatives "pull request", "merge request", bare "pull", and case-insensitivity are untested**

The shorthand regex accepts five alternatives ("pull request", "merge request", "PR", "MR", "pull") case-insensitively, and the code comment explicitly calls out lowercase "pr 310" as an intended catch, but the positive test cases only exercise the "PR 310" and "MR#7" forms — the "pull request", "merge request", bare "pull", and any lowercase input are never asserted to classify correctly.

**Suggested fix:**
Add positive test cases for 'pull request 45', 'merge request 89', 'pull 45', and a lowercase variant like 'pr 310' to pin every documented alternation branch and the case-insensitive flag.

workflows/src/args.js
```diff
@@ -135,21 +135,269 @@ export function normalizeArgs(raw) {
...
+      ok: false,
+      error: message,
+      phaseReached: 'args',
+      failingPhase: 'args',
+      artifactPaths: {},
+      stats: {},
+      gaps: [message],
+    },
+  };
+}
```
leehopper, line 362: **🟡 [MEDIUM] Args-refusal envelope shape duplicated across two construction sites with no shared factory**

Encapsulation: 6/10, Expression: 5/10, Usefulness: 7/10, Enforcement: 4/10. entryArgs() in args.js and the validateArgs-reject arm in stages.js runWith() each hand-build an object literal with the identical seven-field shape (ok, error, phaseReached, failingPhase, artifactPaths, stats, gaps), and a comment asserts they are 'byte-identical in shape... field for field', but no shared type/factory enforces that; the only test checks one site's output keys against a hardcoded literal array rather than comparing the two construction sites to each other, so a future edit to either literal (e.g. adding a field) can silently desync the two signals' shape without failing any test tied to the other site.

**Suggested fix:**
Extract a single makeArgsRejectEnvelope(message, gaps) factory in args.js (or stages.js) used by both entryArgs() and the validateArgs-reject arm, so the two signals are structurally guaranteed to share a shape rather than relying on a comment plus a test that pins one side's keys as a literal.

workflows/src/args.js
```diff
@@ -135,21 +135,269 @@ export function normalizeArgs(raw) {
...
+
+// entryRefusalMessage(raw) -> string | null. The SINGLE owner of the guard's wording and of
+// its accept/refuse decision. Both seams reach it through `entryArgs` — `parseEntryArgs`
+// throws the message, `runWith` returns it inside the envelope — so the two cannot say
+// different things about the same input (pinned by entry_guard.test.js's "one message, two
+// signals" test). This wrapper is the standalone form, for callers that want only the
+// verdict; `null` means the value may proceed to normalizeArgsReport/validateArgs.
+export function entryRefusalMessage(raw) {
+  return refusalFrom(raw, unwrapWaist(raw));
+}
```
leehopper: **🟡 [MEDIUM] entryRefusalMessage is exported but never called from any production or test code**

entryRefusalMessage(raw) is a new exported wrapper that just calls refusalFrom(raw, unwrapWaist(raw)) — identical unwrap-then-decide logic to what entryArgs already does inline. A repo-wide grep shows no caller anywhere (not parseEntryArgs, not runWith/stages.js, not entry_guard.test.js or args.test.js, not the bundle's own internal callers) — only its own definition and comments reference the name. It is dead code that duplicates entryArgs's unwrap+refusalFrom pairing and risks silently drifting from it if entryArgs's unwrap strategy changes later.

**Suggested fix:**
Remove the unused export (and its bundle twin) unless it is meant as a documented public helper — if so, add at least one caller/test that uses it. Before: export function entryRefusalMessage(raw) { return refusalFrom(raw, unwrapWaist(raw)); } — After: delete the function; entryArgs already performs the identical unwrap+refusalFrom pairing for its only two real callers (parseEntryArgs, runWith).

**Cited rule:** null

workflows/src/args.js
```diff
@@ -162,6 +410,35 @@ export function validateArgs(args) {
   if (args.nonce !== undefined && (typeof args.nonce !== 'string' || !NONCE_RE.test(args.nonce))) {
     errors.push(`invalid nonce: must match ${NONCE_RE} (AST-safe, non-splitting — interpolated into the verify command argv per slice)`);
   }
+  // Path-bearing waist fields (requirement 6, issue #27). repoRoot/outputDir/headShaShort/
+  // diffPath interpolate into the shared-context path
+  // (`${outputDir}/code-gauntlet-context-${headShaShort}.md`, stages.js:2380), which reaches
+  // every discovery prompt, and headShaShort/diffPath also reach the verify executor's argv
+  // (--head-sha, --diff-file) — the same argv-splitting hazard NONCE_RE already guards
```
leehopper, line 412: **💡 [LOW] Stale line-number citation "stages.js:2380" now points at the wrong line**

The path-bearing-field validation comment in validateArgs cites the shared-context path build as living at stages.js:2380, but that line in the current file (`const contextPath = ...`) is now at stages.js:2402. The 22-line discrepancy is caused by this very PR's own edits to runWith in stages.js (two hunks adding 16 and 6 lines respectively before that line), so the citation was accurate for the pre-edit file and was never updated after the author's own change shifted it. The identical stale citation is duplicated verbatim in the generated bundle and in the args.test.js copy of the same comment.

**Suggested fix:**
Update the citation to stages.js:2402 in workflows/src/args.js, and rebuild the bundle / re-sync the duplicate comment in workflows/test/args.test.js so all three copies agree.

PR #77, opened 2026-07-29
fix(wait): replace the Phase 3 sleep/Read poll loop with a blocking awaiter (#26)

### LINKED ISSUE #26: Deterministic workflow-await: replace the Phase 3 sleep/Read poll loop with a blocking await script
## Problem / Goal

Phase 3's wait-for-Workflow protocol is prompt-driven: the orchestrator runs up to ~30 `Bash(sleep 60)` + `Read(<task output file>)` iterations and must, by prompt adherence alone, hold its turn, count iterations, and recognize the terminal `{ ok, ... }` object. Any lapse is fatal in headless mode: a `-p` child that yields its turn is killed at the CLI's 600s ceiling before Phase 8, and the review is silently lost (the `config_echo_mismatch`/no-payload symptom seen in bench children).

Goal: the orchestrator must never depend on prompt adherence to observe the Workflow's terminal result. A single stdlib script blocks until the task output file holds a terminal `{ ok, ... }` object or a bounded timeout, reducing the wait to a handful of deterministic Bash calls. This follows the standing preference for deterministic scripts over prompt tuning.

Provenance: this item is net-new relative to #21 — it comes from the headless-kill/backgrounding analysis, not from a #21 finding.

## Requirements

1. Stdlib-only Python script (per CLAUDE.md scripts rule) taking the task output file path and a per-invocation timeout that stays under the Bash tool's 600s ceiling. On success it prints the terminal compact JSON verbatim; on timeout it prints a machine-readable timeout marker. Exit codes distinguish the two outcomes.
2. Terminal detection is robust both to a partially-written output file (JSON parse failure means "not yet terminal" — keep waiting, never error out) and to the file's actual on-disk format: the terminal `{ ok, ... }` object may be embedded in or appended after other task output rather than being the entire file (SKILL.md says the file "shows" the terminal object), so whole-file `json.loads` must not be the only detection path. The pickup session must characterize the real file format first, and pytest coverage must include a fixture with surrounding non-JSON content.
3. Both copies of the wait protocol are rewritten around the script: the SKILL.md Phase 3 wait protocol (with its :254 headless hard rule) and the duplicate "Wait Protocol (MANDATORY — do not end the turn early)" section in `references/phase3-dispatch.md`. The rewritten protocol is a bounded number of script invocations, never yielding the turn between them, with the timeout marker mapping onto the existing `workflow-timeout` gap flow. No copy of the sleep/Read loop may survive anywhere in the skill or references after the rewrite (a grep-for-`sleep 60`-style check is the testable acceptance form). Also audit `references/headless-mode.md` and SKILL.md:254's headless hard rule for pointer updates.
4. The protocol must not assume interactive vs headless operation; bench `-p` children are the primary beneficiary.

## Evidence

```
skills/code-gauntlet/SKILL.md:214-232 — current MANDATORY protocol:
  Bash(sleep 60) + Read(<task output file>), up to 30 iterations,
  turn-holding is model-discretionary (prompt adherence only).

references/phase3-dispatch.md:26-36 — duplicate of the full poll
  protocol (sleep 60 + Read loop, 30-iteration cap, workflow-timeout
  gap, headless highest-risk note); both copies must move together
  or the old loop stays live.

SKILL.md:254 headless hard rule — a headless `-p` child that yields its
  turn is killed at the CLI's 600s ceiling before Phase 8; no compact
  return, review silently lost (config_echo_mismatch / no-payload symptom).

Bash tool timeout ceiling: 600000 ms — bounds any single blocking script
  invocation to <10 minutes, hence bounded repetition rather than one call.
```

- Poll-with-sleep inside the script at a short interval (context, not prescription).
- Optionally accept a max-total-wait parameter so headless callers can cap the whole protocol in a known invocation count.

The pickup session owns the design (spec-driven, TDD).

## Out of scope

- Any change to the Workflow runtime, pipeline stages, or findings semantics — this touches only the wait/observe path.
- Raising or working around the Bash 600s ceiling itself.
- Bench harness changes; the bench benefits passively via the headless protocol.

## Verification

Measurement tier: **suites-only** — this is an error-path/latency change with no scoring surface touched, so no paired bench is required. Measurement policy (ratcheted, 2026-07-23): every item ships behind the always-on deterministic suites (pytest + `node --test`, parity and bundle-freshness included). Pytest coverage must include terminal detection, partial-JSON tolerance, embedded-terminal-object detection (surrounding non-JSON content), and the timeout marker/exit-code contract. The change is exercised by the next per-sub-release 2–3-PR functional smoke under the ratcheted policy — appropriate here because the failure mode being fixed is a silently lost review in headless `-p` children, which only a live run observes. Changes landing without paired measurement are covered by the next owner-triggered paired run, which measures the accumulated delta since the last baseline of record (`custom-20260723-102149-381e9ff`, recall 19/30 = 0.6333 / noise 0.2233).

Expected PR shape: single PR — `scripts/await_workflow.py` + pytest coverage + wait-protocol rewrite in SKILL.md and `references/phase3-dispatch.md` (plus pointer audit of `references/headless-mode.md`).

leehopper:
## Why?

Closes #26.

Phase 3's wait protocol asked the orchestrator to hold its turn, count to 30, and recognize a terminal `{ ok, ... }` object by eye. Every one of those is a computation, and the one that failed was the judgment call.

Characterizing the real file first changed the design twice.

**The `Workflow` tool result does not contain the task output file path.** It carries a Task ID, a transcript dir and a Run ID — nothing else. `SKILL.md`'s "Take the task output file path from the `Workflow` tool result" described something impossible. That path *and* the compact return appear only in the completion `<task-notification>`, which is delivered **only between turns**. So a session that yields its turn to receive the notification is the session that gets killed, and a session that holds its turn can never be handed it. Neither can wait correctly on its own — which is why the waiting is now done by something that watches the file directly.

**The kill is narrower than this repo has been writing it.** It is a `claude -p` mechanism, not a general turn-end one: a `-p` run blocks on background tasks still running when the main turn ends, but only up to `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS` (default 600000 ms), then prints `Background tasks still running after Ns; terminating.` and exits. `smoke-20260719-190902-a14b4cc` lost all three PRs exactly this way, which is why `bench/runner/invoke.py:515` exports that variable as `"0"`. Interactive sessions are not subject to it — the #38-profiled run yielded its turn for 994.1s and resumed cleanly. The docs now say this accurately instead of repeating a blanket claim.

## What Changed?

**`scripts/await_workflow.py`** (new, stdlib-only) blocks until the task output file holds a terminal result, then branches by exit code:

| exit | meaning |
|---|---|
| 0 | the compact return, on stdout |
| 3 | not terminal yet — stdout carries the literal `next_command` to run |
| 5 | the persisted artifacts landed but the return was never observed |
| 4 | attempts exhausted, or an unexpected failure → `workflow-timeout` gap |
| 2 | argparse's own; the invocation is malformed |

Exactly one compact JSON line on stdout for every wait outcome, including an unexpected exception — the same rule `assemble_artifacts.py` has, for the same reason: an empty stdout is indistinguishable from a dead process. Nothing is written to stderr on the terminal path either, so a caller that merges the streams (the Bash tool does) still sees exactly one line.

**Detection is structural, not a keyword match.** The file is an envelope of `{summary, agentCount, logs, result, workflowProgress, totalTokens, totalToolCalls}` with the return nested at `.result` — measured across 66 real task files, in which `.result` is a dict in 55, a list in 7 and a string in 4, because the Workflow tool serves other pipelines too. A terminal result therefore requires a boolean `ok` **plus** at least one field only the compact return carries. A bare `ok` is not enough: the assemble receipt has one, and mistaking a receipt for the return would hand Phase 8 the wrong object. Both observed return shapes are covered, including the early-failure shape that omits `resolvedPolicy` and `checkpoints` entirely.

**Two independent observables, deliberately.** The task file is the harness's, at a path derived from the Task ID (`$CODE_GAUNTLET_TASKS_DIR` overrides the search). The four persisted artifacts are *ours*, at a path Phase 2 constructs. A harness layout change alone cannot lose a review, and a writer failure alone cannot either. Only artifacts newer than the wait's start count, so a stale set at the same head SHA can never be delivered as this run's.

**Both copies of the protocol moved together** — `SKILL.md` and `references/phase3-dispatch.md` — or the old loop stays live. A pytest greps `skills/` to keep it that way (#26's stated acceptance form). `references/headless-mode.md` gained the pointer.

## Additional Notes

**Issue #26's Requirement 4 is inaccurate and should be corrected rather than silently shipped past.** It states "bench `-p` children are the primary beneficiary." They are not: `bench/runner/invoke.py:515` already sets `CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS="0"` (wait unbounded), so bench children are insulated from the ceiling this fixes. The real beneficiary is an ordinary headless `-p` run on the stock 600000 ms default. R4's first clause — that the protocol must not assume interactive vs headless — *is* satisfied: one protocol, no mode branch. Detail on the issue.

**One spec item was deliberately dropped.** The plan called for adding `await_workflow.py` to CLAUDE.md's covered-scripts list; `tests/test_claude_md_budget.py`'s #46 guard rejected it, and by that guard's own two-part test the guard is right — the fact is derivable from `tests/test_await_workflow.py` existing. Reverted rather than raise the ceiling.

**Three adversarial review rounds ran, and each found real defects in the previous round's fixes.** Round 1: an O(n²) scan (`find_terminal('{'*200000)` took 7.96s; 8 MB would have been hours) and a FIFO target blocking `open()` forever with zero stdout — the exact silent-loss mode this issue exists to close. Round 2: round 1's bounds had over-corrected and were silently dropping valid results two ways. Round 3: three of the four scan bounds still truncated the search without disclosing it, and `emit()` outside the try meant a closed pipe exited 1 with a traceback. Worst adversarial input now measures ~20 ms. Round 1's version looked finished and was not — worth knowing when reviewing the later commits.

**Verified on real data, not only fixtures:** a genuine headless smoke-run task file yields the compact return at exit 0; a bare Task ID resolves to its real path; and the awaiter blocked exactly 120.0s of wall clock against a live in-flight background workflow, exiting 3 with a runnable `next_command`.

Measurement tier is **suites-only** per #26 — an error-path change with no scoring surface touched. Zero non-stdlib imports; parses under the 3.10 grammar (oldest CI target).

- [x] Pipeline tests pass: `python -m pytest tests/ -q` — 928 passed, 301 subtests
- [x] Bench self-tests pass: `python -m pytest bench/tests/ -q` — 529 passed, 22 subtests
- [x] Workflow tests pass: `node --test workflows/test/*.test.js` — 539 passed
- [x] Parity fixtures regenerated with `workflows/test/tools/record_parity.py` if a deterministic transform changed — n/a, no deterministic transform changed
- [x] Docs, skill references, and agent contracts updated if behavior changed

https://claude.ai/code/session_014VRYZyjmAViszmkDE8kgtC

---

> [!NOTE]
> **Medium Risk**
> Changes the mandatory Phase 3 wait path for all reviews; mistakes in terminal detection or exit-code handling could still lose or mis-route results into Phase 8, though behavior is heavily tested and documented.
> 
> **Overview**
> Replaces the Phase 3 **model-driven `sleep 60` + Read loop** with **`scripts/await_workflow.py`**, which holds the turn inside one Bash call, polls the background Workflow task output, and returns **one compact JSON line** on stdout for every outcome.
> 
> The orchestrator now runs the awaiter with the **Task ID** (not a path from the tool result), **`timeout: 600000`**, and optional **`--artifacts-dir` / `--head-sha`**, then branches on **exit code** (`0` terminal return, `3` rerun `next_command`, `4`/`5` `workflow-timeout` degradation) instead of judging file contents itself. Docs in **`SKILL.md`**, **`phase3-dispatch.md`**, and **`headless-mode.md`** were aligned on the Task-ID vs between-turn notification problem and the narrower **`claude -p`** **`CLAUDE_CODE_PRINT_BG_WAIT_CEILING_MS`** kill behavior.
> 
> The script resolves task files under Claude temp dirs (or **`CODE_GAUNTLET_TASKS_DIR`**), detects the pipeline compact return inside the Workflow envelope (boolean **`ok`** plus corroborating fields, not bare receipts), uses bounded embedded JSON scanning and a **four-artifact freshness fallback**, and ships with **`tests/test_await_workflow.py`** (including guards that skills no longer document sleep-based polling).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9876942f2acdd8a95fc13d0265ad7be0b847ea76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/await_workflow.py
```diff
@@ -0,0 +1,913 @@
...
+        return 4
+    try:
+        emit(payload)
+    except OSError:
+        # The payload could not be delivered at all, so there is nothing useful to
+        # print — but the exit code must still be one the caller's branch table
+        # covers. 4 is the degrade-and-disclose family; exit 1 was not documented
+        # anywhere and would read as an interpreter crash.
+        return 4
+    return code
```
leehopper, line 919: **🟡 [MEDIUM] emit() can raise OSError uncaught from the KeyboardInterrupt/Exception error handlers in main()**

emit()'s own docstring promises "Never raises", but on a broken pipe it deliberately redirects stdout to /dev/null and then re-raises the original OSError by design (per its own comment: "let the caller fall through to a documented degrade"). main() only wraps the third call site (emit(payload) for the normal terminal/marker path) in try/except OSError; the two emit({...}) calls inside the except KeyboardInterrupt and except Exception handlers are called unguarded. If the read end of stdout is closed (or otherwise raises OSError) exactly while one of those two error-path emits runs, the OSError propagates out of main() uncaught, producing an unhandled Python traceback and exit code 1 instead of the documented single-JSON-line + exit-4 contract the whole script exists to guarantee.

**Suggested fix:**
Wrap the two emit({...}) calls in the KeyboardInterrupt and Exception handlers with the same try/except OSError: return 4 pattern used for the terminal payload emit, so a broken pipe during an error report degrades identically to the happy path instead of crashing with an undocumented exit code.

scripts/await_workflow.py
```diff
@@ -0,0 +1,913 @@
...
+        payload, code = await_terminal(args, environ)
+    except KeyboardInterrupt:
+        # Even an interrupt gets a line: the caller branches on stdout, and a
+        # silent exit is the one shape it cannot interpret.
+        emit({
+            "await": "error", "gap": "workflow-timeout",
+            "message": "interrupted", "target": args.target,
+            "attempt": args.attempt, "max_attempts": args.max_attempts,
+        })
+        return 4
```
leehopper: **💡 [LOW] KeyboardInterrupt handling in main() has no test**

main() has a dedicated except KeyboardInterrupt branch (distinct from the generic except Exception branch below it) that emits a specific {await:error, message:interrupted, ...} marker and returns exit 4, but no test in tests/test_await_workflow.py exercises it — the existing generic-failure test only raises RuntimeError, which is caught by the separate except Exception clause since KeyboardInterrupt does not inherit from Exception in Python 3.

**Suggested fix:**
Add a test that patches scripts.await_workflow.await_terminal with side_effect=KeyboardInterrupt() and asserts main() returns 4, prints exactly one JSON line, and that the marker's message field equals 'interrupted' (distinct from the generic exception's dynamic type:message format).

scripts/await_workflow.py
```diff
@@ -0,0 +1,913 @@
...
+    except Exception as exc:  # noqa: BLE001 - the never-print-nothing contract
+        emit({
+            "await": "error", "gap": "workflow-timeout",
+            "message": "%s: %s" % (type(exc).__name__, exc),
+            "target": args.target,
+            "attempt": args.attempt, "max_attempts": args.max_attempts,
+        })
+        return 4
```
leehopper: **💡 [LOW] Duplicate error-marker construction in main()'s KeyboardInterrupt and Exception handlers**

The KeyboardInterrupt and bare Exception handlers in main() each build an identical five-key error marker dict and call emit(), differing only in the "message" value. Before: two ~7-line emit({...}) calls repeating "await"/"gap"/"target"/"attempt"/"max_attempts". After: a small helper such as _emit_wait_error(args, message) called once per except clause with just the message argument, e.g. except KeyboardInterrupt: return _report_wait_error(args, "interrupted"). This removes the duplication so the shared field set can't drift between the two branches while a future edit updates only one.

**Suggested fix:**
Extract a helper, e.g. def _wait_error_payload(args, message): return {"await": "error", "gap": "workflow-timeout", "message": message, "target": args.target, "attempt": args.attempt, "max_attempts": args.max_attempts}, and call emit(_wait_error_payload(args, "interrupted")) / emit(_wait_error_payload(args, "%s: %s" % (type(exc).__name__, exc))) from the two except blocks, both still followed by return 4.

scripts/await_workflow.py
```diff
@@ -0,0 +1,913 @@
...
+    outcome that leaves the caller unable to tell a failure from a dead process.
+    """
+    try:
+        line = json.dumps(payload, separators=(",", ":"))
+    except (TypeError, ValueError) as exc:
+        line = json.dumps({
+            "await": "error",
+            "gap": "workflow-timeout",
+            "message": "result would not serialize: %s" % exc,
+        }, separators=(",", ":"))
```
leehopper, line 680: **💡 [LOW] emit()'s non-serializable-payload fallback branch is untested**

emit() documents and implements a fallback: if json.dumps(payload) raises TypeError or ValueError, it substitutes a hand-built minimal error line instead of letting the exception propagate, preserving the never-print-nothing contract. No test in tests/test_await_workflow.py passes emit() (or main(), indirectly) a payload that fails to serialize, so this branch has zero coverage.

**Suggested fix:**
Add a unit test calling emit() directly (or via a monkeypatched await_terminal returning a payload containing a non-JSON-serializable object, e.g. a set() or a custom class instance) and assert stdout is exactly one line, that line is valid JSON, and it has await:'error' with a message mentioning the original TypeError/ValueError text.

scripts/await_workflow.py
```diff
@@ -0,0 +1,913 @@
...
+#: Fields that only the pipeline's own compact return carries. A bare `ok` is NOT
+#: enough to call an object terminal: the assemble receipt
+#: ({ok, planVersion, planChecksum, verified, written, errors}) and other agent
+#: receipts also carry one, and mistaking a receipt for the return would send
+#: Phase 8 off with the wrong object — a far worse outcome than waiting longer.
+#: Both observed return shapes are covered: the success shape
+#: {ok, phaseReached, stats, artifactPaths, resolvedPolicy, checkpoints, gaps} and
+#: the early-failure shape {ok, phaseReached, failingPhase, error, stats,
+#: artifactPaths, gaps}, which omits resolvedPolicy and checkpoints entirely.
+#: `error` is deliberately not in this list — it is far too generic to corroborate.
```
leehopper, line 147: **💡 [LOW] COMPACT_RETURN_KEYS comment mischaracterizes the pipeline's failure-return shapes**

The comment claims the codebase has exactly two observed compact-return shapes — a success shape and 'the early-failure shape {ok, phaseReached, failingPhase, error, stats, artifactPaths, gaps}, which omits resolvedPolicy and checkpoints entirely'. That description matches only the args-rejection envelope (makeArgsRejectEnvelope in workflows/src/args.js). The mid-run catch-all failure in workflows/src/stages.js's runPipeline (the catch block around line 2653) returns {ok:false, error, phaseReached, failingPhase, artifactPaths, stats, checkpoints, gaps} — it DOES carry checkpoints (via buildResumeCheckpoints) and only omits resolvedPolicy, contradicting the comment's blanket claim that failures omit checkpoints entirely. Detection itself still works (any COMPACT_RETURN_KEYS member corroborates), so this is a documentation-accuracy gap rather than a functional bug.

**Suggested fix:**
Rephrase the comment to note there are (at least) two distinct failure shapes — the early args-rejection envelope (no checkpoints, no resolvedPolicy) and the mid-run catch failure (has checkpoints, no resolvedPolicy) — or generalize the claim to avoid asserting an exhaustive, field-exact enumeration.

skills/code-gauntlet/references/phase3-dispatch.md
```diff
@@ -22,14 +22,26 @@ Workflow(
...
 
-Headless `-p` child sessions are the highest-risk case (they background aggressively) — they must poll, never assume completion. A missing/empty compact return is a failure to surface, not an empty-but-successful review.
+1. **Never end the turn and never begin Phase 8 without a terminal workflow result.**
+2. **If the tool result carries the compact return inline** (no Task ID) → carry it into Phase 8.
+3. **Otherwise** → run the awaiter with the Task ID, in-turn, under an explicit Bash `timeout: 600000` (the tool's 120s default would kill it mid-wait):
```
leehopper: **💡 [LOW] Wait Protocol section asserts Workflow 'always' backgrounds, then immediately documents handling for the case where it doesn't**

Line 25 (new in this diff) states unconditionally that '`Workflow` always runs the pipeline as a background task' and that what comes back is 'not the compact return' — implying the inline-return case cannot occur. Five lines later, step 2 (also new in this diff) instructs the reader to check 'If the tool result carries the compact return inline (no Task ID) → carry it into Phase 8', treating that same case as a live branch to handle. The two statements in the same section make incompatible claims about whether an inline compact return can ever be observed, which is confusing for whoever (model or human) is following this MANDATORY protocol.

**Suggested fix:**
Soften line 25's 'always' claim (e.g. 'Workflow typically runs the pipeline as a background task') to match the fallback branch retained in step 2, or drop step 2 if inline resolution is truly impossible and update await_workflow.py's bare-return handling (terminal_from case 3) accordingly.

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: e7debc48fe7ebfd0f04276ad80be1f65df7d30d7

scripts/await_workflow.py
```diff
@@ -0,0 +1,913 @@
...
+            # alone would count it as present and fire the fallback on a run that
+            # persisted nothing.
+            fresh = (os.path.isfile(path)
+                     and stat.st_size > 0
+                     and stat.st_mtime >= since_epoch)
+        except OSError:
+            fresh = False
+        (state["present"] if fresh else state["missing"]).append(
+            os.path.basename(path)
+        )
```
leehopper: **💡 [LOW] Conditional-object-then-append idiom in artifacts_state is less immediately readable than an if/else**

artifacts_state() selects between state["present"] and state["missing"] with a parenthesized ternary expression immediately followed by .append(), which requires the reader to mentally evaluate the ternary before recognizing which list is being mutated. Before: (state["present"] if fresh else state["missing"]).append(os.path.basename(path)). After: an explicit if/else block appending to the named list in each branch makes the mutated target obvious without evaluating the expression first.

**Suggested fix:**
Replace with: name = os.path.basename(path); if fresh: state["present"].append(name); else: state["missing"].append(name) — or equivalently target = state["present"] if fresh else state["missing"]; target.append(os.path.basename(path)), separating the selection from the mutation.

PR #79, opened 2026-07-29
feat(context): resolve @import pointers so project rules actually reach agents (#49)

### LINKED ISSUE #49: Decide whether AGENTS.md (and QODO.md) content should reach the review agents

leehopper:
Resolves the owner decision issue #49 asked for, and lands it.

## The decision, and why it went the other way from the earlier recommendation

A 2026-07-27 sequencing pass recommended **"#49 — decline, do not defer"**, on three premises: AGENTS.md is 80 lines at repo root; *"CLAUDE.md already reaches the agents through the shared context"*; and adopting it grows the context file *"at the exact moment #48 has proven that context size is a live silent-failure mode"* — closing with **"Decline now, revisit after #48 lands and read-completeness is provable."**

That was correct when written. Every premise has since flipped:

- **#48 landed** (`5106d06`). Read-completeness is provable, and the revisit trigger the recommendation itself set has fired. Measured: a typical AGENTS.md leaves `contextReadPlan` at **4 chunks → 4** — no extra `Read` call for any agent.
- **"CLAUDE.md already reaches the agents" is false** for the layout most repos now use (below).
- **"80 lines at repo root"** describes *this* repo's AGENTS.md, which is pure Cursor-Cloud/CI content. It is an unrepresentative sample — this repo is one of the few that still keeps its real rules in CLAUDE.md.

## The defect

Phase 2 read project rules with the `Read` tool. `Read` does **not** expand Claude Code's `@path` import directive (verified empirically). Anthropic's current docs say:

> "Claude Code reads `CLAUDE.md`, not `AGENTS.md`. If your repository already uses `AGENTS.md`…, create a `CLAUDE.md` that **imports** it"

So the pipeline was blind precisely on repos that followed Anthropic's own documented guidance. Measured at HEAD across the five benchmark mirrors:

| repo | root `CLAUDE.md` | pipeline saw | now |
|---|---|---:|---:|
| sentry | `@AGENTS.md\n` (identical blob to grafana) | **11 B** | 8,517 B |
| grafana | `@AGENTS.md\n` | **11 B** | 8,589 B |
| discourse | `See @AI-AGENTS.md for all instructions.\n` | **40 B** | 7,570 B |
| cal.com | symlink → AGENTS.md | 9,225 B | 9,239 B (no regression) |
| keycloak | absent | 0 | 0 + disclosed gap |

For three of five real repos, the entire "project rules" section handed to all nine context-consuming agents was a pointer string — silently. Grafana's unread AGENTS.md contains rules that map straight onto this tool's own dimensions (*"Never include links … in code comments"* → comment_accuracy; *"Write tests for new functionality"* → test_coverage).

**A filename allowlist is not the fix.** discourse points at `AI-AGENTS.md`, which no `{CLAUDE,AGENTS,QODO}.md` list contains. The pointer is the mechanism, so the pointer is what gets resolved.

## What ships

`scripts/collect_project_rules.py` — stdlib-only, language-agnostic. Discovers `CLAUDE.md`/`AGENTS.md`/`QODO.md` at the root and in directories with changed files, follows `@path` imports matching the real contract (inline mid-sentence, skipped inside code spans/fences, resolved against the *containing file's* directory, 4-hop cap), writes the block to `--out`, and prints exactly one line of provenance JSON.

This is a **security boundary**, not plumbing: its output reaches nine agent prompts and potentially a posted PR comment, which `SECURITY.md` puts in scope. Every resolved path is `realpath`-confined with a separator-aware check, must be `.md`, and every bound comes off `os.stat` **before** any `open`. cal.com proves a symlinked CLAUDE.md is a real shape, so symlinks are resolved-and-confined rather than banned.

Also: the ad hoc AGENTS.md/QODO.md self-read in `agents/conventions-and-intent.md` is **retired** — it was a third read path with no cap, no import resolution, no truncation disclosure and no test (the #48 anti-pattern, one file at a time).

**QODO.md is included** (requirement 3). It is maintainer-authored repo metadata written *for a review tool*; `.pr_agent.toml` is Qodo's behavior config. `PROJECT_RULE_FILENAMES` is the one place a name is added.

## Requirement 2, satisfied structurally

The read and the carry-forward are now the *same program*, so "a read whose output reaches no assembly step" is not expressible. Two supporting rules in `phase2-triage.md`:

- The block is folded into the **same `content` string** measured for `contextLines`/`contextChars`. Appending after the count would silently reopen #48.
- The read is deliberately **unguarded** — no try/except, no `exists()`, no fallback. The script always writes `--out` (empty when a repo genuinely has none), so a *missing* file can only mean the step never ran, and failing loudly is the guard.

## Verification — deterministic suites, deliberately not a smoke

**970 Python (+36) / 539 node / lint clean.**

A smoke run **cannot observe this change**: at all 50 pinned golden SHAs, none of those repos has an AGENTS.md, so it would pass vacuously. Coverage is fixtures plus the real mirror layouts above. Security tests use canary strings for traversal, absolute (the silent `os.path.join` base-discard), home-relative, `repo-evil` sibling-prefix, escaping symlink, in-repo `.env`, a `.md`-named symlink aimed at a secret, and proof that `open` is never called on an over-cap file.

**Known limits, stated plainly:** no test can force a model-executed Phase 2 to invoke the script — the loud failure is the guard, and a doc-grep test would be the phrase-count guard CLAUDE.md forbids. Past benchmark numbers are **not** contaminated (the pinned SHAs predate these repos' AGENTS.md adoption), but this does change agent input, so recall/noise effects are unmeasured by construction.

Closes #49

https://claude.ai/code/session_01RGSPj6a4H9yk1KduSnPfkp

---

> [!NOTE]
> **Medium Risk**
> New security boundary reads attacker-controlled repo files into agent/PR context; mitigated by confinement, caps, and tests, but Phase 2 still depends on the orchestrator actually invoking the script (unguarded `open` only fails if skipped).
> 
> **Overview**
> Fixes **silent empty project rules** when a repo’s root `CLAUDE.md` is only a Claude Code `@path` import (e.g. `@AGENTS.md`, `@AI-AGENTS.md`) — `Read` never expands those pointers, so agents often saw an 11–40 byte stub instead of real conventions.
> 
> Adds **`scripts/collect_project_rules.py`**: discovers `CLAUDE.md` / `AGENTS.md` / `QODO.md` (root + dirs tied to changed files), follows `@` imports with repo **`realpath` confinement**, `.md`-only targets, depth/size/file-count caps, and **`stat` before `open`**. Writes assembled markdown to `--out` (always, including empty) and prints a one-line JSON receipt with **`gaps[]`** for skips/refusals.
> 
> **Phase 2 / SKILL** now runs the script, emits `project-rules-*.md`, and folds it into the shared context via **unguarded `open().read()`** inside the same `python3 -c` that builds **`content`** — so **`contextLines` / `contextChars`** include the rules block (avoids reopening #48). Triage can surface receipt **`gaps`**.
> 
> **Agents** (`bug-detector`, `code-simplifier`, `conventions-and-intent`) stop **`Read`ing** convention files themselves; they use the pre-loaded project-rules section. **`REVIEW.md`** stays on its separate structured path (not collected by the script).
> 
> **Tests**: new `tests/test_collect_project_rules.py`; **`CLAUDE.md`** byte budget bumped for the new script bullet.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 550a9dab2007939a723dd5889b2f6ab0682c5bf9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/collect_project_rules.py
```diff
@@ -0,0 +1,514 @@
...
+    except Exception as exc:  # noqa: BLE001 — a receipt on every path
+        sys.stderr.write("collect_project_rules: %s\n" % exc)
+        try:
+            write_text_atomic(args.out, render(collector.sources) if collector else "")
+        except Exception:  # noqa: BLE001
+            pass
+        _emit({"ok": False,
+               "sources": [], "skipped": collector.skipped if collector else [],
+               "total_bytes": 0, "truncated": False, "out": args.out,
+               "gaps": ["project_rules_failed: %s" % exc]})
```
leehopper: **🟡 [MEDIUM] Receipt dict shape is independently reconstructed in four places**

The 7-key receipt shape is hand-built as a separate dict literal at four call sites instead of one builder function.

agents/conventions-and-intent.md
```diff
@@ -14,11 +14,7 @@ You are a conventions, intent, and documentation accuracy reviewer. Your job is
...
-   - `CLAUDE.md` (root and any subdirectory-level files) — primary project conventions
-   - `REVIEW.md` — custom review rules and checklists specific to this project's code review process
-   - `AGENTS.md` — agent-specific instructions that may contain code quality rules
-   - `QODO.md` — additional review configuration and rules
+1. **Project rules arrive pre-loaded — do not `Read` them yourself.** `CLAUDE.md`, `REVIEW.md`, `AGENTS.md`, and `QODO.md` rules are already assembled into the shared context file you read at startup: Phase 2 gathers `CLAUDE.md`/`REVIEW.md` directly and resolves `AGENTS.md`/`QODO.md` — including any `@path` import a repo's CLAUDE.md points at (e.g. `@AGENTS.md`) — through `scripts/collect_project_rules.py` before concatenating the result into that file. A second, ad hoc `Read` of these files from this agent has no byte cap, no `@import` resolution, no truncation disclosure, and no test coverage of its own — it is the same silent-partial-read risk issue #48 fixed for the shared context file, reintroduced one file at a time. Rely on the shared context's project-rules section instead.
```
leehopper, line 17: **💡 [LOW] Policy fix for direct CLAUDE.md reads applied to one agent but not propagated to code-simplifier.md or bug-detector.md**

conventions-and-intent.md now tells its agent not to Read CLAUDE.md/REVIEW.md/AGENTS.md/QODO.md directly, but code-simplifier.md and bug-detector.md still instruct direct Read of CLAUDE.md, missing the same fix.

scripts/collect_project_rules.py
```diff
@@ -0,0 +1,514 @@
...
+        if match:
+            marker = match.group(1)[0]
+            if fence is None:
+                fence = marker
+            elif fence == marker:
+                fence = None
+            out.append("")
+            continue
+        out.append("" if fence is not None else _INLINE_CODE_RE.sub("", line))
+    return "\n".join(out)
```
leehopper, line 159: **💡 [LOW] _strip_code closes a fence on any same-type marker regardless of length, mis-parsing nested fences of differing backtick/tilde counts**

_strip_code tracks only the fence character type via marker[0], not the actual run length, so a nested fence of differing length incorrectly closes the outer fence early.

leehopper (commented):
## Code Gauntlet Review

Reviewed 1,069 changed lines across 8 files implementing @import pointer resolution for project rules (issue #49).

**8 high-confidence findings** — all survived validation, filtering, and blind challenge verification. 5 main findings (1 medium, 4 low), 3 improvement suggestions.

See inline comments below, or the full report for details.

skills/code-gauntlet/SKILL.md
```diff
@@ -271,7 +271,11 @@ The assembled `reviewConfig` is exactly the `parseReviewMd` output shape — **`
...
+Write the shared context to `{output_dir}/code-gauntlet-context-{head_sha_short}.md` using `python3 -c "import json; ..."`. Contents, concatenated in this order into one `content` string: (1) CLAUDE.md/REVIEW.md project rules (2d steps 1-2, by value); (2) AGENTS.md/QODO.md project rules, resolved by `scripts/collect_project_rules.py` (2d step 3) and folded in via `open(path).read()` on its `--out` file inside this SAME `python3 -c` — never retyped by the model (CLAUDE.md's "Artifact persistence" section records the artifact-writer's transcription of a multi-KB payload diverging from its input on 3 of 3 measured runs; hand-copying this block risks the identical failure one stage earlier); (3) risk classification (2e) and AI-generated-code status (2k); (4) the full diff inside `<untrusted-code-content>` tags. The workflow's discovery, validate, and summarize agents Read this file at `{output_dir}/code-gauntlet-context-{head_sha_short}.md` — the workflow threads exactly this path to them, so the filename must match. (The change **summary** is no longer written here — the workflow's Summarize stage produces it internally.)
```
leehopper: **💡 [LOW] Context piece (2) is documented as 'AGENTS.md/QODO.md' but the script it invokes also re-collects CLAUDE.md, duplicating piece (1)**

SKILL.md labels piece (2) as only AGENTS.md/QODO.md, but PROJECT_RULE_FILENAMES also includes CLAUDE.md, so CLAUDE.md content is folded in twice for repos with a real CLAUDE.md.

CLAUDE.md
```diff
@@ -5,6 +5,7 @@
 - **stdlib-only Python.** No pip dependencies. All scripts must use only the Python standard library.
 - **Language-agnostic.** Scripts must not assume any particular programming language in the reviewed codebase. No `--include=*.py` or similar language filters — use `--exclude-dir` for non-source directories instead.
 - **Repo root for searches.** `verify_findings.py` resolves the repo root at startup via `git rev-parse --show-toplevel`. Symbol searches use `git grep -l` with `cwd=REPO_ROOT` and a 3-second per-symbol timeout.
+- **`scripts/collect_project_rules.py` resolves `@path` import pointers, not just filenames (issue #49).** `Read` does not expand Claude Code's `@import` directive, and real repos increasingly ship a CLAUDE.md that is nothing but one (`@AGENTS.md`) — a fixed filename allowlist still misses an arbitrary import target (e.g. `@AI-AGENTS.md`). The script confines every resolved path inside the repo via `realpath`, requires `.md`, and bounds depth/size via `os.stat` before any `open`.
```
leehopper: **💡 [LOW] New CLAUDE.md bullet misattributes the import-depth bound to os.stat**

The bullet claims depth/size are both bounded via os.stat before any open, but the depth cap is a plain integer comparison with no stat call.

scripts/collect_project_rules.py
```diff
@@ -0,0 +1,514 @@
...
+        return []
+    out = []
+    for item in data:
+        if isinstance(item, str):
+            out.append(item)
+        elif isinstance(item, dict):
+            value = item.get("path") or item.get("file")
+            if isinstance(value, str):
+                out.append(value)
+    return out
```
leehopper, line 369: **💡 [LOW] `_load_changed_files`'s dict-shaped entries and malformed-input handling are untested**

Dict-shaped changed-file entries (path/file keys) and malformed/missing changed-files input are never exercised by tests.

scripts/collect_project_rules.py
```diff
@@ -0,0 +1,514 @@
...
+        try:
+            st = os.stat(real)
+        except OSError:
+            return None, "missing"
+        if not stat.S_ISREG(st.st_mode):
+            return None, "not_regular"
+        if st.st_size > self.max_file_bytes:
+            return None, "too_large"
+        if self.total_bytes + st.st_size > self.max_total_bytes:
+            self.truncated = True
```
leehopper, line 287: **💡 [LOW] `not_regular` skip reason has zero test coverage**

stat.S_ISREG check in _Collector._read is documented but no test creates a non-regular file to verify the branch fires and is disclosed correctly.

scripts/collect_project_rules.py
```diff
@@ -0,0 +1,514 @@
...
+        Order matters. Absolute and home-relative tokens are rejected *before*
+        any join: ``os.path.join(base, "/etc/passwd")`` silently discards the
+        base and returns ``/etc/passwd``, a failure with no exception to catch.
+        Native Claude Code asks a human to approve an import that resolves
+        outside the working directory; this script cannot prompt, so it refuses.
+        """
+        if raw.startswith("~") or raw.startswith("\\") or "\\" in raw:
+            return None, "absolute_path"
+        if os.path.isabs(raw) or _WINDOWS_DRIVE_RE.match(raw):
+            return None, "absolute_path"
```
leehopper: **💡 [LOW] Two separate absolute-path guards both return the same (None, "absolute_path") result**

Two consecutive if-statements each return the identical tuple; could be merged into one guard clause.

Issue #22, opened 2026-07-23
Consensus consolidation: re-land deterministic agent identity without cross-agent dedup recall collapse


leehopper:
## Problem / Goal

Cross-agent proximity dedup treats convergence as redundancy: `dedupCrossAgent` fires when 2+ distinct agents land within a 5-line bucket and drops all but one winner. But multiple agents converging on the same location is the normal signature of a *real* bug, so eliminations land disproportionately on goldens. This stayed latent until V3.1 item 4 made the merge-injected `agent` identity survive the verify echo deterministically — correct identity woke the dedup up everywhere, and same-6 recall collapsed 20/30 → 13/30. Item 4 was surgically reverted to recover.

Goal: redesign cross-agent consolidation so convergence **boosts, never drops**, then re-land the deterministic agent-identity echo on top of it. Consolidation must also govern the posted PR-comment payload, not only report-render (PR-310 posted near-duplicate inline comments from two dimensions).

## Requirements

1. Deterministic merge-injected `agent` identity survives the verify echo end-to-end; identity is a prerequisite for consolidation, not a casualty of it.
2. Cross-agent convergence on one location must corroborate the finding — no drop-one-winner semantics anywhere in the pipeline.
3. Consolidation applies to the posted PR-comment payload (eliminating the PR-310 L11 near-duplicate inline-comment class), not only at report-render.
4. Parity contract holds: the Python filter twin and golden parity fixtures move together with the JS change (CLAUDE.md JS/Python parity rules; bundle-freshness stays green).
5. Rider (non-urgent, same files in scope): normalize the shell-era `\'`/`\\'` escapes — invalid JSON string escapes — in the security-reviewer, code-simplifier, and type-design-analyzer example blocks to valid JSON.
6. Deterministic preference: consolidation semantics live in workflow-runtime code/scripts, not prompt tuning.

## Evidence

```
V3.1 mini-subset A, first leg custom-20260723-070640-c1dd46f (item 4 echo declaration active):
  dedup eliminations 7 -> 33 vs Gate-1 on the same 6 PRs
  same-6 recall 20/30 -> 13/30; challenge held at historical 0-1/PR

Gate-1 = V3 gate baseline run subset-20260721-015119-639e4bc (15-PR gate run,
  recall 0.6949 / noise 0.1801; the 6 pre-registered mini-subset PRs are a subset
  of it, on which it scored 20/30; record also in issue #17's closeout comment).

[...]

Issue #69, opened 2026-07-28
One stray '}' from the artifact-writer costs the whole verify stage — 23 findings degraded to origin=unknown


leehopper:
## Problem / Goal

On smoke run `smoke-20260728-144630-a162ecd` (2026-07-28), **2 of 3 PRs lost verification entirely** — 23 findings degraded to `origin=unknown`, surfaced-classification skipped — because the artifact-writer appended **one stray `}`** after an otherwise complete, valid slice-input document.

```
verify: UNVERIFIED — slice 0: status=failed (ERROR: Invalid JSON in findings file:
Extra data: line 1 column 20523 (char 20522)); all 11 finding(s) marked origin=unknown
```

The file is 20,524 chars. The complete JSON value ends at 20,522. The trailing bytes are exactly `}\n`:

```
...,"cross_file_refs":[],"agent":"code-simplifier"}],"base_branch":"main"}  <- valid JSON ends here
}                                                                          <- writer added this
```

Identical corruption, identical shape, on both failing PRs (grafana#80329 at col 20522, discourse-graphite#6 at col 20200).

**Every finding is recoverable.** `json.JSONDecoder().raw_decode()` parses the leading complete value and reports where it ended:

```
grafana-grafana-80329       strict=FAIL  raw_decode=OK  findings=11  trailing='}\n'
discourse-graphite-6        strict=FAIL  raw_decode=OK  findings=12  trailing='}\n'
   recovered findings all have id+description: True
```

So `verify_findings.py` discards 23 fully-intact findings' verification over one character a sampled agent appended.

## Why this is the right thing to fix in code

The writer is a sampled agent, not a function — CLAUDE.md already says so ("The by-value writer is not trustworthy — that is what the proof is for", measured 3/3 transcription divergence on the 2026-07-27 smoke). The pipeline's answer everywhere else is a deterministic proof or a deterministic recovery. Here it is neither: a strict `json.loads` turns a one-character model artifact into a total loss of the verify stage.

This is not "tell the writer to be careful". It is a parser that can safely recover a well-defined corruption class and currently does not.

## Requirements

1. `verify_findings.py` parses its input with `raw_decode` rather than strict `json.loads`: take the leading complete JSON value, and **accept** a trailing remainder that is only whitespace and/or unbalanced closing punctuation (`}`, `]`).
[...]

PR #83, opened 2026-07-29
fix(verify): echo a per-id delta with a content proof, not the findings (#25 PR2)

### LINKED ISSUE #25: Harden the verify executor/writer boundary: per-id delta echo, slice-input content proof, per-slice degradation, and a deterministic report degradation banner

leehopper:
Closes #25's requirements **1** (delta echo), **2** (trustSlice adapts), and **10** (parity/tests). PR1 (#71) landed requirement 3; requirements 4–6 (PR3) and 7–9 (PR4) stay open.

## The problem

The verify boundary pushed every finding through a sampled agent and back. The executor ran `verify_findings.py`, read its `--output` file, and re-typed the whole `result.verified`/`result.eliminated` arrays through StructuredOutput — and whatever came back **became** the pipeline's findings for every later stage. The workflow already held those findings by value, so the round trip transported data it already had, and paid for it three ways:

- a **1,990-byte `VERIFY_SCHEMA`** (the file's largest) dispatched per slice, which had to union every dimension's extras purely to survive transcription;
- **22.95s of one profiled executor's 31.9s generation**, spent strictly between its `Read` of the slice output and its `StructuredOutput` call;
- a trust boundary that could only check **shape**. `trustSlice` bound nonce, sha, `n_in` and a count sum, never content. Live: a script that wrote 10 verified / 0 eliminated to disk echoed as 7/3 under a valid receipt. And because `verifySliceSize` is a constant, an envelope carrying a **sibling slice's** findings satisfied every guard — the real slice's findings were **dropped outright**, not degraded, while the run reported `verified: true`.

## The change

The executor now answers with the one thing only it can know — what the script decided:

```
receipt: { sha, n_in, nonce, deltas_checksum }
result:  { deltas: [ {id, verified, origin?, severity?, confidence?, elimination_reason?}, ... ] }
```

`joinVerifyDeltas` rebuilds each slice's verified findings by walking the **dispatched** slice and applying its delta. Findings never cross the boundary, so no echo can substitute or delete one — the worst an untrusted answer achieves is its own slice's honest degrade. The unqualified never-drop assertions #71 had to remove from the docstring, CLAUDE.md and `validation-pipeline.md` come back.

### The PoC on the issue thread, before and after

```
$ node poc_verify_content_swap.mjs        # verbatim from the issue

on main                          on this branch
---------------------------      ------------------------------------------
ids out: F2,F3,F2,F3             ids out: F0,F1,F2,F3
F0 present: false                F0 present: true
verified: true                   verified: false
                                 gaps: 2 x "verify: UNVERIFIED — ..."
```

### trustSlice's guards, in increasing strength

1. **Delta-id coverage** replaces the count guard. The deltas must name exactly the dispatched ids — no missing, no duplicate, no stranger. A count is trivially satisfied by any same-length sibling; an id set is not.
2. **Shape and stamp.** `verified` must be boolean; a `verified:false` delta must carry the `elimination_reason` stamp `run_verification` always writes, and a `verified:true` delta must not. Kept even though (3) subsumes it — a precise reason in the gap beats a checksum mismatch when a human reads the run.
3. **Content proof.** `verify_findings.py` checksums its own deltas (`fnv1a32(js_stringify_pretty(deltas))`); the workflow recomputes over a canonical rebuild in dispatch order. This closes coherent drift — one flipped `origin`, one shifted `confidence`, one plausible elimination — that satisfies (1) and (2). An executor that *transcribes* cannot recompute it. Imported from `assemble_artifacts.py` rather than copied, so `tests/test_assemble_artifacts.py` guards both boundaries.

The canonical form is rebuilt from the dispatched id order, so a **reordered** echo is tolerated while a **changed value** is not.

### Requirement 1's agent-withholding constraint is now explicit

`agent` used to be withheld because the echo schema didn't declare it — which worked only stochastically (2 of 6 PRs measured surviving). Joining onto findings held by value would have made survival deterministic, which is the measured dedup recall-collapse mechanism (mini-subset A: eliminations 7→33, same-6 recall 20/30→13/30). `joinVerifyDeltas` deletes it; a test fails if a joined finding is ever filter-visibly tagged. It re-lands with #22.

### Notes

- `result.deltas` is the **first** key of the script's `result`: the executor's `Read` is length-capped with no truncation notice, so what it echoes must be a prefix of the document. The full `verified`/`eliminated`/`batches`/`stats` arrays stay on disk unchanged for bench/v2; the legacy positional CLI is untouched.
- `_delta_confidence` canonicalises confidence to an integer in **one** runtime, so the two never have to agree on float spelling (the precondition `assemble_artifacts.py` already enforces). The JS rejects a non-integer.
- A slice whose own findings have no usable id set is never dispatched — the join is keyed by id, and spending an executor on an unusable answer is worse than degrading now. Degrade-and-disclose, not reject.
- **Side effect worth noting for PR3:** delta-id coverage also catches a *corrupted slice input* (a writer that put slice 1's content in slice 0's file now fails coverage instead of silently verifying the wrong set). PR3's proof-before-spend is still worth having — it is cheaper and pre-dispatch — but that threat is no longer silent.

## Verification

- `python -m pytest tests/ -q` — 994 passed, 308 subtests
- `node --test workflows/test/*.test.js` — 568 passed
- Bundle freshness (`tests/test_bundle_fresh.py`) green; bundle rebuilt from source.
- **Dual-runtime golden fixtures** (`tests/fixtures/parity/verify_deltas/`, 7 cases) pin requirement 1's equivalence claim itself: Python's `build_deltas` produces the golden, JS's `joinVerifyDeltas` must reconstruct from it exactly what the script left on the finding, and both runtimes must compute the same content proof (including an astral-plane case for the UTF-16 code-unit definition).
- `workflows/test/stages_verify_delta.test.js` — 22 new guard tests, including the issue's published PoC.
- The functional smoke this issue specifies is running separately; results will be posted here.

## Doc surface

`agents/executor.md` (the return contract), `references/validation-pipeline.md` (envelope, trust rule, the now-closed shape-vs-content gap), `references/phase2-triage.md`, `references/phase3-dispatch.md`, and a new CLAUDE.md "Verify boundary" section carrying only the four rules that span two files. The CLAUDE.md byte ceiling is re-pinned in the same PR with its justification, per `tests/test_claude_md_budget.py`'s own discipline — the first draft of that section was +4,292 bytes and was cut ~60% to exactly the cross-file rules.

https://claude.ai/code/session_016cn2LvUDJ4pQyrLFqJextp

---

> [!NOTE]
> **Medium Risk**
> This changes the core verify trust boundary and finding enrichment path; mistakes could mis-classify or degrade slices, though per-slice degradation and extensive parity tests limit blast radius.
> 
> **Overview**
> **Verify no longer re-types full finding arrays through the executor.** After `verify_findings.py` runs, the agent returns only `status`, receipt fields (`sha`, `n_in`, `nonce`, **`deltas_checksum`**), and **`result.deltas`** — compact per-id decisions (`verified`, `origin`, `severity`, `confidence`, `elimination_reason`). Full `verified`/`eliminated` output stays on disk for bench/v2; the script puts **`deltas` first** in `result` so a length-capped Read still captures what the executor must echo.
> 
> **The workflow rebuilds trusted findings locally** via **`joinVerifyDeltas`**: it walks the slice it already dispatched, applies each delta, pins numeric fields (including half-up rounding aligned with Python), and **strips `agent`** so dedup behavior stays as before. **`VERIFY_SCHEMA`** shrinks to flat delta items (no per-dimension finding union). **`trustSlice`** now requires exact delta-id coverage, shape/stamp checks, and **`fnv1a32`** over canonical deltas (shared with `assemble_artifacts.py`). Slices without usable ids skip dispatch and degrade with a gap.
> 
> **Python side:** `build_deltas` / `deltas_checksum`, receipt envelope changes, **`InputError`** so receipt mode writes honest failure JSON instead of silent exit, and input-boundary numeric normalization. **Docs/tests:** executor contract, validation pipeline, CLAUDE.md verify-boundary rules, parity fixtures under `verify_deltas/`, and expanded unit/subprocess tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a553754e3b26bee2c66e7da5191e2307a4e49353. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:
## Functional smoke: `smoke-20260729-193917-f08d4f6` (3 PRs, commit `f08d4f6`)

**All three PRs terminal `ok`, `unknown_origin=0` across the run, 31 comments delivered, plugin identity validated on 4/4 records.** The verify boundary was **trusted on every slice of every PR**.

```
status: ok=3
Check: ok=False pr_dirs=3 comments=31 findings_files=3
       workflow_records=4 script_paths=4 unknown_origin=0
  FAIL: pr-grafana-grafana-80329: writer degrade signal in workflows/wf_a62f7348-a3f.json
```

The single FAIL reads a **superseded** record — grafana's first attempt, killed by a sustained API 529, which `--retry-failed` then ran clean. `check.py`'s `_iter_workflow_records` globs every `wf_*.json` in the PR dir with no notion of which attempt is current, so the failed attempt's signal is immortal. Filed as **#85**. Verified per record with the checker's own `_DEGRADE_RE`:

| record | gaps | degrade hits |
|---|---|---|
| `wf_744f098e-8c6` (retry, current) | 1 — a *disclosure* gap, no UNVERIFIED token | **0** |
| `wf_a62f7348-a3f` (superseded, 529-killed) | 11 | 1 — `writeArtifacts: writer returned null` |

### What the run proves about this change

**Three live captures, on real executor output rather than fixtures.**

**1. The content proof matched live.** The first `phase4-output` file to survive a run in this whole effort:

```
receipt: {"sha":"04cfa3bf","n_in":12,"nonce":"57b012bea3a1206b.0.r1",
          "deltas_checksum":"fnv1a32:0x9d703cd5"}
result key order: ['deltas','verified','eliminated','batches','stats']

script checksum    : fnv1a32:0x9d703cd5
workflow recomputed: fnv1a32:0x9d703cd5   -> MATCH
joined: 12 of 12 dispatched | agent stripped: true
```

**2. The prefix property holds with enormous margin.** Those deltas occupy the first **230 of 43,704 chars**. The executor's length-capped `Read` reaches everything it must echo in the first 0.5% of the document — which is the entire reason `deltas` is the first key of `result`.

**3. The `.r1` nonce is doing its job, on the exact failure this issue was filed over.** That receipt came from the *retry*: attempt 1 came back with `receipt nonce mismatch (got missing, expected 57b012bea3a1206b.0)` — the same dropped-nonce echo that cost `smoke-20260727-205454-f99d948` all 16 findings of discourse-graphite#6. Here it produced a disclosure gap and a trusted second dispatch. No finding lost its classification.

And **discourse-graphite#6 itself** — the PR quoted in #25's evidence block — came back `verified: true` with zero verify gaps.

### The aborted first run earned its cost

`smoke-20260729-191253-8ae2ee3` was killed and restarted: `--plugin-dir` points at the live repo tree, so commits I pushed mid-run would have left PRs 2–3 on a different plugin than PR 1. My error; the tree was frozen for the clean run. Before killing it, it surfaced two real defects, both fixed in `f08d4f6`:

- the artifact-writer appended **one stray `}`** to grafana's slice input (issue #69's transcription defect, byte-identical). `die()` raised `SystemExit`, which flew past `_run_receipt`'s `except Exception`, so the script wrote **no output file at all** — the slice degraded with "no file" instead of the reason. Now an honest failure envelope carrying `Invalid JSON in findings file: Extra data: line 1 column 23918`.
- repairing that one byte and re-running the real command gave the first cross-runtime parity proof on real data (13 findings, `fnv1a32:0x65846377` both sides).

### Adversarial pass

6 candidates from 4 independent lenses, each routed to an independent refuter: **5 refuted, 1 confirmed**. The confirmed one — `_delta_confidence` silently rounding a fractional confidence, so the script's on-disk `verified[]` said 64.5 while the delta carried 65 — is fixed at the *input* boundary instead, where `_coerce_numeric_fields` already normalised numerics. Input, script output, delta and joined finding now all carry the same number; there is no divergence left to disclose. A whitespace-id false-degrade class was fixed before its refuter even ran. The degraded path keeping `agent` is real, pre-existing, and filed as **#84**.

workflows/src/stages.js
```diff
@@ -834,65 +937,125 @@ function verifySliceWriterPrompt(entries) {
...
+// build valid envelopes with the REAL computation rather than a second copy of it — a
+// helper that re-derives the canonicalisation would happily agree with a broken one. The
+// cross-runtime half is pinned separately, by the golden fixture whose checksum
+// verify_findings.py itself produced.
+export function deltaContentProof(ids, deltas) {
+  const byId = new Map();
+  for (const d of Array.isArray(deltas) ? deltas : []) {
+    if (d && typeof d.id === 'string') byId.set(d.id.trim(), d);
+  }
+  return fnv1a32(JSON.stringify(canonicalDeltas(ids || [], byId), null, 2));
```
leehopper, line 1000: **🟡 [MEDIUM] deltaContentProof trims delta ids before keying, contradicting the file's own "matched EXACTLY, everywhere" claim**

The comment directly above dispatchableIds (stages.js lines 766-776, duplicated in pipeline.js ~3226) states ids are matched EXACTLY at every site — dispatchableIds, trustSlice's coverage check, and joinVerifyDeltas — and that trimming was tried and removed because "the proof compares the id text the script wrote, so an echo that added or stripped whitespace failed there instead." That statement is false for deltaContentProof itself: it builds its lookup map with byId.set(d.id.trim(), d) while canonicalDeltas looks entries up with the untrimmed dispatched id from `ids`. For any finding whose id legitimately carries surrounding whitespace (explicitly treated as a valid, distinct id by dispatchableIds and pinned by the 'ids differing only by surrounding whitespace are distinct' test), the trimmed key never matches the untrimmed lookup, canonicalDeltas silently substitutes an empty object for that entry, and the recomputed checksum diverges from the one verify_findings.py computed over the raw (untrimmed) id — even for a perfectly honest echo. The JS-only tests don't catch this because deltaEnvelope/verifyDelta.js computes its "expected" checksum by calling this same deltaContentProof function, so the bug is tautologically invisible there, and none of the seven Python/JS golden parity fixtures use a whitespace-containing id.

**Suggested fix:**
Drop the `.trim()` in deltaContentProof's byId.set call (use the raw d.id as the key, matching dispatchableIds/trustSlice/joinVerifyDeltas and Python's build_deltas), or update the surrounding comment if trimming here is actually intended and explain why it's the one exception — then add a golden/JS test with a whitespace-containing id that exercises an honest round-trip through deltaContentProof (not one that derives its own expected value from the same function) to catch this class going forward.

workflows/src/stages.js
```diff
@@ -834,65 +937,125 @@ function verifySliceWriterPrompt(entries) {
...
+}
```
leehopper, line 1001: **🟡 [MEDIUM] deltaContentProof trims ids when building its lookup map but canonicalDeltas looks them up untrimmed, and the only test for whitespace-differing ids cannot detect the resulting mismatch**

deltaContentProof builds its internal byId map with `byId.set(d.id.trim(), d)` but canonicalDeltas looks entries up with the untrimmed dispatched id (`byId.get(id)`), so any dispatched finding id containing leading/trailing whitespace is silently dropped from the canonical form (and two ids differing only by whitespace collide into one map entry), producing a checksum that will not match what verify_findings.py's deltas_checksum computed honestly, causing the slice to be needlessly degraded. The only existing test for this scenario, stages_verify_delta.test.js's whitespace-id test, builds its mock envelope's checksum via the same deltaContentProof function it later asserts against, so it can never catch this defect — the two sides always agree with each other by construction, regardless of whether the id association is correct.

**Suggested fix:**
Add a test (ideally a golden/parity fixture recorded from the real Python build_deltas/deltas_checksum, not a JS-only mock) with two dispatched findings whose ids differ only by surrounding whitespace and DIFFERENT delta decisions, and assert that deltaContentProof(ids, pythonProducedDeltas) equals the checksum Python actually computed. That would surface the trim/no-trim mismatch. Separately, fix canonicalDeltas/deltaContentProof to use the identical (untrimmed) matching everywhere, consistent with dispatchableIds' documented policy.

**Cited rule:** CLAUDE.md > Verify boundary (delta echo): the checksum reuses assemble_artifacts.py's fnv1a32/js_stringify_pretty and tests/test_assemble_artifacts.py is the cross-runtime parity guard for this boundary

leehopper, line 1001: **💡 [LOW] deltaContentProof trims ids when indexing but not when looking up, breaking the checksum for whitespace-padded finding ids**

deltaContentProof builds its lookup map with byId.set(d.id.trim(), d) (trimmed key) but canonicalDeltas looks entries up with the untrimmed dispatched id (byId.get(id)), since dispatchableIds keeps ids exactly as-is (only checking that trim() is non-blank). For any dispatched finding whose id has leading/trailing whitespace, the lookup misses and canonicalDeltas silently substitutes an empty object, producing a checksum that will not match the one verify_findings.py computed (which never trims). Two ids differing only by whitespace also collide into the same trimmed key, so one delta's values silently overwrite the other's inside the map before the (still-failing) lookup. The net effect is that a completely honest, faithful executor echo for a whitespace-padded id gets rejected by trustSlice's content-proof check and the whole slice is falsely degraded to unverified.

**Suggested fix:**
Build the byId map in deltaContentProof with the untrimmed id (`byId.set(d.id, d)`), matching joinVerifyDeltas and the stated 'match ids exactly, everywhere' design; add a cross-runtime parity fixture in tests/fixtures/parity/verify_deltas/ with a whitespace-padded id so the Python-computed checksum would catch a regression here instead of only the self-referential JS test.

**Cited rule:** Verify boundary (delta echo): '_DELTA_FIELDS (Python) and DELTA_KEYS (JS) are one list in two runtimes... a field added to one and not the other either silently drops out of the content proof or makes every honest receipt fail it' — the same cross-runtime-agreement principle is violated here for id-keying rather than field-keying.

leehopper, line 1001: **💡 [LOW] deltaContentProof trims ids before keying its map but canonicalDeltas looks them up untrimmed, breaking cross-runtime checksum parity for whitespace-padded ids**

deltaContentProof builds byId with byId.set(d.id.trim(), d) but canonicalDeltas then looks entries up with the exact untrimmed dispatched id (byId.get(id)), so any dispatched finding whose id carries leading/trailing whitespace gets silently mapped to {} (or, if two ids trim to the same value, to the wrong sibling delta) instead of its real delta content; verify_findings.py's deltas_checksum never trims and computes over the dispatch-order list directly, so the JS recomputation will diverge from the script's authoritative checksum and the whole slice will spuriously degrade to UNVERIFIED even though the executor echoed the script's output honestly, which directly contradicts the file's own "ids matched EXACTLY, everywhere" design comment above dispatchableIds.

**Suggested fix:**
Key byId in deltaContentProof by the untrimmed d.id (matching canonicalDeltas' lookup and matching the exact-match contract enforced everywhere else in this boundary — dispatchableIds, trustSlice's id-coverage loop, and joinVerifyDeltas all use the raw id), and add a golden/parity case with a whitespace-padded id whose checksum is computed by the real verify_findings.py so the cross-runtime parity test would actually catch this class of drift.

**Cited rule:** CLAUDE.md 'Verify boundary (delta echo)': "The checksum reuses assemble_artifacts.py's fnv1a32/js_stringify_pretty ... so tests/test_assemble_artifacts.py is the cross-runtime parity guard for this boundary as well as the persist one" — the parity guarantee this bug breaks for whitespace ids.

scripts/verify_findings.py
```diff
@@ -912,6 +1003,145 @@ def _resolve_head_sha():
...
+    check against a STALE, DRIFTING or CONFUSED executor, not authentication. The
+    checksum travels in the same envelope as the data it covers, so a Byzantine executor
+    could recompute it — but an LLM transcribing a document cannot, which is precisely
+    the failure this boundary keeps observing (the by-value writer's transcription of
+    findings.json diverged on 3 of 3 measured runs).
+
+    Returns None rather than raising if the deltas contain something unserialisable: an
+    absent proof makes the workflow degrade the slice honestly, where an exception here
+    would take out the whole envelope including the honest failure shape.
+    """
```
leehopper, line 1158: **💡 [LOW] deltas_checksum's documented None-return path (unserialisable deltas) is never exercised by a test**

deltas_checksum's docstring states it returns None rather than raising when the deltas contain something the JS/Python serialisation pair cannot spell identically, specifically so the envelope stays status:'ok' with a disclosed-missing proof rather than the whole receipt collapsing to a failure envelope that would throw away a legitimate verified/eliminated result. No test in TestDeltasChecksum or TestReceiptDeltaEchoEndToEnd constructs a delta containing an unserialisable value (e.g. a non-string origin/severity that assert_js_reproducible rejects) and checks that deltas_checksum returns None and that the resulting receipt still carries status:'ok' with deltas_checksum:null instead of crashing.

**Suggested fix:**
Add a unit test that builds a finding whose origin/severity is set to a non-string, JS-unspellable value (bypassing the normal string-producing code paths), calls build_deltas/deltas_checksum on it, and asserts deltas_checksum returns None; add an end-to-end subprocess test asserting the receipt envelope in that case still reports status:'ok' with receipt.deltas_checksum null rather than status:'failed'.

workflows/src/stages.js
```diff
@@ -834,65 +937,125 @@ function verifySliceWriterPrompt(entries) {
...
+
+  const expected = new Set(ids || []);
+  const byId = new Map();
+  for (const d of result.deltas) {
+    if (!d || typeof d !== 'object') return { ok: false, reason: 'delta entry is not an object' };
+    const id = typeof d.id === 'string' ? d.id : '';
+    if (!id.trim()) return { ok: false, reason: 'delta entry has no id' };
+    if (!expected.has(id)) return { ok: false, reason: `delta names a finding this slice did not dispatch (${id})` };
+    if (byId.has(id)) return { ok: false, reason: `delta repeats a finding id (${id})` };
+    if (typeof d.verified !== 'boolean') return { ok: false, reason: `delta ${id} has no boolean verified flag` };
```
leehopper, line 1059: **💡 [LOW] trustSlice's new malformed-delta-entry guards (non-object entry, missing id, missing/non-boolean verified) have no direct test**

trustSlice's per-delta validation loop added by this PR rejects a delta array entry that is not an object, one whose id is blank/missing, and one whose verified field is not a boolean, each with its own distinct reason string, but the test suite (stages_verify.test.js and stages_verify_delta.test.js) only exercises the coverage, stamp, content-proof and confidence-type failure classes — none of these three defensive branches is hit by any dispatched envelope in the tests I reviewed.

**Suggested fix:**
Add three targeted tests in stages_verify_delta.test.js: (1) an envelope whose result.deltas contains a bare string/number instead of an object, asserting the slice degrades with 'not an object'; (2) a delta object with a blank/missing id, asserting 'has no id'; (3) a delta whose verified field is a string like 'true' instead of a boolean, asserting the 'no boolean verified flag' reason.

scripts/verify_findings.py
```diff
@@ -851,17 +917,42 @@ def sort_key(f):
...
     for key in _NUMERIC_FIELDS:
         value = finding.get(key)
         if isinstance(value, str) and _INT_RE.fullmatch(value.strip()):
             finding[key] = int(value.strip())
+            continue
+        if isinstance(value, bool) or not isinstance(value, float):
+            continue
+        if value != value or value in (float("inf"), float("-inf")):
+            continue  # NaN/inf: leave it for the guards below to reject as before
+        finding[key] = int(math.floor(value + 0.5))
```
leehopper: **💡 [LOW] Manual NaN/inf check duplicated in two functions instead of using math.isfinite**

Both _coerce_numeric_fields (line 953) and _delta_confidence (line 1067) independently spell out the same non-obvious NaN/inf test, `value != value or value in (float("inf"), float("-inf"))`, immediately before the identical half-up rounding `int(math.floor(value + 0.5))`. The module already imports math for the floor call, and math.isfinite(value) is the standard, self-documenting way to express exactly this predicate (False for NaN and both infinities, True otherwise), removing the need for the `value != value` NaN idiom and the finite-set-membership check, and removing the duplication across the two call sites in the same file.

**Suggested fix:**
Replace both occurrences of `value != value or value in (float("inf"), float("-inf"))` with `not math.isfinite(value)`, e.g. in _coerce_numeric_fields: Before: `if value != value or value in (float("inf"), float("-inf")): continue` / `finding[key] = int(math.floor(value + 0.5))`. After: `if not math.isfinite(value): continue` / `finding[key] = int(math.floor(value + 0.5))`. Optionally also extract the shared round-or-reject step into one helper, e.g. `def _half_up_int_or_none(value): return int(math.floor(value + 0.5)) if math.isfinite(value) else None`, used by both _coerce_numeric_fields and _delta_confidence, so the NaN/inf handling and rounding rule live in exactly one place.

leehopper (commented):
Code Gauntlet review of PR #25 PR2 (verify delta echo). 7 findings survived discovery, validation, and blind challenge with no contested or downgraded results. All center on the same root defect: deltaContentProof (workflows/src/stages.js) trims delta ids before keying its internal map, but canonicalDeltas looks entries up with the untrimmed dispatched id -- so a whitespace-padded finding id silently maps to {} and the recomputed checksum diverges from verify_findings.py's honest (untrimmed) computation, even for a faithful echo, contradicting the file's own 'ids matched EXACTLY, everywhere' comment. The only existing whitespace-id test is self-referential (it derives its expected checksum from the same buggy function) so it cannot catch this. Two secondary test-coverage gaps and one style/DRY simplification round out the set. 

**Gap:** the security-reviewer agent failed to dispatch this run (server overload, 529) -- the security dimension was not covered.

---
Generated by code-gauntlet | Reviewed up to: f08d4f62a74b67b0bb6dfbb6d98fc6b50ad34461

Issue #85, opened 2026-07-29
bench --check can never pass a PR that failed and was then successfully retried


leehopper:
## Problem

`bench/run.py --check` scans **every** `workflows/wf_*.json` in a PR directory. `--retry-failed` leaves the failed attempt's record next to the successful one, so its degrade signals are read forever and the run can never go green — even though the PR's current, terminal outcome is clean.

`bench/runner/check.py`:

```python
def _iter_workflow_records(pr_dir):
    ...
    return sorted(wf_dir.glob("wf_*.json"))
```

No notion of which attempt is current; no timestamp ordering; no exclusion of superseded records.

## Observed

`smoke-20260729-193917-f08d4f6`, grafana#80329. The first attempt died to a sustained **API 529 Overloaded** (every dispatch from `verify-input-writer-0` onward returned null) and the runner recorded `timeout`. `--retry-failed` then ran it clean.

Both records sit in `pr-grafana-grafana-80329/workflows/`:

| record | timestamp | gaps | checker degrade hits |
|---|---|---|---|
| `wf_744f098e-8c6.json` (the retry, current) | 21:44 | 1 (a *disclosure* gap, no UNVERIFIED token) | **0** |
| `wf_a62f7348-a3f.json` (superseded, 529-killed) | 20:12 | 11 | **1** — `writeArtifacts: writer returned null — artifacts not persisted (partial-artifacts)` |

Run-level result after the retry:

```
status: ok=3
Check ...: ok=False pr_dirs=3 comments=31 findings_files=3
           workflow_records=4 script_paths=4 unknown_origin=0
  FAIL: pr-grafana-grafana-80329: writer degrade signal in workflows/wf_a62f7348-a3f.json
```

Every gate passes on current state — 3/3 terminal `ok`, 3/3 findings artifacts, `unknown_origin=0`, 31 comments, plugin identity on 4/4 — and the single FAIL reads an attempt that no longer represents anything.

Note `workflow_records=4` for a 3-PR run is itself the tell: the count already discloses the extra record, nothing acts on it.

## Why it matters

[...]

PR #87, opened 2026-07-29
fix(verify)+feat(report): slice-input content proof, and the deterministic degradation banner (#25 PR3+PR4)

leehopper:
Closes the third of issue #25's four PRs: **requirement 4** (slice-input proof), **requirement 5** (the length-vs-hash decision and the writeArtifacts generalization assessment), and **requirement 6** (the writer contract line — which turns out to already exist; see below).

## The gap

PR2 made the delta echo prove what the verify executor **said**. Nothing proved what it **read**. The `--input` file crosses an artifact-writer agent's transcription, and `materializeVerifySlices`' write proof only ever confirmed that a *path* came back in the echo — never that the bytes at that path were the bytes dispatched.

That is not a theoretical gap. Measured over every verify slice-input file this repo has retained:

| | |
|---|---|
| Unparseable slice-input files | **4 of 31 (12.9%)** |
| Signature | every one is a complete document with bytes appended after its final byte — `}` ×3, `</content></invoke>` ×1 |
| Recorded cost | `bench/MEASUREMENT.md`: "2 of 3 PRs — 23 findings lost" |

And the retry could not recover any of them: it re-dispatches the **executor**, never the **writer**, so the same bad file is read twice. `wf_3f640577-31c` failed both attempts at the *identical byte offset*.

## The change

`verify_findings.py` reports `receipt.input_checksum` over the canonical form of the document it parsed; the workflow grades it against its own expectation for what it dispatched.

**No new CLI flag.** The proof rides the receipt the script already emits, so an older `verify_findings.py` meeting a newer bundle degrades to "unproven" rather than dying on an unknown argparse flag — which matters here, because this repo has already measured stale plugin copies hijacking bench children. Mechanism lives in the script, policy in the workflow: the same split `assemble_artifacts.py` + `trustAssembleReceipt` already use.

- **Detection** — a file that parses cleanly but drifted in *value* (right ids, right count, wrong text) passed every existing guard silently. It now mismatches.
- **Recovery** — the measured trailing-bytes corruption is accepted, but *only once the content proof confirms the recovered document is the one dispatched*. That is what makes lenient parsing a proven recovery rather than a guess. Disclosed either way.
- **Degradation** — an input-implicated failure re-materializes that slice's input before the single existing retry (no extra executor attempt). A slice whose input never comes back proven degrades to `origin='unknown'` like every other untrusted slice.

The re-materialize trigger is deliberately narrow. A rewrite is not free — the writer's own transcription diverged from its payload on **3 of 10** instrumented runs (−8, −16, +8 chars) — so only the script's structured `reason` code or a proven mismatch spends it. A dropped nonce echo still takes the plain retry over the unchanged file.

## Decisions worth reviewing

**Canonical-value proof, not raw-bytes.** Keyed to the writer's *wire contract*, not the file type: slice inputs are handed to the writer as an object to serialize, so the bytes are its choice and only the parsed value is the contract. `findings.json` is JSON too and stays on the raw-bytes side because it is handed over pre-serialized. Verified on real data — all 27 parseable slice-input files produce identical checksums in both runtimes.

**A residual mismatch degrades rather than shipping its verdicts.** An earlier draft kept the verdicts behind a run-level flag, by analogy to the tolerated *primary* content-proof mismatch on the persist path. That analogy is wrong: a persist primary is still complete, self-consistent content that gets delivered as-is, whereas here the script computed its verdicts *from data known not to be ours*. Degrading also keeps the failure inside the signal bench already watches (G3 counts `origin=unknown`) instead of inventing a softer state no gate can see.

**Requirement 6 was already satisfied** — `agents/artifact-writer.md:39` carries the exact line, landed incidentally in #38 (`348901a1`). Cited rather than duplicated.

**Requirement 5's generalization question** is answered as a written rule rather than more machinery: proof kind follows the wire contract (new CLAUDE.md bullet). The remaining unproven path is the legacy by-value writer, which stays that way by design — it is the pathological-input safety net.

## Measurement, added

`verifyStage` now keeps an `inputProof` ledger (`slices/proven/unproven/recovered/rewritten/degraded`) that rides the envelope, and `bench --check` reports it **structurally** from `result.stats.inputProof` — a stat, not a gate, because a degraded slice already fails G3 and a second verdict would double-count it. Absent on pre-PR3 runs and reported as "not measured", never as zeros. This is the slice-input drift rate the repo previously had no way to measure.

## Verification

Suites-only tier plus an adversarial review pass, per the issue's measurement policy — the happy path is byte-identical to today, and #25's functional smoke ran against PR2.

- `python -m pytest tests/ -q` — **1014 passed**, 309 subtests
- `node --test workflows/test/*.test.js` — **588 passed**
- `python -m pytest bench/tests/ -q` — **535 passed**, 22 subtests
- Bundle freshness green; `tests/test_assemble_artifacts.py` gains a class pinning the two proof **endpoints** against each other across runtimes (astral, lone surrogates, U+2028/9, control chars, 60-level nesting).
- New `workflows/test/stages_verify_input_proof.test.js` — one test per row of the policy table. Mutation-checked: disabling mismatch detection fails 2 of them, so the detection is genuinely covered rather than merely exercised.
- New `tests/test_agent_contracts.py::TestExecutorReceiptEnumeration` reads `VERIFY_SCHEMA` as the source of truth and fails if `agents/executor.md` stops naming a declared receipt field — the #47 lesson applied to this boundary.

Note for reviewers: `tests/test_verify_findings.py::test_malformed_input_yields_the_honest_failure_envelope_not_silence` changed **fixture**, not guarantee. Its old fixture (the incident's stray `}`) now recovers by design, so it moved to a truncated document, which is still fatal — and the recovery got its own named test.

https://claude.ai/code/session_017TDkW35gGT32gtMKxCwEvL

---

> [!NOTE]
> **Medium Risk**
> Changes sit on the verify trust boundary and PR delivery disclosure path; mistakes could false-degrade slices or fail smokes, but behavior is heavily tested and mostly additive guards rather than refactors of core classification logic.
> 
> **Overview**
> **Verify slice inputs** now carry a cross-runtime content proof: `verify_findings.py` reports `input_checksum` (and optional `input_trailing_bytes`) over the parsed document, accepts trailing garbage only in receipt mode, and tags input-file failures with structured `reason` codes. Agent contracts (`executor`, `artifact-writer`) and tests pin receipt field enumeration and Python/JS parity so executor echoes cannot silently drop proof fields.
> 
> **Workflow behavior** (documented in validation pipeline / CLAUDE): after delta trust checks, the verify stage grades the input proof, may **re-materialize** one slice’s `--input` via the artifact-writer when input is implicated, then retries the executor once; unproven or mismatched inputs still degrade slices to unclassified origins.
> 
> **Degraded review disclosure** moves the banner out of Phase 8’s `review_body` into **`health_banner`**, with `post_review.py` prepending it idempotently. Persist/assemble schemas and delivery docs reflect the split so summaries cannot overwrite degradation text.
> 
> **Bench smoke** extends gate 3 to fail when a degraded run lacks the health banner sentinel on **either** `code-gauntlet-report-*.md` or `code-gauntlet-post-review-*.json` (union of unclassified findings and `stats.health.degraded`). `--check` also prints non-gating **`input_proof`** and **`health`** aggregates from `wf_*.json`, picking the newest record by `timestamp` instead of glob order.
> 
> **Report-writer** instructions rename the secondary bucket to “not blind-challenged” and forbid inventing pipeline-health narrative (banner/methodology come from the envelope elsewhere).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e80b6ef6dfb260b0a87d62c291d2174194b3ff90. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:
## PR4 is now on this branch too (#25 requirements 7–10)

Heads-up on packaging first: I intended to stack PR4 as its own PR based on this branch, but separating them now needs a force-push to this branch, which my permissions here refuse. Rather than work around that I've left both on one branch and retitled. **If you'd rather review them separately, say so and I'll split it** — the PR4 commits are `92bad5a`, `0615691`, `beba59f`, `acee083`, `99a6d73`, all cleanly after PR3's last commit `94b58b4`.

### The defect

Twice on the record a run shipped a report claiming **"0 unverified/pipeline-degraded findings"** while every finding it delivered carried `origin=unknown`. The number wasn't wrong about what it measured — the challenge-skipped bucket really was empty — it was a bucket count wearing a whole-pipeline health claim.

### The fix

Health is now derived in code from the delivered findings' own `origin` values and prepended as an unmissable banner. The report-writer is forbidden from characterising health, because **it cannot**: it is handed findings and challenge buckets, never the verify/discover outcomes, so any health sentence it writes is composed from data that does not contain the answer.

The `unverified` bucket is relabelled to what it actually is — *not blind-challenged* (cap overflow, or challenger unavailable). The **key** stays `unverified`: the checkpoint round-trip pins its name and position.

### Four design calls the issue didn't anticipate

| | |
|---|---|
| **`origin` can be ABSENT, not just `'unknown'`** | It isn't in `VERIFY_SCHEMA`'s delta `required` list, and `joinVerifyDeltas` copies only keys the echo carried — so an executor that omits it leaves a finding with no origin at all. Keying on `=== 'unknown'` would have missed that whole class. Mutation-checked: the naive predicate fails 2 tests. |
| **The banner is applied at the persistence boundary, not in `reportStage`** | A resume replays a prior report object and never calls that stage — banding there would leave exactly the runs most likely to be degraded unbannered. It's idempotent, because a replayed body may carry the *previous* run's banner and a stale banner is worse than none. |
| **On a resume, stage-level signals are withheld** | A resume re-runs discover/verify over a population it then discards. Quoting that recomputation would print one population's health above another's results. The findings' own origins travel through the checkpoint, so they're the primary signal on every path. |
| **Unproven slice inputs do NOT raise the banner** | PR3 established unproven as the pre-existing baseline, not a degradation. They're named once it fires for another reason. A banner that cries wolf is a banner nobody reads. |

### Two evasions the adversarial pass caught, both fixed

**The banner only reached `report.md`.** But `pr_comments` is a legal standalone delivery mode and the realistic one for an automated PR bot — in that mode nobody ever opens `report.md`. A degraded run was posting findings to GitHub with no disclosure anywhere a GitHub reader would look. The banner now also rides the post-review wrapper's `review_body`.

**The stale-banner strip searched the whole document.** This tool reviews its own repository, where the sentinel literals are source code — so a finding quoting them deleted everything between the quoted begin and the next end, on healthy runs too. The strip is now anchored at position 0: report content is structurally out of its reach rather than relying on nobody quoting it. Mutation-checked.

### Detection

`bench --check` pairs the two halves G3 left open: a PR with an unclassified delivered finding must disclose on **either** delivery surface. EITHER, not BOTH — delivery mode decides which one a human sees, and on the empty-report path the report is deliberately nulled, leaving `review_body` as the only surface. `stats.health` is reported alongside, absent meaning not-measured.

Also fixed a quieter instance of the same defect class in the report template: `{summary or "Clean"}` per dimension rendered a dimension whose agent never ran as **Clean**, and "Failed/skipped agents | {list or 'none'}" asked for a fact the writer isn't given, defaulting to a false clean bill of health. "Prompt injection" was checked and **kept** — `filter.stats.injections_removed` really does reach the writer.

### Verification

`1564 pytest + 331 subtests` · `611 node --test` · bundle-freshness green. The banner sits **inside** the report's content proof, so PR3's persistence guarantee covers PR4's disclosure.

The one thing I'd most want a second opinion on: a residual mismatch and an unclassified finding both now *degrade* rather than ship quietly. That's the right default for a review tool, but it does mean more runs will visibly say they're degraded than before — which is the point, and also a change in what the output looks like day to day.

leehopper:
### Two late fixes, and one commit-hygiene note

**Bench: the health snapshot was picked by filename order.** A PR with several `wf_*.json` records needs exactly one chosen — health is a snapshot of the delivered review, not a per-dispatch delta — and "last in sorted glob order" chooses arbitrarily, because record filenames are `wf_<random>`. Measured on `smoke-20260729-193917-f08d4f6`, the record that sorts **last is the oldest of four by 90 minutes**, so on precisely the PRs that were retried the checker reported a dead attempt's health as the run's:

```
glob-order pick  : unclassified=5   <- old behaviour
timestamp pick   : unclassified=0   <- newest record
```

Selection is now by the record's own ISO8601 `timestamp`; with nothing timestamped to order by, a record already reporting `degraded: true` wins, because under-reporting degradation is the wrong direction to fail in for a disclosure signal. This is the same currency problem **#85** files against `_iter_workflow_records` generally — fixing it there changes existing gate behaviour, so this narrows it for this stat only and cites the issue. Both new tests are written to *fail* under the old implementation, using the real filenames and timestamps from the run that exposed it, plus an explicit precondition assertion that glob order is not chronological so the fixture cannot quietly stop exercising the bug.

**Report template: "Review scope" removed.** Last of five rows the writer has no data to fill. Worth naming because it looks harmless and is not — an unfilled `Review scope: Full` is a completeness claim in exactly the shape `Failed/skipped agents: none` was: the unknown case rendering as the healthy one. `reportPrompt` serializes exactly `{summary, findings, unverified, stats}` and nothing else. `Prompt injection` stays — `filter.stats.injections_removed` genuinely reaches the writer.

**Hygiene:** commit `40fd111`'s message describes only the bench fix, but it also carries that one-line `report-format.md` change — I swept it in with `git add -A` while both landed together. The change is correct and reviewed; the message just under-describes it. Amending would need a force-push this session cannot make, so I am recording it here rather than leaving it for a reviewer to trip over.

**Final:** `1572 pytest + 331 subtests` · `611 node --test` · bundle-freshness green.

leehopper (commented):
## Code Gauntlet Review

**9 high-confidence findings, all Low severity.** All survived validation, filtering, and blind challenge.

Findings cluster in three areas: the new `inputProofGap()` / `gradeInputProof` logic in `verifyStage`, the legacy bare-array delivery path that was not updated alongside the `health_banner` wrapper, and some duplicated helper bodies in the bench checker.

Two notes on this review's provenance, since they affect how much weight to give it:

1. **This review was recovered, not delivered normally.** The run completed all eight stages, but the artifact-writer corrupted `findings.json` during persistence, so `post-review.json` was never derived and the pipeline could not post. The findings below were recovered byte-exact from the writer's own dispatch payload and re-assembled through `assemble_artifacts.py`, with both content proofs matching. That corruption is the defect this PR's own series is adjacent to, and the durable fix is queued separately.

2. **A couple of findings are near-duplicates** (`inputProofGap` at `stages.js:962` appears twice; the `health_banner` legacy-path gap appears at both `phase8-delivery.md:89` and `:98`). Cross-agent dedup did not collapse them. Treat them as one issue each.

Nothing here is blocking on its own — they are documentation drift from the terminology rename, a nullability/schema mismatch, and test-coverage gaps in the new verify-input-proof machinery.

---

### Two findings that could not be anchored inline

GitHub only accepts inline comments on lines present in the diff, so these two are reported here instead of being dropped:

- **`workflows/src/stages.js:3269` — no test covers the degradation banner when the report is simultaneously empty and the review is degraded.** In `runWith`, `bannered = !emptyReport` gates whether the banner is applied; the interaction of the `empty_report` guard with a degraded review is untested.
- **`scripts/assemble_artifacts.py:29` — the module docstring's plan-wrapper shape is stale.** It documents a wrapper that no longer matches what `persistPlan` emits.

---
Generated by code-gauntlet | Reviewed up to: 476a06c2eb4bcf936608d34b0526c344c791f961

workflows/src/stages.js
```diff
@@ -753,9 +929,46 @@ async function dispatchVerifySlice(c, inp, i, slice, { model, headShaShort, slic
...
+  // The unproven case is NOT reported here. It is a per-RUN condition in practice — an
+  // older verify_findings.py, or an executor habitually dropping the field, makes every
+  // slice unproven at once — and one near-identical gap per slice would bury the gaps
+  // channel in a message that says the same thing N times. verifyStage emits a single
+  // aggregated gap instead (see unprovenInputGap).
+  if (proof.trailingBytes > 0) {
+    return `verify-input-recovered: slice ${i}: the slice-input file carried ${proof.trailingBytes} byte(s) after a complete JSON document (the artifact-writer transcription defect of issue #69). The recovered document's content proof MATCHES what the pipeline dispatched, so its verification stands.`;
+  }
+  return null;
 }
```
leehopper, line 1018: **💡 [LOW] inputProofGap() asserts a false content-proof match when the slice input is actually UNPROVEN but happens to carry trailing bytes**

gradeInputProof() computes `trailingBytes` independently of `state` (unproven/match/mismatch): it is derived solely from `receipt.input_trailing_bytes` regardless of whether `expected` (the workflow's own proof) is null or `got` (the executor's echoed `input_checksum`) is empty. inputProofGap() only checks `proof.trailingBytes > 0` before emitting a gap claiming the recovered document's 'content proof MATCHES what the pipeline dispatched' -- it never checks `proof.state === 'match'`. So when a slice is genuinely UNPROVEN (e.g. the executor drops `input_checksum` while still echoing `input_trailing_bytes`, or the workflow's own expectation is null due to an unrepresentable number) AND the receipt also reports nonzero trailing bytes, the per-slice gap falsely asserts the recovery was proven, directly contradicting the correctly-computed aggregate 'verify-input-unproven' gap emitted by verifyStage for the same slice.

**Suggested fix:**
Guard inputProofGap with `if (proof.state === 'match' && proof.trailingBytes > 0)` before emitting the 'recovered/MATCHES' message, so an unproven slice with stray trailing-byte metadata is never described as content-proven.

leehopper, line 1018: **💡 [LOW] inputProofGap never checks proof.state, so an UNPROVEN slice with trailing bytes would be mis-disclosed as a proven recovery, and no test catches it**

gradeInputProof can return state:'unproven' (expected checksum is null because the pipeline's own document has an unrepresentable number) while still reporting a nonzero trailingBytes count from the script's receipt. inputProofGap only branches on `proof.trailingBytes > 0`, not on `proof.state`, so in that combination it emits the 'verify-input-recovered ... content proof MATCHES what the pipeline dispatched' message even though there is no expectation to match against — contradicting the function's own comment that 'the unproven case is NOT reported here.' No test drives trailingBytes>0 together with an unprovable (null-expected) proof to catch this.

**Suggested fix:**
Add a test where the slice content contains a value pinNumericFields does not pin (making sliceInputProofFor return null) AND the mocked script/executor reports input_trailing_bytes > 0. Assert the per-slice gap does NOT claim the content proof matches, and that inputProofStats.recovered stays 0 (only proven/unproven should move) for that slice.

skills/code-gauntlet/references/phase8-delivery.md
```diff
@@ -86,7 +86,16 @@ Track which findings were selected (**pr_comment_set**) for Stage 2 shortcut.
...
...r_number, sha, health_banner, review_body, findings }`) — consume it directly: compose your summary into `review_body`, keep its `sha` field (it pins the marker to the commit the review ran against) and its `health_banner` field, and for the **default** selection pass the file to `post_review.py` unchanged. For **"Let me pick"** the user's deselections apply to the wrapper too: replace the wrapper's `findings` array with the user's chosen subset (a strict subset of the wrapper's entries, order preserved — deselection only, never re-ranking or re-filtering), keep every other wrapper field, then post. Only when the artifact is the legacy bare findings array (no prIdentity — e.g. a local-diff review that later gains a PR target) do you hand-wrap: for the **default** selection the "selected findings" are the `artifactPaths.postReview` entries **verbatim** — do not drop, reorder, or cap them; only wrap them with `review_body`, `owner`, `repo`, `pr_number`, and `sha` (the full head SHA the review ran against, from Phase 2 — omitting it leaves `post_review.py` to fall back to `git rev-parse HEAD`, which may not be the commit reviewed). For "Let me pick", they are the user's chosen subset.
```
leehopper: **💡 [LOW] Legacy hand-wrap delivery path omits health_banner, undermining the PR's own goal of disclosing degradation on every delivery surface**

This PR's stated purpose (CLAUDE.md, new bullet) is that a degraded review must disclose itself on whichever surface it actually delivers on, specifically because a pr_comments-only delivery never shows report.md to anyone. The updated phase8-delivery.md text explicitly walks through the 'legacy bare findings array (no prIdentity - e.g. a local-diff review that later gains a PR target)' hand-wrap case and instructs wrapping only with review_body/owner/repo/pr_number/sha - health_banner is not mentioned, even though the compact return's top-level healthBanner value is always computed regardless of prIdentity. post_review.py's compose_review_body reads data.get('health_banner') and silently treats an absent key as no banner, so a degraded review delivered exclusively via PR comments through this documented corner case would ship with the disclosure lost - the exact defect issue #25 req 7 was built to close.

**Suggested fix:**
Add health_banner (sourced from the compact return's top-level healthBanner field) to the legacy hand-wrap instructions in both phase8-delivery.md and SKILL.md's Deliver section, the same way it is already threaded through the prIdentity wrapper path.

**Cited rule:** "Health is derived in code (`reviewHealth`/`applyHealthBanner`) ... so that a `pr_comments`-only delivery, which never shows `report.md` to anyone, still discloses that the review is degraded on the surface it actually delivers on." (CLAUDE.md, Verify boundary section)

skills/code-gauntlet/references/phase8-delivery.md
```diff
@@ -86,7 +86,16 @@ Track which findings were selected (**pr_comment_set**) for Stage 2 shortcut.
...
+> **`review_body` is yours; `health_banner` is the pipeline's. Leave the latter alone.**
+> Compose your summary into `review_body` freely — it persists as `""` and nothing of
+> yours has to share it. When the review is DEGRADED the pipeline fills a SEPARATE
+> `health_banner` field, and `post_review.py` prepends that to your summary at post time,
+> mechanically, the same way it appends the marker footer. You do not copy it, echo it, or
+> work around it — just don't delete the field. It exists because `pr_comments` delivery
+> never shows anyone `report.md`, so this is the only surface on which a GitHub reader
+> learns the review is degraded (issue #25 requirement 7).
```
leehopper, line 101: **💡 [LOW] Legacy bare-array hand-wrap path omits health_banner, dropping the degradation disclosure for that delivery shape**

The PR adds health_banner as the second delivery surface for the degradation banner, and both SKILL.md and phase8-delivery.md were updated to tell Phase 8 to preserve health_banner when consuming the prIdentity-stamped postReview wrapper. But the sibling instruction for the legacy bare-array case (no prIdentity, e.g. a local-diff review that later gains a PR target) still tells Phase 8 to hand-wrap with only review_body/owner/repo/pr_number/sha, with no mention of health_banner. Since runWith's compact return never surfaces the rendered banner markdown at the top level (only the structured stats.health object and the persisted report.md/postReview wrapper carry it), Phase 8 following this instruction verbatim for a degraded run in this rare but real combination would post a PR summary with no banner and no way to reconstruct one without violating the project's own no-agent-composes-health rule.

**Suggested fix:**
Add health_banner to the legacy hand-wrap field list in both SKILL.md and phase8-delivery.md, and surface the rendered banner text (or the health object) in the compact return so Phase 8 can attach it even when no PR identity was stamped at pipeline-run time.

bench/runner/check.py
```diff
@@ -239,6 +321,128 @@ def _extract_script_paths(path):
...
+    try:
+        with open(path, encoding="utf-8") as fh:
+            data = json.load(fh)
+    except (OSError, json.JSONDecodeError, ValueError):
+        return None
+    if not isinstance(data, dict):
+        return None
+    result = data.get("result")
+    if not isinstance(result, dict):
+        return None
```
leehopper: **💡 [LOW] _extract_input_proof and _extract_review_health are byte-for-byte duplicate bodies**

Both new functions open the wf_*.json record, json.load it, and walk data -> result -> stats with the identical four isinstance guards; they differ only in which stats key they pull out at the end (inputProof vs health) and in their docstrings, one of which even says it 'mirrors _extract_input_proof exactly'. This is the textbook case for a small shared helper rather than two independently-maintained copies of the same walk, which risks the two drifting the next time either is touched.

**Suggested fix:**
Factor the shared walk into a private helper and have both public-ish names delegate to it: def _extract_stats_field(path, field):\n    try:\n        with open(path, encoding="utf-8") as fh: data = json.load(fh)\n    except (OSError, json.JSONDecodeError, ValueError): return None\n    if not isinstance(data, dict): return None\n    result = data.get("result")\n    if not isinstance(result, dict): return None\n    stats = result.get("stats")\n    if not isinstance(stats, dict): return None\n    value = stats.get(field)\n    return value if isinstance(value, dict) else None\n\ndef _extract_input_proof(path): return _extract_stats_field(path, "inputProof")\ndef _extract_review_health(path): return _extract_stats_field(path, "health")

workflows/src/stages.js
```diff
@@ -532,10 +548,66 @@ const VERIFY_SCHEMA = {
...
+// The workflow's own code for "the file parsed, but it is not the file we sent". The
+// script cannot report this one — it never learns the expectation — so the comparison
+// happens here and this code joins the script's at the re-materialize decision.
+const INPUT_PROOF_MISMATCH = 'input_proof_mismatch';
+
+// inputFaultOf(env) -> code | null. Structural: reads the declared `reason` field, never
+// the message. An envelope that failed for any other cause returns null and takes the
+// ordinary retry, because re-writing a file that was never the problem is not free —
+// the artifact-writer's own transcription drifted on 3 of 10 measured runs, so a
+// needless rewrite can turn a fine file into a bad one.
```
leehopper: **💡 [LOW] New comments cite an unsourced '3 of 10 measured runs' figure that contradicts the repo's established '3 of 3' writer-transcription statistic**

This PR adds several new comments (workflows/src/stages.js lines 581 and 865, mirrored in the generated workflows/pipeline.js bundle, and in workflows/test/stages_verify_input_proof.test.js) stating the artifact-writer's transcription 'drifted on 3 of 10 measured runs.' The same repository already carries a specific, cited measurement of the identical fact - the by-value writer's transcription of a dispatched payload diverging - as '3 of 3' runs, in scripts/verify_findings.py's deltas_checksum docstring (line 1255: 'diverged on 3 of 3 measured runs') and in an untouched stages.js comment (line 2334: 'diverging from its payload on 3 of 3 runs'), both traceable to the cited 2026-07-27 3-PR smoke in CLAUDE.md. The new '3 of 10' claim gives no run IDs, dates, or smoke name, breaking this project's own strict measured-not-assumed citation convention, and asserts a materially different drift rate (30% vs 100%) for what reads as the same underlying phenomenon without reconciling the two numbers anywhere.

**Suggested fix:**
Either cite the specific measurement session that produced the '3 of 10' figure (matching this project's own citation discipline used everywhere else in this diff, e.g. the 4-of-31 and 27-of-27 slice-input figures) or reconcile it explicitly with the existing 3-of-3 statistic so a reader does not encounter two conflicting drift rates for the same claim.

bench/runner/check.py
```diff
@@ -616,7 +979,74 @@ def check_run(run_dir, *, repo_root=None, plugin_pipeline=None):
...
+        if wf_records:
+            # pr_health was resolved above, where the G3 banner gate needs it.
+            if pr_health is None:
+                health_unmeasured_prs += 1
+            else:
+                health_measured_prs += 1
+                for k in _HEALTH_INT_FIELDS:
+                    v = pr_health.get(k)
+                    if isinstance(v, int) and not isinstance(v, bool):
+                        health_totals[k] += v
```
leehopper: **💡 [LOW] The int-safe field-summing loop is duplicated verbatim for input_proof and health**

The per-PR aggregation for input_proof (lines 994-1000) and for health (lines 1029-1032) both walk a fixed field tuple, read a value, and only add it into the running total when isinstance(v, int) and not isinstance(v, bool) — the exact same three-line guard, restated twice with different variable names. Extracting a tiny shared helper would remove the duplicate arithmetic-safety logic and mean a future change to the int-safety rule (e.g. also rejecting negative values) only has to be made once.

**Suggested fix:**
Add a small helper: def _add_int_fields(dest, source, fields):\n    for k in fields:\n        v = source.get(k)\n        if isinstance(v, int) and not isinstance(v, bool):\n            dest[k] += v\nThen call `_add_int_fields(pr_proof, proof, _INPUT_PROOF_FIELDS)` and `_add_int_fields(health_totals, pr_health, _HEALTH_INT_FIELDS)` at the two sites, replacing the inlined loops.

leehopper:
Heads-up for the rebase onto #99, which restructures the instruction files.

#99 moves repo rules from `CLAUDE.md` into a canonical `AGENTS.md` plus directory-scoped `workflows/AGENTS.md`, `scripts/AGENTS.md` and `agents/AGENTS.md` (each with a generated `CLAUDE.md` twin — edit AGENTS.md, then run `python3 scripts/sync_agent_rules.py`). Root `CLAUDE.md` is now a 16-line pointer, so this PR's CLAUDE.md additions will conflict and should move rather than merge.

Specifically, three rules belonging to **this** PR were briefly written into #99 and have been removed from it, because they describe code that only exists here:

- `INPUT_FAULT_REASONS` as a cross-runtime list alongside `_DELTA_FIELDS`/`DELTA_KEYS` → `workflows/AGENTS.md`, "The verify boundary"
- the proof-kind rule (an artifact handed over as a pre-serialized **string** gets the raw-bytes proof; one handed over as an **object** gets the canonical-value proof) → same section
- "Health is derived, never asked for" → `workflows/AGENTS.md`

#99 also adds a guard that every backticked path and symbol in the instruction files resolves in the tree, so adding these back before the code lands will fail the suite — and adding the code without them will simply leave them undocumented. Land both together.

workflows/src/stages.js
```diff
@@ -956,6 +1124,101 @@ async function materializeVerifySlices(c, inp, slices, policy) {
...
+// this slice, or null if it is not computable.
+//
+// Exported for the tests, on the same reasoning the adjacent deltaContentProof export
+// already states: a harness that builds a valid receipt by RE-DERIVING the canonical form
+// would happily agree with a broken implementation. Test mocks must call the real
+// computation, through the same single construction site the dispatch uses.
+export function sliceInputProofFor(slice, baseBranch) {
+  return sliceInputProof(sliceInputContent(slice, baseBranch));
+}
+
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — `verifyPrompt` now names `input_checksum`, `input_trailing_bytes`, and failure `reason`, aligned with `VERIFY_SCHEMA` / `agents/executor.md`. Added a regression test that fails on the stale "all four fields" wording.

tests/test_verify_findings.py
```diff
@@ -1453,43 +1458,177 @@ def test_receipt_failure_envelope_is_schema_valid(self):
         finally:
             os.unlink(findings_path)
 
+    def _receipt_for(self, text):
+        """Run receipt mode over `text` as the --input file; return the envelope."""
+        import io, json
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — removed the redundant local `json` import from `_receipt_for` (and the sibling sites).

tests/test_verify_findings.py
```diff
@@ -1453,43 +1458,177 @@ def test_receipt_failure_envelope_is_schema_valid(self):
...
+        self.assertEqual(envelope["status"], "ok")
+        self.assertEqual(
+            envelope["receipt"]["input_checksum"],
+            self._receipt_for(clean)["receipt"]["input_checksum"],
+        )
+        self.assertEqual(envelope["receipt"]["input_trailing_bytes"], len("\n</content>\n</invoke>\n"))
+
+    def test_missing_input_file_is_tagged_unreadable(self):
+        """A missing --input file is an input fault and says so with a code."""
+        import io, json
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — same cleanup: keep the local `io` import, use the module-level `json`.

scripts/verify_findings.py
```diff
@@ -1255,26 +1357,62 @@ def _run_receipt(args):
...
         base_branch = data.get("base_branch") or args.base_branch
         result = run_verification(findings, base_branch, args.diff_file, verbose=False)
         deltas = build_deltas(findings, result["verified"])
+        receipt = {
+            "sha": sha,
+            "n_in": len(findings),
+            "nonce": args.nonce,
+            "deltas_checksum": deltas_checksum(deltas),
+            "input_checksum": proof,
+        }
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — when `input_content_proof` returns `None`, the receipt now omits `input_checksum` entirely instead of emitting `null`, so Structured Output can still echo a legal receipt and the slice stays soft-unproven.

workflows/src/stages.js
```diff
@@ -651,11 +733,39 @@ export async function verifyStage(ctx, input) {
...
     if (!attempt.ok) {
+      // A slice whose input never came back proven degrades like every other untrusted
+      // slice — origin='unknown', findings kept, loud gap. It is NOT delivered with its
+      // verdicts intact and a run-level footnote: verify_findings.py computed those
+      // verdicts from data known not to be ours, and this stage's contract is that every
+      // finding leaves either as trusted output or as itself marked unknown. Degrading
+      // also keeps the failure inside the signal bench already watches (G3 counts
+      // origin=unknown), instead of inventing a second, softer failure state that no
+      // existing gate can see.
+      inputProofStats.degraded += 1;
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — `inputProof.degraded` now increments only when `attempt.inputImplicated` is set. Nonce/trust/executor failures still degrade the slice for G3, but no longer inflate the input-proof ledger. MEASUREMENT.md updated to match.

tests/test_verify_findings.py
```diff
@@ -1453,43 +1459,261 @@ def test_receipt_failure_envelope_is_schema_valid(self):
...
+                "an input fault — the two try/excepts have been collapsed into one",
+            )
         finally:
-            os.unlink(corrupt_path)
+            os.unlink(in_path)
             os.unlink(out_path)
 
+    def test_a_plain_crash_in_the_verification_body_is_still_untagged(self):
+        """The ordinary case, kept beside the discriminating one above."""
+        import io, json
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — redundant local `json` import removed.

workflows/src/stages.js
```diff
@@ -1164,7 +1448,13 @@ export function worstCaseAgentCount(limits, nFiles, nFindings) {
...
   const findings = Math.max(0, nFindings || 0);
   const summarizeCalls = ceilDiv(files, effectiveBucketSize(L)) + 1;
-  const verifyCalls = ceilDiv(findings, effectiveSliceSize(L, findings)) * VERIFY_ATTEMPTS_PER_SLICE;
+  // Per slice: VERIFY_ATTEMPTS_PER_SLICE executor dispatches, plus at most ONE
+  // artifact-writer re-materialize of its --input file before the retry (issue #25 PR3).
+  // The rewrite is counted rather than left to headroom precisely because it scales with
+  // the finding count, which is the property that made this term the exact one — the
+  // report/writer/persist terms below stay nominal because they do not.
+  const verifySlices = ceilDiv(findings, effectiveSliceSize(L, findings));
+  const verifyCalls = verifySlices * (VERIFY_ATTEMPTS_PER_SLICE + VERIFY_REWRITES_PER_SLICE);
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — `coarsenLimits` now sizes the verify term with `VERIFY_ATTEMPTS_PER_SLICE + VERIFY_REWRITES_PER_SLICE`, matching `worstCaseAgentCount`.

workflows/src/stages.js
```diff
@@ -753,9 +925,46 @@ async function dispatchVerifySlice(c, inp, i, slice, { model, headShaShort, slic
...
+// a reader can tell a proven recovery from an unproven guess without reading the code.
+function inputProofGap(i, proof) {
+  // The unproven case is NOT reported here. It is a per-RUN condition in practice — an
+  // older verify_findings.py, or an executor habitually dropping the field, makes every
+  // slice unproven at once — and one near-identical gap per slice would bury the gaps
+  // channel in a message that says the same thing N times. verifyStage emits a single
+  // aggregated gap instead (see unprovenInputGap).
+  if (proof.trailingBytes > 0) {
+    return `verify-input-recovered: slice ${i}: the slice-input file carried ${proof.trailingBytes} byte(s) after a complete JSON document (the artifact-writer transcription defect of issue #69). The recovered document's content proof MATCHES what the pipeline dispatched, so its verification stands.`;
+  }
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — `inputProofGap` now requires `proof.state === 'match' && proof.trailingBytes > 0`, the same predicate as the `recovered` counter. Added a regression test for unproven + trailing bytes.

workflows/pipeline.js
```diff
@@ -3213,9 +3385,46 @@ async function dispatchVerifySlice(c, inp, i, slice, { model, headShaShort, slic
...
+function inputProofGap(i, proof) {
+  // The unproven case is NOT reported here. It is a per-RUN condition in practice — an
+  // older verify_findings.py, or an executor habitually dropping the field, makes every
+  // slice unproven at once — and one near-identical gap per slice would bury the gaps
+  // channel in a message that says the same thing N times. verifyStage emits a single
+  // aggregated gap instead (see unprovenInputGap).
+  if (proof.trailingBytes > 0) {
+    return `verify-input-recovered: slice ${i}: the slice-input file carried ${proof.trailingBytes} byte(s) after a complete JSON document (the artifact-writer transcription defect of issue #69). The recovered document's content proof MATCHES what the pipeline dispatched, so its verification stands.`;
+  }
+  return null;
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — same `inputProofGap` guard as the stages.js finding (`state === 'match'`). Bundle rebuilt.

workflows/pipeline.js
```diff
@@ -3127,7 +3237,29 @@ async function verifyStage(ctx, input) {
     if (attempt.gap) gaps.push(attempt.gap);
   }
 
-  return { findings: out, verified: degradedSlices === 0, gaps };
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — the empty-findings early return now emits a zeroed `inputProof` ledger so post-PR3 zero-slice runs read as measured-at-zero, not "not measured".

workflows/pipeline.js
```diff
@@ -3176,12 +3308,45 @@ function verifyDegradeGap(detail, k, n) {
...
+      c, [sliceInputEntry(inp, i, slice)], writerModel, `verify-input-rewriter-${i}`,
+    );
+    if (!rewrite.ok) {
+      // The file on disk is unchanged and still bad, so a second executor dispatch would
+      // read the same bytes and fail the same way. Degrade now rather than spend it.
+      return {
+        ok: false,
+        rewritten: true,
+        reason: `${first.reason} — re-materializing this slice's input failed (${rewrite.reason}), so the retry was skipped: a second dispatch would have read the same bytes`,
+      };
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — a failed rematerialize now returns `rewritten: false` (file unchanged) with `inputImplicated: true` so the degraded counter still attributes it correctly.

bench/runner/check.py
```diff
@@ -616,7 +927,74 @@ def check_run(run_dir, *, repo_root=None, plugin_pipeline=None):
...
+                    # violates it must not corrupt the aggregate with a float or
+                    # a bool (bool is an int subclass in Python).
+                    if isinstance(v, int) and not isinstance(v, bool):
+                        pr_proof[k] += v
+            if pr_proof is None:
+                input_proof_unmeasured_prs += 1
+            else:
+                input_proof_measured_prs += 1
+                for k in _INPUT_PROOF_FIELDS:
+                    input_proof_totals[k] += pr_proof[k]
```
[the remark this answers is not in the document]
leehopper: Fixed in bc5554b — `input_proof` now picks one snapshot per PR via `_select_pr_input_proof_snapshot` (newest by record `timestamp`), same currency fix as health. Added a regression test that fails under sum-of-retries.

workflows/src/stages.js
leehopper: Fixed in bc5554b — guarded with `proof.state === 'match' && proof.trailingBytes > 0`. Regression test covers null-expected + trailing bytes.

leehopper: Fixed in bc5554b — same guard, plus the test you sketched: unpinnable nested float → null expectation, mocked `input_trailing_bytes > 0`, asserts no MATCHES gap and `recovered === 0`.

skills/code-gauntlet/references/phase8-delivery.md
leehopper: Fixed in bc5554b — legacy hand-wrap instructions in `phase8-delivery.md` and `SKILL.md` now include `health_banner`, sourced from the compact return's new top-level `healthBanner`. Also covered by an empty-report + degraded end-to-end test.

skills/code-gauntlet/references/phase8-delivery.md
leehopper: Fixed in bc5554b — same as the sibling thread: `healthBanner` is now on the compact return, and the legacy hand-wrap field list names `health_banner`.

bench/runner/check.py
leehopper: Fixed in bc5554b — factored the shared walk into `_extract_stats_field`; both extractors delegate to it.

workflows/src/stages.js
leehopper: Fixed in bc5554b — replaced the unsourced "3 of 10" with the established, cited "3 of 3" by-value writer measurement (same figure already in `writeArtifactsDerived` / `deltas_checksum`).

bench/runner/check.py
leehopper: Fixed in bc5554b — extracted `_add_int_fields` and use it for both `input_proof` and `health` aggregation.

leehopper:
## Review comments addressed in bc5554b

Closed every open review thread on this PR (Bugbot + code-quality + the code-gauntlet review). Summary of what landed:

### Correctness
- **`verifyPrompt`** now names `input_checksum` / `input_trailing_bytes` / failure `reason` (was still saying \"all four fields\")
- **Null `input_checksum`** omitted from the receipt (schema is string-typed; null broke the soft-unproven path)
- **`inputProof.degraded`** counts only input-implicated failures; **`rewritten`** only when the rematerialize landed
- **`inputProofGap`** requires `state === 'match'` before claiming MATCHES
- **Empty verify** emits a zeroed `inputProof` ledger (measured-at-zero, not \"not measured\")
- **`coarsenLimits`** includes `VERIFY_REWRITES_PER_SLICE`
- **Bench `input_proof`** picks one snapshot per PR by timestamp (no retry double-count)
- **Legacy hand-wrap** gets `health_banner`; compact return surfaces top-level `healthBanner`
- **Empty report × degraded** end-to-end test added
- **`assemble_artifacts.py`** wrapper docstring includes `health_banner`

### Docs / hygiene
- Restored the three #99-deferred rules into `workflows/AGENTS.md`
- Reconciled unsourced \"3 of 10\" to the cited \"3 of 3\"
- Factored `_extract_stats_field` / `_add_int_fields`; removed redundant local `json` imports

### Verification
`1048` pytest + `409` subtests · `625` node --test · bench check suite green · bundle rebuild idempotent

PR #93, opened 2026-07-30
docs(claude-md): delete the false persistence premise it was all built on

leehopper:
## The premise

`CLAUDE.md`'s "Artifact persistence" section opened with:

> *"The sandbox has no disk, so every persisted byte must be emitted as some agent's tool-call argument at least once. The floor is **one generation pass per unique byte** — not zero."*

**It is false.** True of the current design; not of the platform. It was never measured — it was asserted, and then a whole subsystem was justified by it. Two runs lost their entire review to a writer that cannot transcribe 50 KB of escape-dense JSON, and this premise is why nobody went looking for another channel.

## The measurement

Zero-subagent probe: a workflow builds a deterministic string, computes `fnv1a32` in-sandbox, returns it; the value is then compared against `tasks/<taskid>.output` on disk.

| requested | on-disk chars | file bytes | checksum | byte-exact |
|---|---|---|---|---|
| 200,000 | 200,000 | 261,606 | MATCH | ✅ |
| 500,000 | 500,000 | 653,268 | MATCH | ✅ |
| **4,000,000** | **4,000,000** | **5,222,716** | MATCH | ✅ |

Byte-exact at every size — including a **lone surrogate** left by slicing an emoji in half, which no model would survive. The harness serializes it; nothing retypes it. `await_workflow.py` already resolves and parses that file, and Phase 8 has Bash.

Findings + report top out at **~66 KB — 1.6%** of the proven-safe size. And a recorded run whose writer failed and persisted *nothing* was found with its complete output sitting in that file, recoverable the whole time.

## Two corrections, both against things I shipped or wrote

**`hardenEscapeRuns` (shipped this morning in 3.3.2) did not work in live use.** It removes every backslash run from the wire — verified, 28 `\` sites and zero runs — and the writer rewrote **0 of 28** back to `\\`, producing a fresh unparseable document. Both spellings fail: one gets collapsed, the other expanded. Recorded as a dead end so nobody tries a third encoding. And encoding isn't even the whole fault — the same run showed the writer *expanding an ellipsis* inside a finding's quoted evidence by opening the file the quote referenced. No encoding stops a model editing content it thinks it's improving.

**The retry rationale was wrong.** *"The writer is a sampled agent, not a function, so a second dispatch is a fresh sample"* holds only for a stochastic fault. Twice, on two different documents, the retry corrupted the same sites and failed at the **identical byte offset**.

## Net −2,192 bytes

The correction cost ~1.7 KB and the section still shrank, because the same pass deleted four bullets that failed this file's own two-part test — derived-document content proof, never-throws, JS-reproducible numbers, derived-never-retyped. Each was already stated in full at the code implementing it, and the last was reasoning *from* the false premise. Ceiling lowered from 27,591 to 25,399 to lock it in.

The note left for whoever raises the ceiling next: the two-part test screens for duplication and caught none of this, because unverified claims read exactly like verified ones when written in the same confident voice. Prefer a sentence that names its measurement.

## Verification

`node --test workflows/test/*.test.js` — 585 pass. `python -m pytest tests/ -q` — 1004 pass.

Docs-only; no behaviour change. The persistence redesign this unblocks is a separate PR.

https://claude.ai/code/session_01Pnva1Qsmpqon81QyBj7Wxf

---

> [!NOTE]
> **Low Risk**
> Documentation and test ceiling only; no workflow or persistence code changes in this diff.
> 
> **Overview**
> **Docs-only correction** to `CLAUDE.md`’s **Artifact persistence** section: it no longer opens on the false claim that the sandbox has no disk and every byte must pass through an agent tool call. A blockquote now states the **measured** alternative (harness serializes workflow return to `tasks/<taskid>.output` byte-exact up to 4M chars) and labels the rest as describing the **current writer-based design**, not a justification.
> 
> The bullets are **reframed around writer failure modes** (36% write-contract failures in a 2026-07-30 census, silent summarization, encoding dead ends including `hardenEscapeRuns`, retry not fixing deterministic corruption) while **trimming** four items the edit treats as duplicated at code sites or tied to the old premise (derived-never-retyped, writer never-throws, JS-reproducible numbers path, derived content-proof narrative). Cross-runtime rules (`fnv1a32`, checkpoint guards, assemble/trust policy, legacy by-value path, report unwrap) stay, with wording tightened.
> 
> `tests/test_claude_md_budget.py` **lowers** `CLAUDE_MD_BYTE_CEILING` from 27,591 to **26,212** and documents why (net shrink despite the new premise block).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f63a966ab27b106a64e62c9aa7be18d08258acc3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,15 +40,24 @@ The v3 review pipeline runs inside `workflows/pipeline.js`, invoked from SKILL.m
 > What follows describes the CURRENT writer-based design, which is being replaced. Read it as a
 > description of the code, not a justification for it. Do not cite the deleted premise.
 
-- **The writer is a language model, and it does not transcribe.** Two independent measurements: 2 of
-  5 recorded runs corrupted; and on 2026-07-30 a 53 KB findings.json lost 18 of 18 backslash-run
-  sites, then — after `hardenEscapeRuns` removed every run from the wire — **0 of 28** `\`
-  escapes survived, the writer having rewritten them all back to `\\`. Both spellings fail: one gets
-  collapsed, the other expanded. **No further encoding tweak is worth trying**, and encoding is not
-  even the whole fault: the same run showed the writer expanding an ellipsis inside a finding's
-  quoted evidence by opening the file the quote referenced. Nothing stops a model editing content it
-  believes it is improving. `hardenEscapeRuns` ships because it is harmless and its cross-runtime
-  claim is guarded, but **it is not the fix**.
+- **The writer is a language model, and it does not transcribe. Census over every recorded run
+  (2026-07-30, 38 writer journals / 84 artifacts): 26 of 73 attempted writes — 36% — failed their
+  own contract, plus 12 artifacts never written at all.** Corruption is pervasive, not a slip:
+  median 8 divergence sites per bad document, max 46, only 2 of 10 single-site. **Nothing predicts
+  it.** Not size (a 47 KB findings.json with 104 backslashes came through byte-perfect; a 6.4 KB
+  zero-backslash `.md` was truncated), not escape density, not the two combined — so there is no
+  safe-size rule to design to. And **no encoding tweak is worth trying**: 18 of 18 backslash-run
+  sites collapsed, then after `hardenEscapeRuns` cleared every run from the wire, **0 of 28**
+  `\u005c` escapes survived, rewritten back to `\\`. Both spellings fail, in opposite directions.
+- **The worst failures are SILENT SUMMARIZATION, which no encoding or format can prevent.** Same
+  census: one checkpoint lost 29,132 chars because the writer dropped 11 fields from *every*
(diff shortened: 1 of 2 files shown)


leehopper:
Superseded by #99. This PR corrected the *content* of the CLAUDE.md persistence section but kept the wrong *shape* — it left a quoted tombstone of the false premise and measurement tables in a file loaded on every turn, which is the bloat the file needed removing, not adding to.

#99 restructures instead: AGENTS.md canonical, CLAUDE.md down to a 16-line pointer, path-scoped rules as references rather than prose. The corrections this PR made are preserved where they belong — the persistence facts are in `workflows/AGENTS.md` as current rules with no history, and the full measurement record stays in this PR's commits and in #46.

Issue #98, opened 2026-07-30
AGENTS.md is injected as project rules into every review of this repo but is scoped to Cursor Cloud


leehopper:
## Problem

`AGENTS.md` is a `PROJECT_RULE_FILENAMES` source (`scripts/collect_project_rules.py`, per `references/phase2-triage.md:137`), so its full contents reach all seven discovery agents as project rules on every code-gauntlet review of this repo.

Its only top-level heading is `## Cursor Cloud specific instructions`, and much of the body is environment setup for a VM the review agents are not running on — nvm PATH exports, `/exec-daemon/node`, `pip install --user` locations.

Review agents are therefore being handed environment troubleshooting as if it were review criteria, while the genuinely portable rules in it (lint gate semantics, the skip-ci-token prohibition, the contribution surface) are not labelled as such.

## Scope

Separate what is portable project rule from what is Cursor-Cloud environment setup, and decide where each belongs. Note this interacts with the AGENTS.md/CLAUDE.md restructure — if AGENTS.md becomes the canonical cross-tool rules file, the environment section needs a home that is not injected into reviews.

Also stale while we are here: line 11 claims "CI and `CLAUDE.md` pin Node `24.18.0`", but `.github/workflows/ci.yml:76` pins `node-version: "24"` (major only).

## Verification

Per `bench/MEASUREMENT.md`. Changing what reaches discovery agents can move recall, so escalate accordingly.

leehopper:
Re-scoped (2026-07-30 triage, adversarially re-verified): the portable-rules half and the stale-Node-pin half landed via #99. What remains is one decision this issue's own Scope section asked for: the Cursor Cloud environment-setup content (nvm PATH export, exec-daemon node caveat, pip install --user) was deleted outright rather than relocated, and no .cursor/rules or equivalent picked it up. Decide where — or whether — that content lives. Wave 2 milestone.

Issue #97, opened 2026-07-30
Per-dimension confidence thresholds are documented in review-md-spec.md but parsed by nothing


leehopper:
## Problem

`references/review-md-spec.md:133-153` documents per-dimension confidence thresholds at length. No parser reads them: `scripts/filter_findings.py` and `workflows/src/filterFindings.js` recognise only flat `confidence_threshold` and `security_min_confidence`, and the args waist (`workflows/src/args.js:495-514`) has no field for them.

A user who configures per-dimension thresholds gets no error and no effect.

## Scope

Either implement per-dimension thresholds through the parsers and the waist, or remove the section from the spec. Which way — and whether the feature is wanted at all — is an implementation decision for whoever picks this up.

## Verification

Per `bench/MEASUREMENT.md`. Escalates if filtering behaviour changes.

leehopper:
Consolidated into #94 (2026-07-30 triage, owner-approved): all four REVIEW.md contract defects (#94, #95, #96, #97) share one root cause — the documented dialect and the parser dialect are different languages — and one format decision settles them together. #94's body now enumerates this issue's specific defect; implementation lands there.

Issue #96, opened 2026-07-30
review-md-spec.md states a security confidence minimum of 60; the code constant is 70


leehopper:
## Problem

Template comments at `references/review-md-spec.md:345` and `:357` tell the reader "Security findings always use a minimum of 60" / "Security cannot be set below 60".

**60 appears nowhere in the code.** The default is 70 (`workflows/src/filterFindings.js:143-159`), and the security bar is `Math.min(confidence_threshold, security_min_confidence)` — so it is not a floor at all, it is a ceiling-taking-the-lower-value.

Two wrong claims in one place: the number, and the direction of the guarantee.

## Related trap

`SKILL.md:270` notes that writing an explicit threshold silently raises the non-security bar from the built-in 55. The root template ships `70` and `build-review-md` generates `85`, so following either doc changes filtering behaviour in a way the reader is not told about.

## Scope

Correct the spec text to match the code, and decide whether the templates' shipped defaults are the ones we actually want users to start with. Whether to document the min() semantics or change them is an implementation decision.

## Verification

Per `bench/MEASUREMENT.md`. Docs-only unless defaults change.

leehopper:
Consolidated into #94 (2026-07-30 triage, owner-approved): all four REVIEW.md contract defects (#94, #95, #96, #97) share one root cause — the documented dialect and the parser dialect are different languages — and one format decision settles them together. #94's body now enumerates this issue's specific defect; implementation lands there.

Issue #95, opened 2026-07-30
REVIEW.md ignore entries in the documented dimension:"pattern" form cannot match anything


leehopper:
## Problem

`applyExclusions` (`workflows/src/filterFindings.js:946-957`) regex-escapes each `ignore` entry whole and substring-matches it case-insensitively against `title + "\n" + description`. Nothing anywhere strips a `dimension:` prefix — `grep -rn 'dimension:"' workflows/src scripts` returns no hits.

So a documented entry like `security:"console.log in development mode"` only matches a finding whose text literally contains `security:"console.log in development mode"`. In practice it matches nothing.

Both the bench fixture (`bench/golden/review_md_fixture.md`) and the Phase 8 dismissed-findings write-back (`references/delivery-guide.md:196-250`) emit exactly this inert form, so users who dismiss findings get entries that silently never apply.

## Scope

Either implement the `dimension:` prefix semantics the docs describe, or change the documented/emitted format to the bare substring the matcher actually supports. Which way to resolve it is an implementation decision.

Note the waist already rejects structured entries: `args.js:501-514` requires `reviewConfig.ignore` to be an array of strings.

## Verification

Per `bench/MEASUREMENT.md`.

leehopper:
Consolidated into #94 (2026-07-30 triage, owner-approved): all four REVIEW.md contract defects (#94, #95, #96, #97) share one root cause — the documented dialect and the parser dialect are different languages — and one format decision settles them together. #94's body now enumerates this issue's specific defect; implementation lands there.

PR #99, opened 2026-07-30
docs: make AGENTS.md canonical and give every tool one source

leehopper:
Closes the CLAUDE.md half of #46. Supersedes #93, which I'm closing — it corrected the content but kept the wrong shape.

## Why

CLAUDE.md had become a design-history document — incident narratives, run IDs, measurement tables, version references — **loaded on every turn**. Anthropic's guidance is blunt: *"bloated CLAUDE.md files cause Claude to ignore your actual instructions"*, with a 200-line target and a one-question deletion test: *would removing this cause Claude to make mistakes?*

This repo also takes contributions through Cursor and Codex. Neither reads CLAUDE.md — a fact I verified against Anthropic's docs, which state it outright, and which several widely-cited blog posts get wrong.

## Layout

```
AGENTS.md                          canonical portable rules   (Codex + Cursor, native)
{workflows,scripts,agents}/AGENTS.md   directory-scoped rules
CLAUDE.md                          @AGENTS.md + Claude-only tail   (16 lines)
.claude/rules/*.md                 paths: frontmatter + @import — no prose
.cursor/rules/*.mdc                globs + @reference — no prose
```

Directory nesting is the **only** scoping primitive all four tools implement, so it carries the split. The rule files are pure scoping metadata; content exists once.

**Import, not symlink.** On a Windows checkout without `core.symlinks=true`, git materializes the symlink as a plain text file whose entire content is the string `AGENTS.md` — silent and content-shaped, the same failure class as #48. A symlink also can't carry the Claude-only tail.

**No Copilot file.** It's the one format with no include mechanism, so it would be a real second copy — for a tool this project doesn't use.

## The number that matters

| | before | after |
|---|---|---|
| CLAUDE.md | 178 lines / 26,212 B | **16 lines / 819 B** |
| CLAUDE.md + AGENTS.md | **31,417 B** | — |
| whole AGENTS.md set | — | **13,522 B** |

Codex caps concatenated instructions at **32,768 bytes** and truncates with **no notice**. We were **1,351 bytes** from silently dropping instruction tail on every Codex run.

For scale: #46 measured CLAUDE.md at 76 lines on 2026-07-27. It hit 178 by 2026-07-30 — more than doubling in three days, while #46 was open.

## Three design rules, stated once

Added per owner direction. Several sections were previously re-arguing these incident by incident:

> **Build the mechanism, not the instruction.** Whatever code, a schema, a data structure, or a removed capability can enforce, it must.
> **"Add more text" is a design smell.** If that's the fix under consideration, the shape is wrong.
> **Extending should cost one edit.** If it takes coordinated edits across N files, fix the shape.

## Guards

`tests/test_agent_instruction_layout.py`. Every failure it covers is **silent** — none raise, all leave a file that still looks correct:

- **Codex 32 KiB cap** on the AGENTS.md set.
- **Import expansion** — Claude's parser skips code spans, so one backtick around the pointer turns it into decoration and CLAUDE.md carries nothing.
- **Prose regrowth** — 40-line budget on CLAUDE.md, plus a duplicate-heading check against AGENTS.md.
- **Stale pointers.** Note the two tools resolve from *different bases* — Claude from the importing file, Cursor from the project root — so each spelling is wrong in the other tool. The guard checks per tool. It caught this during development.

`tests/test_claude_md_budget.py` is deleted: a byte ceiling on one file can't see eight rule files or the Codex cap. Its pre-commit hook now points at the new test.

`tests/test_dimensions_registry.py` follows the findings schema to `agents/AGENTS.md` — it caught the move, then caught that my first draft had dropped two enumerations it pins. Working as intended.

`.gitignore`: `.claude/` → `.claude/*` + `!.claude/rules/`. Without this, contributors get none of the path-scoped rules and CI fails the new guard.

## REVIEW.md

Adds this repo's own, dogfooding the tool's config format. It satisfies **both** the fenced block the parsers read and the headings the spec documents, because those disagree — verified to parse identically through `scripts/filter_findings.py` and the JS twin.

Writing it surfaced five contract bugs, filed underspecified: #94 (spec format ≠ parser format), #95 (`dimension:"pattern"` ignore entries match nothing), #96 (spec claims a security minimum of 60; the constant is 70), #97 (per-dimension thresholds parsed by nothing), #98 (AGENTS.md injected into every review while scoped to Cursor Cloud).

## Verification

`python -m pytest tests/ -q` — 1011 pass. `node --test workflows/test/*.test.js` — 585 pass. `pre-commit run --all-files` — all hooks pass.

Gate decision for #46's requirement 1 recorded on the issue thread.

https://claude.ai/code/session_01Pnva1Qsmpqon81QyBj7Wxf

---

> [!NOTE]
> **Low Risk**
> Mostly documentation, tests, and rule-collection behavior; collector dedup changes what counts toward byte/file caps but is covered by new tests.
> 
> **Overview**
> **Restructures agent instructions** so portable rules live in root `AGENTS.md`, directory rules in `workflows/`, `scripts/`, and `agents/` `AGENTS.md`, and root `CLAUDE.md` is only `@AGENTS.md` plus a small Claude-only section. Adds `scripts/sync_agent_rules.py` to generate subdirectory `CLAUDE.md` twins (on-demand loading does not expand `@imports`).
> 
> **Replaces** `tests/test_claude_md_budget.py` with `tests/test_agent_instruction_layout.py` and wires pre-commit to it—guarding Codex’s 32 KiB cap, bare root `@AGENTS.md`, twin freshness, and that instruction files reference real paths/symbols. Adds **`REVIEW.md`** for this repo’s review config and updates **`CONTRIBUTING.md`** to point at `AGENTS.md` and the sync script.
> 
> **`collect_project_rules.py`** fingerprints rule text (HTML comments stripped for comparison) to skip duplicate `CLAUDE.md`/`AGENTS.md` pairs, applies the total byte budget after dedup, and counts files **walked** toward `--max-files`. **`tests/test_dimensions_registry.py`** locks schema prose to `agents/AGENTS.md`; **`tests/test_agent_contracts.py`** blocks v2 NDJSON residue in `agents/` directory rules and asserts the by-value findings contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 306b5fec3fccfafb56b4e089bcb13f402d1a3eec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:
Fixed in `db03a2d` — thanks, this one was worth more than it looked.

**The finding is correct.** The apostrophe rule did change from the escape sequence to a bare `'` when the emission guidance moved into `agents/AGENTS.md`. In a rule telling agents to build a single-quoted `printf` command, a bare apostrophe terminates the shell quote and corrupts the line — so the rule was self-defeating.

**But the fix isn't restoring the escape.** Checking why the guidance was there at all:

```
bug-detector             tools: Read, Grep, Glob, LSP
security-reviewer        tools: Read, Grep, Glob, LSP
test-analyzer            tools: Read, Grep, Glob, LSP
cross-file-impact        tools: Read, Grep, Glob, LSP
conventions-and-intent   tools: Read, Grep, Glob, LSP
type-design-analyzer     tools: Read, Grep, Glob, LSP
code-simplifier          tools: Read, Grep, Glob, LSP
```

No discovery agent has Bash. They return findings **by value** as the task's structured result. There is no shell emission path in v3 — the entire NDJSON/printf contract was v2 residue (tracked as #65), and it had landed in a file that loads alongside every contract in that directory, re-teaching exactly what the v3 scrub had removed from the seven agent bodies.

A broken shell-quoting rule for a shell command no agent can run is the tell that the section itself was stale. So it is deleted rather than corrected: `## Emitting findings` → `## Returning findings`, stating the by-value contract.

Two guards added, both mutation-verified (each fails when the mechanism is reverted): the existing emission-residue pattern now also runs against `agents/AGENTS.md` and its generated twin, and a positive test requires the by-value statement to be present — deleting stale guidance without stating the real rule just leaves a gap.

leehopper:
Swept for the same defect class — *content moved without re-verifying it against this tree* — and found three more, all documenting code that exists only on unmerged #87:

| claim | reality |
|---|---|
| `INPUT_FAULT_REASONS` named as a cross-runtime list | not on main |
| the proof-kind rule (pre-serialized string vs object) | describes `input_content_proof`/`sliceInputProofFor` |
| the whole *"Health is derived, never asked for"* section | describes `reviewHealth`/`healthBanner` |

None of these was in main's CLAUDE.md — I wrote them from context carried out of the PR-87 branch earlier. On main they send an agent looking for functions that do not exist, and nothing reports a dangling reference. Removed here; they should land **with #87**, in the AGENTS.md files, when that branch rebases onto this layout.

**Structural fix, since prose review plainly does not catch this.** Every backticked path and identifier in the six instruction files must now resolve in this tree — paths against `git ls-files`, symbols against `git grep`, ignoring hits in the instruction files themselves (the twins are copies, so a symbol could otherwise vouch for itself).

That self-vouching turned out not to be hypothetical: the first version of the guard **did not fail its own mutation test**, because the test's docstring named the missing constant and `git grep` found it there. Fixed by rewording the docstring rather than adding an exclusion list — an exclusion would have let any future missing symbol hide in that file too.

Both halves mutation-verified: re-adding the unmerged constant fails the symbol check; naming a nonexistent script fails the path check.

leehopper:
Fixed in `306b5fe` — good catch, and it matters more under this PR than it would have before.

The total-byte budget was enforced in `_read`, off a bare `stat`, **before** `visit` computed the content fingerprint. So a twin that dedup would have made free could still trip the cap, set `truncated`, and emit `project_rules_truncated` — claiming rules were dropped while that exact content already sat in `sources`.

That is a **fabricated gap**, which this repo treats as exactly as wrong as a fabricated success — and it is worse than the duplication the dedup replaced, because the run now misreports its own completeness. This PR's twin layout turns it from a corner case into the common one.

The budget now runs in `visit`, after dedup, against the bytes actually being included. The per-file cap still runs on `stat` before any `open`, so an oversized file is never read — `test_over_cap_file_is_never_opened` pins that and still passes.

Two tests, mutation-verified as a pair: a free duplicate must not trip the cap, **and** genuinely distinct content of the same size still must. The second one exists because relocating a check is precisely the edit that quietly disables it — reverting the reorder fails the first and passes the second.

PR #119, opened 2026-07-31
docs(audit): v2→v3 residue inventory, string registry, and CI guards

### LINKED ISSUE #37: Systematic audit for v2→v3 residue: orphaned references, dead code paths, doc/code contract skew, and diverged duplicates
## Problem / Goal

The v2→v3 conversion moved orchestration out of prose and into the JS pipeline, but the sweep for what that left behind was never done systematically. Every recent investigation that looked at an unrelated area found residue as a side effect — which is strong evidence the population is larger than the handful now known, and that the ones found so far were found by luck rather than by method.

Confirmed instances, each discovered incidentally: a formatting template still read by the orchestrator but never delivered to the component that replaced it as renderer; two parser functions shipped, parity-pinned and tested, but never called on the live path; an instruction to render a surface that no longer exists in v3; a cross-file pointer stale for one half of what it points at; one decision documented three divergent ways across three files; and a machine-written marker whose payload shape disagrees with the two documents that specify it — in a way that silently breaks the incremental-review feature reading it.

Goal: replace incidental discovery with one systematic pass. Produce a complete inventory of residue across the plugin, fix what is safe, file what is not, and leave behind a cheap regression guard so the most common class cannot silently recur.

## Requirements

1. A defined, reproducible detection method covering at minimum these classes, applied across `skills/`, `agents/`, `scripts/`, `workflows/src/`, and `.github/`:
   - **Orphaned references** — reference docs, templates, or sections that no live component reads; content relocated without updating its pointers.
   - **Dead code paths** — exported functions, branches, or config values never reachable on the live path; dormant plumbing left from a removed feature.
   - **Doc/code contract skew** — documentation specifying a shape, field, or value that the code does not produce (or vice versa), especially where another component parses it.
   - **Stale version and roadmap pointers** — retired phase/stage names, superseded version placeholders, references to closed or superseded issues.
   - **Diverged duplicates** — content deliberately duplicated but no longer identical, and duplication lacking the repo's `<!-- Canonical source: … -->` marker convention.
2. The output is a written inventory artifact (in-repo or on the issue thread) listing every instance with location, class, evidence, and a disposition: fix-in-place, file-as-issue, or intentional-and-documented.
3. Triage discipline: mechanical, provably-safe corrections (stale pointers, retired names, orphaned prose) are fixed in this work; anything requiring a design decision or touching pipeline behavior becomes its own issue rather than expanding this one.
4. **No re-litigation of owned scope.** These specific instances already belong to filed issues and must be recorded in the inventory as owned-elsewhere, not fixed here: the report-formatting contract that never reaches the renderer and the comment-marker payload skew (#36); the shipped-but-uncalled `parseReviewMd`/`loadExclusions` and the prompt-only limits defaults (#24); the three-way "no REVIEW.md" divergence and the stale prompts pointer (#35); the v2-era issue-form phase names and version placeholders (#30); the `#17` roadmap pointers (#29). The audit's value is the instances *not* on this list.
5. Intentional duplication and retained-but-unused code are distinguished from residue and explicitly documented as intentional, not deleted: the false-positive exclusion list duplicated across the 7 discovery agents, the retained Python twins and their golden parity fixtures, and the v2 alias fields at the artifact-writer persist boundary are all deliberate per CLAUDE.md.
6. **A machine-parsed string registry.** Every string another component parses — the `Headless config:` knob lines and their byte-identical three-copy emission, the `code-gauntlet-findings` marker, the `Generated by code-gauntlet` / `Reviewed up to:` prior-review substrings, `HEADLESS CONFIG ERROR:` / `HEADLESS INPUT ERROR:`, the verbatim Workflow-absent message, `[validate_ndjson]` prefixes, and the verify executor receipt — is enumerated in one authoritative place with its parser named. Three separate pieces of recent work each had to re-derive this list before they could safely change output; it should exist once.
7. **A regression guard for the cheapest class.** At minimum, an automated check that every file under `skills/code-gauntlet/references/` is reachable from `SKILL.md`, an agent file, or another reference — so a document cannot become orphaned again without CI noticing. Broader guards (dead-export detection, doc/code contract assertions) are assessed and either implemented or recorded as deliberately skipped with a reason.
8. Language-agnostic and stdlib-only per repo rules; any new check runs in the existing suites.

## Evidence

Confirmed residue found incidentally, each by an investigation aimed at something else:

```text
Orphaned from its consumer — references/report-format.md holds the full report
  convention (emoji headings, footer) and is read by the orchestrator
  (SKILL.md:262, phase8-delivery.md:35, :60), but the agent that actually
  renders the report in v3 (agents/report-writer.md, dispatched from
  workflows/src/stages.js:1168-1178) never receives it.   [owned by #36]

Instruction to render a surface that no longer exists — phase8-delivery.md:60
  tells the orchestrator to output the full report per report-format.md; in v3
  the report-writer agent produces it.                      [owned by #36]

Shipped, tested, never called — workflows/src/filterFindings.js:71
  parseReviewMd(text) and :363 loadExclusions(text) are parity-pinned and
  dead on the live path while SKILL.md:121-135 hand-parses the same file.
                                                            [owned by #24]

Pointer stale for half its target — phase2-triage.md:185 points at
  review-md-spec.md "for the full prompts and scaffolding templates"; the
  prompts half was replaced in phase2-triage itself.         [owned by #35]

One decision, three divergent treatments — phase2-triage.md:164 (plain text),
  review-md-spec.md:263 (full question), phase1-preflight.md:189 (a third
  option set).                                               [owned by #35]

Doc/code contract skew with a functional consequence — post_review.py:347-353
  writes {"version":"3.0","findings_count":N,"sha":...} while
  report-format.md:240 and delivery-guide.md:129 document
  {"version":1, ..., "findings":[...]}; phase1-preflight.md:100 parses this
  marker for incremental review, so the docs promise an array the code never
  writes.                                                    [owned by #36]

Retired names in the public contribution surface — bug_report.yml lists v2
  phases ("Phase 3 — Review agents", "Phase 4 — Classify & verify") against a
  v3 architecture, plus a 2.3.4 version placeholder.         [owned by #30]
```

Each was found while investigating an unrelated question — none by a search for residue. The classes above are therefore lower bounds, not an inventory.

Deliberate duplication that must NOT be swept (per CLAUDE.md): the false-positive exclusion list across the 7 discovery agents (failure isolation — every agent keeps the list even if a file read fails), the retained Python twins with frozen golden fixtures, and the union-schema v2 aliases at the persist boundary.

- Reachability sweep: for each file in `references/`, grep the plugin for its filename; anything with no inbound reference from a live path is a candidate.
- Dead-export sweep: for each `export function` in `workflows/src/`, check for a call site outside tests and the bundle.
- Contract sweep: for each machine-parsed string in the requirement-6 registry, diff the producer against every document that specifies it.

The pickup session owns the method and the design (spec-driven, TDD).

## Out of scope

- Any change to review-pipeline semantics — this is discovery plus mechanical correction.
- The specific instances enumerated as owned-elsewhere in requirement 4.
- Deleting retained Python twins, parity fixtures, or the deliberate duplications in requirement 5.

## Verification

Tier: **suite-only** for the audit artifact, the registry, the regression guard, and documentation-only corrections — the always-on suites (pytest + `node --test`, parity and bundle-freshness included) plus any new reachability check are sufficient, and CI must stay green.

Escalates to **smoke** if the pass removes or rewires any live code path (e.g. deleting a dead export that turns out to be reachable, or touching a machine-parsed string): those ride the release manager's per-sub-release functional smoke — mechanical pass/fail, asserting nothing about recall or noise. No paired benchmark under any outcome; if the audit uncovers something that would plausibly move recall or noise, it is filed as its own issue with its own tier rather than measured here.

Done means: the inventory artifact exists and every entry carries a disposition; the machine-parsed string registry exists with parsers named; the references-reachability check runs in CI and passes; and every deferred instance has a filed issue.

leehopper:
## Summary
- Point-in-time residue inventory (`docs/v3-residue-audit-2026-07.md`) with dispositions
- Living machine-parsed string registry + presence CI
- References reachability + workflows dead-export CI guards
- Mechanical fixes in-PR; design-touching items filed as follow-ups

## Test plan
- [ ] `python -m pytest tests/ -q`
- [ ] `python -m pytest bench/tests/ -q`
- [ ] `node --test workflows/test/*.test.js`
- [ ] `node workflows/build.js && git diff --exit-code workflows/pipeline.js`
- [ ] Confirm inventory every row has a disposition and deferred rows link issues

Closes #37


tests/test_machine_parsed_strings.py
```diff
@@ -0,0 +1,112 @@
...
+        string, producers, parsers, notes = (c.strip() for c in m.groups())
+        if string == "string" or set(string) <= {"-", " "}:
+            continue
+        rows.append({
+            "string": string,
+            "producers": [_strip_ticks(p) for p in producers.split(",") if p.strip()],
+            "parsers": [_strip_ticks(p) for p in parsers.split(",") if p.strip()],
+            "notes": notes,
+        })
+    return rows
```
leehopper, line 52: **💡 [LOW] parse_registry/_ROW table parser has no fixture-based unit test, unlike the sibling reachability guard**

parse_registry() and the _ROW regex are the sole logic that turns docs/machine-parsed-strings.md rows into producer/parser lists that every other test in the file asserts against. All three test methods call parse_registry only on the real registry's current content, so a bug in the row regex or column splitting (e.g. mishandling commas inside a cell, or the NO_PARSER sentinel) would only be caught if it happened to also break on today's exact table shape.

**Suggested fix:**
Add a unit test that feeds parse_registry() a small synthetic markdown table (a normal row, a NO_PARSER `—` row with and without a note, a row with multiple comma-separated producers) and asserts the exact parsed structure, independent of the live registry file's current contents.

tests/test_workflows_dead_exports.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def _opens_regex(code_tail: str) -> bool:
+    """True when a `/` following *code_tail* (recent non-space code) starts a regex."""
+    if not code_tail:
+        return True
+    if code_tail[-1] in _REGEX_PRECEDERS:
+        return True
+    word = _TRAILING_WORD.search(code_tail)
+    return bool(word) and word.group(1) in _REGEX_KEYWORDS
```
leehopper, line 52: **💡 [LOW] Regex-vs-division keyword branch of _opens_regex is never exercised**

The new dead-export scrubber's _opens_regex helper has two ways a leading `/` is classified as a regex literal: following a punctuation preceder, or following one of the _REGEX_KEYWORDS (return, typeof, case, new, etc). Only the preceder path is exercised by test_scrubber_blanks_comments_and_literals_but_keeps_code; the keyword path has no test case and, per a grep of workflows/src, is not even indirectly exercised by the guard's real-file run today.

**Suggested fix:**
Add a scrubber test line such as `return /ghostName/.test(x);` or `case /ghostName/:` to source and assert the token inside the regex is blanked, proving the keyword-preceded branch of _opens_regex correctly opens a regex rather than misreading `/` as division.

skills/code-gauntlet/references/delivery-guide.md
```diff
@@ -10,7 +10,7 @@ Implementation details for each delivery method in Phase 8.
...
 
-**Inline comment default cap: 6.** When the user selects "Default — top 6 by severity," post the top 6 findings as inline comments (highest severity first); remaining findings go in condensed format in the executive summary comment. When the user selects "Let me pick" and chooses findings via the interactive walkthrough, post **all selected findings** as inline comments — no cap applies. The user made a deliberate selection; respect it.
+**The inline comment set is the pipeline's, not yours.** When the user selects "Default — the pipeline's selected set," post the `artifactPaths.postReview` payload verbatim — `selectDelivery` already applied the delivery tier, ranked the survivors, and capped them at `limits.deliveryCap`. Do not re-rank, re-filter, or re-apply the cap. When the user selects "Let me pick" and chooses findings via the interactive walkthrough, post **the selected subset** as inline comments — deselection only, order preserved, no cap applies. The user made a deliberate selection; respect it.
```
leehopper, line 13: **💡 [LOW] New 'deselection only' claim conflicts with phase8-delivery.md's 'full findings list' walkthrough description**

The PR's corrected text asserts the "Let me pick" walkthrough is deselection-only from the pipeline's postReview set, but the sibling (untouched) phase8-delivery.md:88 says the walkthrough runs "over the full findings list. Includes Improvement Suggestions" — a superset of postReview per phase8-delivery.md:24-25, so hand-selecting from it is not pure deselection from postReview.

**Suggested fix:**
Either reconcile phase8-delivery.md:88 to state the walkthrough operates strictly over the postReview subset (matching the new 'deselection only' claim and Step B.1's 'strict subset of the wrapper's entries'), or soften delivery-guide.md's new wording to acknowledge the walkthrough can select from the full findings list, not just deselect from postReview.

leehopper (commented):
Automated code-gauntlet review (v3 pipeline). 3 low-severity suggestions found and verified via blind-challenge; no functional defects in the fix-in-place corrections themselves.

---
Generated by code-gauntlet | Reviewed up to: bd14e3d7fdb1d8f845b0af54f329cb843176440e

docs/machine-parsed-strings.md
```diff
@@ -0,0 +1,30 @@
...
+| `Reviewed up to:` | `scripts/review_marker.py`, `skills/code-gauntlet/references/delivery-guide.md`, `skills/code-gauntlet/references/report-format.md` | `—` | Prose footer half. The reader is `_PROSE_LABEL_RE`, a whitespace-tolerant regex (`Reviewed\s+up\s+to\s*:`), so no parser holds these bytes; the row pins the three producing copies against each other |
+| `HEADLESS CONFIG ERROR:` | `skills/code-gauntlet/references/headless-mode.md` | `—` | Fail-loud template; the orchestrator must print this exact prefix and a human reads it. No component parses it — the previous self-listing as its own parser proved nothing |
+| `HEADLESS INPUT ERROR:` | `skills/code-gauntlet/SKILL.md`, `skills/code-gauntlet/references/phase2-triage.md` | `skills/code-gauntlet/references/headless-mode.md` | Prefix of working-tree≠PR-head abort |
```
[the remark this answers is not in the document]
leehopper: Fixed in the follow-up commit: `headless-mode.md` is now a third **producer** with `—` in the parser column (same fail-loud/human-read shape as `HEADLESS CONFIG ERROR:`), not a phantom parser.

tests/test_machine_parsed_strings.py
leehopper: Fixed: added `test_parse_registry_handles_fixture_shapes` covering multi-producer rows, `—` with/without notes, and header-row ignoring.

tests/test_workflows_dead_exports.py
leehopper: Fixed: extended `test_scrubber_blanks_comments_and_literals_but_keeps_code` with `return /ghostName/`, `case /ghostName/:`, and `typeof /ghostName/` so the keyword branch of `_opens_regex` is exercised.

skills/code-gauntlet/references/delivery-guide.md
leehopper: Fixed by reconciling `phase8-delivery.md` to the postReview deselection contract (matching `delivery-guide.md`, Step B.1, and the SKILL.md gate): the walkthrough now runs over `artifactPaths.postReview`, not the uncapped findings file. Left `delivery-guide.md`'s wording as-is.

leehopper:
Measurement tier ruling (recorded on #101, 2026-07-31): **suites-only — no smoke needed for this PR.** #37's Verification section escalates to smoke only when a live code path is removed/rewired or a machine-parsed string value changes; this diff's changes to live files are comment/docstring/doc-prose only (verified per file), the pipeline.js delta is the mechanical rebuild of those comment edits, and the three new test files are guards, not path changes. Green suites are the whole gate.

PR #126, opened 2026-08-03
ci: gate JS coverage with floors, allowlist, and presence (#103)

### LINKED ISSUE #103: Gate the JS coverage measurement in CI with floor thresholds and an explicit include allowlist

## Problem / Goal

`node --test --experimental-test-coverage workflows/test/*.test.js` already produces a full coverage report on demand — 98.98% line / 83.16% branch / 98.37% function across `workflows/build.js` and eight `workflows/src/*.js` modules — but nothing in CI runs it and nothing enforces it. The `workflow-tests` job runs the suite bare, so coverage can regress by any amount without a check going red. Separately, the generated bundle `workflows/pipeline.js` is absent from the report only by accident: no test happens to import it (`build.test.js` constructs the bundle in memory and never reads the file from disk). Nothing states that exclusion, and a future test that imports the bundle would silently double-count every line of the source it was built from.

Leaving it costs the repo the enforcement that is cheapest to have: the suite runs in 0.33s and instrumentation adds no meaningful time, yet today a PR can delete a fixture group or land an untested module with no signal. Fixed looks like: the same single CI step runs the suite with coverage on, fails the job below stated floors, measures a set of files named by an allowlist rather than inferred from whatever the tests happened to import, and runs on a pinned Node patch version so the percentages cannot move without a deliberate commit. The one uncovered range that is real logic rather than a defensive guard is closed at the same time.

## Requirements

1. The `workflow-tests` job in `.github/workflows/ci.yml` runs the node test suite with coverage measurement enabled and fails the job when aggregate coverage falls below the configured floors — verifiable by temporarily raising a floor above baseline and observing a non-zero exit.
2. The configured floors are 98% line, 82% branch, 97% function, and the gated command exits 0 at HEAD with no source or test changes (baseline 98.98 / 83.16 / 98.37).
3. The set of files measured is stated by an allowlist in the command itself, not inferred from which modules the tests happen to load. `workflows/pipeline.js` is absent from the report because it is not on the allowlist, and remains absent even if a test imports it.
4. The gated report covers exactly the nine product rows (`workflows/build.js` plus the eight loaded `workflows/src/*.js` modules) and contains no row under `workflows/test/` — verifiable by reading the report emitted in the CI log.
5. The Node version used by the `workflow-tests` job is pinned to an exact patch version matching the runtime the bundle ships into, so a Node minor or patch bump cannot move the measured percentages without a commit that changes the pin.
6. The coverage command string documented in `AGENTS.md` and the one executed in `.github/workflows/ci.yml` are identical, and a deterministic test in `tests/` fails if they diverge.
7. `routeByDimension`'s `type_design` arm is exercised in both directions — keyword-promoted to `main` and not-promoted to `suggestion` — through the parity fixture set, so the assertion binds the Python twin as well as the JS port. `workflows/src/filterFindings.js` lines 709-711 no longer appear as uncovered.
8. A deterministic test fails if a git-tracked `workflows/src/*.js` module other than the unimportable bundle entry (`pipeline_entry.js`) is absent from the coverage report, so a newly added module with no tests cannot pass by simply never being loaded.
9. The chosen floors, and the reason each sits roughly one point below baseline (including the measured run-to-run branch jitter), are recorded adjacent to where the thresholds are set, together with a stated rule for when the floors get raised.
10. The job name and step name in `ci.yml` are unchanged, so the required status-check context `Run Workflow JS Tests` still matches and no branch-protection edit is needed.
11. `python -m pytest tests/ -q`, `python -m pytest bench/tests/ -q`, `node --test workflows/test/*.test.js`, `node workflows/build.js && git diff --exit-code workflows/pipeline.js`, and `pre-commit run --all-files` all still pass.

## Evidence

All figures 2026-07-30, HEAD `ebf399d` (tag v3.3.4), Node v24.18.0.

```text
$ node --test --experimental-test-coverage workflows/test/*.test.js
596/596 pass, duration_ms 289-326, wall-clock 0.42s real
aggregate: 98.98% line / 83.16% branch / 98.37% function over nine rows
```

- 2026-07-30 per-file baseline: `build.js` 86.49/92.00/90.91 (uncovered 91-97, 104-111); `applyChallenges.js` 100/84.21/100; `applyValidations.js` 97.30/82.61/100 (27-28); `args.js` 99.83/97.66/100 (260); `filterFindings.js` 98.56/84.19/97.44 (334-335, 430-435, 709-711, 713-714, 772); `findingDedup.js` 100/86.67/100; `mergeFindings.js` 97.19/89.26/91.67 (52-54, 60-64, 132-133, 175, 244); `registry.js` 100/100/100; `stages.js` 99.55/78.51/99.34 (22-28, 389-393, 1606-1607).
- 2026-07-30 determinism sample, 30 runs of the identical command on the unchanged tree: aggregate branch 83.16 in 25 runs, 83.00 in 5 (`stages.js` 78.51 vs 78.24). Line 98.98 and function 98.37 in all 30. The 83.00 outcome occurred under both default concurrency and `--test-concurrency=1`, and both with and without CPU contention. A baseline-tight branch threshold would therefore fail roughly one CI run in six on green code.
- 2026-07-30, gate behaves:

```text
$ node --test --experimental-test-coverage \
    --test-coverage-include='workflows/src/*.js' --test-coverage-include='workflows/build.js' \
    --test-coverage-lines=98 --test-coverage-branches=82 --test-coverage-functions=97 \
    workflows/test/*.test.js
exit 0, report unchanged at 98.98/83.16/98.37

$ ... same command with --test-coverage-branches=90
exit 1: Error: 83.16% branch coverage does not meet threshold of 90%.
```

- 2026-07-30 exclude-flag footgun: `node --test --experimental-test-coverage --test-coverage-include='workflows/**' --test-coverage-exclude='**/node_modules/**' workflows/test/*.test.js` reports **34 rows** including all 16 test files and both helpers (`pipelineMock.js` 96.41/75.71/100, `verifyDelta.js` 100/94.12/100), aggregate **99.43 / 87.76 / 97.45** — versus 9 rows and 98.98 / 83.16 / 98.37 with the include allowlist alone. Supplying *any* `--test-coverage-exclude` overrides Node's built-in default suppression of test files.
- 2026-07-30 bundle-exclusion proof: `grep -rn "pipeline.js" workflows/test/` returns no matches. `workflows/test/build.test.js` imports only `{ detectTopLevelCollisions, build }` from `../build.js` and runs the collision detector over an in-memory bundle string; it never reads or imports the committed `workflows/pipeline.js`.
- 2026-07-30 allowlist limitation proof: `workflows/src/pipeline_entry.js` matches the `workflows/src/*.js` include glob but does not appear in the coverage report, because no test loads it (`build.js` concatenates it as text; the file ends in a top-level `return await run(__args)`, a syntax error in an ES module). An include allowlist therefore cannot detect an added-but-untested module on its own.
- 2026-07-30, current CI shape (`.github/workflows/ci.yml` at HEAD): the `workflow-tests` job uses `node-version: "24"` (floating major, line 76) and runs `node --test workflows/test/*.test.js` (line 79), then `node workflows/build.js` plus `git diff --exit-code workflows/pipeline.js`. `ci.yml` is the only workflow that runs any of the four deterministic suites.

- Amend the existing `Run node test suite` step (`ci.yml` line 79) rather than adding a step or job — keeping the job and step names unchanged is what satisfies requirement 10. Suggested command:

  ```text
  node --test --experimental-test-coverage \
    --test-coverage-include='workflows/src/*.js' \
    --test-coverage-include='workflows/build.js' \
    --test-coverage-lines=98 \
    --test-coverage-branches=82 \
    --test-coverage-functions=97 \
    workflows/test/*.test.js
  ```

- Use `--test-coverage-include` (repeatable allowlist), never `--test-coverage-exclude` — see the measured exclude-flag footgun in Evidence.
- Pin `ci.yml` line 76 from `node-version: "24"` to `node-version: "24.18.0"`, with a comment tying it to the sandboxed runtime the bundle ships into. One-line change, and it is what stops an unrelated Node patch bump from moving the percentages.
- Close the `type_design` gap with fixtures, not new test code: add `tests/fixtures/parity/filter_findings/dimension_routing/type_design_promoted_by_keyword/` and `.../type_design_not_promoted/` mirroring the existing `conditional_promoted_by_keyword` / `conditional_not_promoted` pair with `"dimension": "type_design"`, then run `python3 workflows/test/tools/record_parity.py filter_findings` to generate `expected.json`. `workflows/test/parity.test.js` and `tests/test_parity_fixtures.py` both discover cases by directory walk, so no test file changes and the Python twin gets asserted too.
- For the command-parity and module-presence assertions, extend `tests/test_contribution_surface.py` — it already owns the pattern of asserting that commands quoted in prose match the ones CI runs, so this is one more case in an existing mechanism rather than a new file.
- When writing the module-presence check: an include allowlist does not surface a listed-but-never-loaded file. Compare the report against `git ls-files 'workflows/src/*.js'` minus `pipeline_entry.js`; do not rely on the allowlist to do it.
- Per the repo's regression-test rule, verify each new assertion fails against the condition it names — revert the `ci.yml` command to the bare form and confirm the parity test goes red; delete the `type_design` fixtures and confirm the coverage range reappears.
- Note that the floors are deliberately floors, not baselines: branch coverage here is dominated by defensive paths that will not climb on their own, so the threshold exists to catch a cliff (a deleted fixture group), not to chase 100%.

## Out of scope

- Python coverage for `scripts/`, `tests/`, or `bench/` — #102 owns it, including any `pyproject.toml` this repo grows.
- Per-file coverage thresholds. Node's flags are global-only and `build.js` at 86.49% line would fail any per-file line floor the aggregate passes; per-file gating would require a third-party tool.
- Introducing c8, nyc, istanbul, `package.json`, or `node_modules` anywhere in the repo tree.
- Codecov/Coveralls uploads, coverage badges, or any external reporting service.
- Adding a new CI job or renaming an existing one — that would require a branch-protection edit (#108) and, once #108 lands, an update to the frozen check-name list.
- Writing tests for the remaining uncovered defensive branches: `stages.js` 22-28 (sandbox-global probe), 389-393 and 1606-1607 (degrade-and-record-gap paths), `applyValidations.js` 27-28 and `filterFindings.js` 430-435 (Python-`int()` parity fallthroughs), `filterFindings.js` 334-335, 713-714, 772, `mergeFindings.js` 52-54/60-64/132-133/175/244, `args.js` 260, `build.js` 91-97/104-111.
- Coverage for `workflows/src/pipeline_entry.js` — it ends in a top-level `return`, valid only inside the workflow host's wrapper and a syntax error as an ES module, so it is structurally unimportable and cannot be instrumented.
- Root-causing the ~0.16-point run-to-run branch jitter in `stages.js`. This issue absorbs it with threshold headroom; diagnosing it is a separate determinism investigation.
- JS linting or formatting — #105 owns that, and any consolidation of `mergeFindings.js` internals belongs to #110.
- Changing the review pipeline's behavior or output in any way.

## Verification

`node --test workflows/test/*.test.js` (bare and gated), `node workflows/build.js && git diff --exit-code workflows/pipeline.js`, `python -m pytest tests/ -q`, `python -m pytest bench/tests/ -q`, and `pre-commit run --all-files` all pass; the gate is proven live by raising a floor above baseline and observing a non-zero exit, and the two new deterministic assertions are each proven by mutation (revert the `ci.yml` command; delete the `type_design` fixtures). Measurement tier: `suites-only`.

leehopper:
## Summary
- Gate `workflow-tests` on Node coverage floors 98/82.3/97 with an include allowlist
- Presence check via lcov + `coverage_scope.json` (`pipeline_entry` exempt from presence only); empty scope hard-fails
- Byte-identical AGENTS↔ci.yml drift test for all three coverage gates; Python legs use `mktemp` spelling
- `type_design` parity fixtures close filterFindings routing arm

## Test plan
- [x] req-11 suites green locally
- [x] floor raise → non-zero exit
- [x] mutations: drift, include sync, presence, type_design fixtures
- [x] **Re-pin JS floors from first green CI run** — Node 24.18.0 measured 98.61 / 83.27 / 97.99; raised branch floor 82 → 82.3 (1.27 pp was outside the 1.0 pp policy; line/function floors stay 98/97)

Closes #103

---

> [!NOTE]
> **Low Risk**
> Changes are CI, test infrastructure, and documentation parity guards; no production pipeline runtime behavior changes beyond new golden fixtures for existing routing rules.
> 
> **Overview**
> **CI now enforces JavaScript coverage** on workflow tests with line/branch/function floors (98 / 82.3 / 97), an include allowlist for `workflows/src/*.js` and `workflows/build.js`, and a post-test **presence check** that every in-scope tracked file appears in the LCOV output (`pipeline_entry.js` exempt from presence only). Node for that job is pinned to **`24.18.0`** so V8 coverage percentages stay stable.
> 
> **Python coverage gate steps** spell `COVERAGE_FILE="$(mktemp -d)/.coverage"` inline in the `run` block instead of using a separate `env:` block. **AGENTS.md** documents all three gates (two Python + one JS) as self-contained commands that must match **ci.yml** byte-for-byte; **CONTRIBUTING.md** documents the exact Node pin and points to that block for the full JS gate.
> 
> **Structural drift guards:** `tests/test_contribution_surface.py` asserts AGENTS ↔ CI gate command identity, JS includes ↔ `coverage_scope.json`, and CONTRIBUTING mentions the pinned Node version. New **`check_coverage_presence.mjs`**, **`coverage_scope.json`**, and **`coverage_presence.test.js`** implement and test the presence tool. **`AGENTS_SET_BUDGET_BYTES`** is raised for the added coverage documentation.
> 
> **Parity fixtures** add two `filter_findings` cases for `type_design` dimension routing (style → suggestion vs type-safety keywords → main).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e86802d7f30a4cce03179422868acdfcd9e66141. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper (commented):
Design-conformant against the locked §1–§4 — byte-identity held in CI, presence/drift/sync guards all landed as specified, fixtures target the right keyword branch. One required change (the re-pin policy fires on the branch floor: CI measured 83.27 vs floor 82), one recommended guard (Node-pin has no test), two minors. Details inline. Everything else is proven by the green run.

.github/workflows/ci.yml
```diff
@@ -90,10 +87,22 @@ jobs:
...
+          node-version: "24.18.0"
 
       - name: Run node test suite
-        run: node --test workflows/test/*.test.js
+        run: |
+          LCOV="$(mktemp -d)/js-coverage.lcov" && node --test --experimental-test-coverage \
+            --test-coverage-include='workflows/src/*.js' \
+            --test-coverage-include='workflows/build.js' \
+            --test-coverage-lines=98 \
+            --test-coverage-branches=82 \
```
leehopper: **Required by the pin policy — branch floor.** First green CI measured **98.61 / 83.27 / 97.99** (job 91786808217). Lines (0.61 pp) and functions (0.99 pp) are within the 1.0 pp policy; **branches sit 1.27 pp below measured** — bump to `--test-coverage-branches=82.3` here and in the AGENTS.md copy (the drift test will enforce the pair). Same treatment as #125's scripts-floor re-pin.

AGENTS.md
```diff
@@ -35,23 +35,41 @@ python -m pytest bench/tests/ -q      # benchmark harness self-tests
...
 ```
 
-Floors 91.3 / 87, pinned 2026-08-03 from the first green 3.12 CI run (92.27 /
-87.96); policy: a floor sits no more than 1.0 pp below the CI 3.12 measurement.
-`workflows/test/tools/record_parity.py` is test infrastructure, outside both
-scopes. Lower a floor only in the PR that causes the drop, reason in the body.
-A sudden multi-point drop means broken subprocess capture — fix capture, do not
-lower.
+Floors: Python 91.3 / 87 (pinned 2026-08-03 from first green 3.12 CI: 92.27 /
+87.96); JS 98 / 82 / 97 (provisional from local HEAD ~98.56 / 83.00 / 97.99 —
```
leehopper: When re-pinning the branch floor to 82.3, rewrite this sentence past-tense to the CI basis — "JS 98 / 82.3 / 97, pinned 2026-08-03 from first green CI: 98.61 / 83.27 / 97.99" — and drop the provisional/local-HEAD clause so it doesn't read as permanently pending (the #125 lesson).

Also in this paragraph: "raise when measured headroom **comfortably** exceeds 1.0 pp" — "comfortably" is the one unquantified word in an otherwise numeric policy. Give it a number or cut the clause.

CONTRIBUTING.md
```diff
@@ -95,9 +95,11 @@ The review pipeline runs inside Claude Code's workflow runtime, so it carries co
   `setTimeout`/`queueMicrotask`, `process`, and others — are absent from the workflow runtime sandbox, so a
   reference keeps every local test green and then throws on the first live dispatch. `workflows/AGENTS.md` holds the
   normative list; deep-clone with the bundle's `deepClone` helper, never `structuredClone`.
-- **CI pins the Node major, not the patch.** `.github/workflows/ci.yml` runs the workflow tests and the build on
-  `node-version: "24"`; 24.18.0 is the reference build this repo develops against. There is no `package.json` and
-  no `node_modules`; use Node built-ins only.
+- **CI pins Node `24.18.0` exactly.** Coverage floors are sensitive to the V8 build; the patch pin stops an
```
leehopper, line 98: **Guard gap:** nothing asserts this `24.18.0` matches `node-version:` in ci.yml — no test references node-version at all. Given this bullet's own rationale (floors are V8-sensitive), a Node bump touching one file but not the other is the expected failure mode. One assertion in `test_contribution_surface.py` — extract `node-version:` from ci.yml, assert the same string appears here — closes it and is squarely this PR's guard family.

tests/test_contribution_surface.py
```diff
@@ -802,5 +802,82 @@ def test_docstring_names_the_ruleset(self):
...
+            indent = len(line) - len(line.lstrip(" "))
+            if content_indent is None:
+                if indent <= base:
+                    break
+                content_indent = indent
+            if indent < content_indent and line.strip():
+                break
+            collected.append(line[content_indent:] if content_indent is not None else line)
+            i += 1
+        blocks.append(_strip_one_trailing_newline("\n".join(collected)))
```
leehopper: Minor robustness: this strips only **one** trailing newline, but a second cosmetic blank line after a gate command in ci.yml would leave a residual `\n` and fail the drift test with a confusing byte-diff. `"\n".join(collected).rstrip("\n")` fails red on every real difference just the same — it only makes the red honest. (Interior blank lines are unaffected.)

workflows/test/tools/check_coverage_presence.mjs
```diff
@@ -0,0 +1,92 @@
...
+      if (p.indexOf('*') !== p.lastIndexOf('*')) return false; // multi-star: refuse
+      if (!f.startsWith(pre) || !f.endsWith(post)) return false;
+      continue;
+    }
+    if (p !== f) return false;
+  }
+  return true;
+}
+
+function expectedSet() {
```
leehopper, line 53: Optional tripwire: `exempt` entries are never checked against `git ls-files`, so if `pipeline_entry.js` is ever renamed the stale exemption goes silently inert. A one-line "every exempt entry must be a tracked file" check matches this tool's belt-and-suspenders character.

.github/workflows/ci.yml
leehopper: Done in 6421cdc — branch floor is now `82.3` in both `ci.yml` and the AGENTS.md gate command (drift test green).

AGENTS.md
leehopper: Done — floors sentence is past-tense to the CI basis (`JS 98 / 82.3 / 97 (pinned 2026-08-03 from first green CI: 98.61 / 83.27 / 97.99)`), and \"comfortably\" is cut so the raise rule is numeric: raise when measured headroom exceeds 1.0 pp.

CONTRIBUTING.md
leehopper: Done — `TestCoverageGateCommandIdentity.test_contributing_documents_the_ci_node_version_pin` extracts `node-version:` from the `workflow-tests` job and asserts the same string appears in CONTRIBUTING.md.

tests/test_contribution_surface.py
leehopper: Done — gate extraction now uses `.rstrip(\"\\n\")` on both the AGENTS fence split and the ci.yml `run: |` bodies.

workflows/test/tools/check_coverage_presence.mjs
leehopper: Done — `check_coverage_presence.mjs` exits 2 if any `exempt` entry is not git-tracked; covered by a new Node test case.

PR #128, opened 2026-08-03
chore(tooling): ruff config, format, and safe autofix (#104 part A)

leehopper:
## Summary
- Part A of #104 (Wave 2): add locked ruff/mypy config to `pyproject.toml`, apply `ruff format` at 0.16.1, then safe `ruff check --fix` only.
- Deliberately **does not** register pre-commit hooks, add `.git-blame-ignore-revs`, or document blame-ignore — those land in part B after this squash SHA exists on `main`.
- Design amendment: split at the mechanical/hand boundary because the repo is squash-only; a mixed squash would either break `git blame` (unresolvable ignore revs) or hide hand authorship.

## Test plan
- [x] `ruff==0.16.1` format `--check` clean over scoped paths
- [x] Safe `--fix` applied; no `--unsafe-fixes`
- [x] Suites: `pytest tests/`, `pytest bench/tests/`, `node --test workflows/test/*.test.js`, bundle freshness
- [x] Parity fixtures re-run as committed
- [x] Coverage floors on this tip: scripts+`.github` 92.50% (≥91.3), bench 88.21% (≥87)
- [ ] CI green on this PR

## Follow-up
Part B (hand fixes + mypy clean + hook registration + CONTRIBUTING + `.git-blame-ignore-revs` naming **this** PR's squash SHA) starts only after merge.

Closes nothing by itself — full #104 closes when part B merges.

Made with [Cursor](https://cursor.com)

PR #129, opened 2026-08-03
chore(tooling): gate ruff and mypy via pre-commit (#104 part B)

### LINKED ISSUE #104: Adopt ruff (check + format) and mypy behind the existing pre-commit gate, with an explicit pinned rule set

## Problem / Goal

The repo has no Python lint, format, or type configuration at all — no `pyproject.toml`, `ruff.toml`, `.flake8`, `mypy.ini`, or `setup.cfg` exists. The pre-commit gate covers YAML/TOML/JSON syntax, markdown, spelling, commit messages, and secrets, but nothing checks ~42k lines of Python across `scripts/`, `tests/`, and `bench/`. The intent was clearly there: 46 `# noqa` directives (`E402`, `BLE001`) are already written into the tree for a linter that never landed, and 38 of them are dead. Measured 2026-07-30 at ebf399d, ruff 0.16.1 reports 662 findings and `ruff format` would reformat 62 of 79 files; mypy reports 38 raw errors and cannot even run over `bench/` without configuration.

Leaving it costs three ways. Style drift is unbounded and every PR review spends attention on it. The 46 hand-written `noqa` comments are documentation nobody enforces — they can name a rule that would never have fired and no one finds out. And nothing pins a Python language level for static analysis, so a contributor on 3.14 (the measured local interpreter) and CI on its 3.10 floor check different languages. Fixed looks like: ruff check, ruff format, and mypy all run inside the existing `pre-commit run --all-files` gate with no change to `.github/workflows/ci.yml`; the enabled rule set is written down explicitly rather than inherited from a ruff default that grew from 59 to 413 available rules in a single 0.16.0 release; the tree is clean against it; and the four deterministic suites pass byte-identically, proving the change touched no behavior.

## Requirements

1. A single Python tool-config file at repo root is the only home for ruff and mypy settings. It declares no `[project]` table and no `[build-system]` table — nothing in the repo may be readable as an installable distribution.
2. Config-file ownership is respected: this issue owns `[tool.ruff]`, `[tool.ruff.lint]`, `[tool.ruff.lint.per-file-ignores]`, `[tool.ruff.format]`, and `[tool.mypy]`, and nothing else. If `pyproject.toml` does not exist when this lands, this issue creates it; if #102 landed first, this issue adds its tables to the existing file and does not touch `[tool.coverage.*]` or any dependency/pin table that issue introduced.
3. This issue introduces no dependency table of its own. Pinning for the tools it adds is done by `.pre-commit-config.yaml` `rev:` fields, which stay the single pin surface for hook-run tooling. (#102 owns the separate question of how the CI-invoked pytest stack is pinned; this issue must not contradict whatever it settles.)
4. Ruff's enabled rules are an explicit `select` list, so a change to ruff's default rule set cannot change what this repo enforces.
5. Ruff's `target-version` and mypy's `python_version` are both pinned to the lowest Python the CI test matrix runs (3.10). A local run on any newer interpreter reports the same findings as CI.
6. `ruff check` exits clean over `scripts/`, `tests/`, `bench/` (excluding `bench/vendor/` and `bench/workspace/`), and `workflows/test/tools/`.
7. `ruff format --check` exits clean over the same set, with no formatter options configured beyond the excludes.
8. `mypy` exits clean over the same set, at default strictness plus `warn_unused_ignores`, `warn_redundant_casts`, `warn_return_any`, and `no_implicit_optional`. It runs to completion in one invocation without a duplicate-module or module-found-twice failure.
9. `warn_unused_ignores` being on means every surviving `# type: ignore` is load-bearing; likewise, no `# noqa` directive in the tree names a rule that is not enabled — a stale directive fails the gate rather than sitting dead.
10. All three checks run from `pre-commit run --all-files` and therefore from CI's existing `lint` job. `.github/workflows/ci.yml` is unmodified by this issue.
11. No check runs from a `language: system` / `repo: local` hook. Each runs in pre-commit's own isolated environment, so the CI lint job — which installs only pre-commit — can execute it.
12. All four deterministic suites pass unchanged: `python -m pytest tests/ -q`, `python -m pytest bench/tests/ -q`, `node --test workflows/test/*.test.js`, and `node workflows/build.js && git diff --exit-code workflows/pipeline.js`.
13. The frozen parity fixtures under `tests/fixtures/parity/` are re-run and pass **as committed**. No fixture is re-recorded. This is the proof that the five JS-twinned scripts (`merge_findings`, `finding_dedup`, `filter_findings`, `apply_validations`, `apply_challenges`) are behaviourally untouched.
14. Adding the config file does not change pytest's behavior: no `[tool.pytest.ini_options]` section is introduced, and both pytest suites collect and pass exactly as before despite the new rootdir anchor.
15. `.git-blame-ignore-revs` exists, contains the whole-tree reformat commit, and CONTRIBUTING.md documents `git config blame.ignoreRevsFile .git-blame-ignore-revs`.
16. CONTRIBUTING.md states how a contributor runs the same checks locally at the same pinned versions as the gate.
17. No commit in the series leaves the gate red: the tree is clean against each tool before that tool's hook is registered.

## Evidence

All measurements 2026-07-30 at HEAD `ebf399d` (tag v3.3.4), working tree clean, over `scripts/`, `tests/`, `bench/` (excluding `bench/vendor/` and `bench/workspace/`), and `workflows/test/tools/`. Tool versions: ruff 0.16.1, mypy 2.3.0, local interpreter Python 3.14.6.

```text
$ ruff check          # no config, default rules
662 findings; 249 fixable with --fix; a further 239 only with --unsafe-fixes
(identical under --isolated, confirming no config file leaked in)

top codes: RUF059 125, UP032 110, UP031 97, I001 50, PLW1510 48, RUF100 38,
           SIM117 32, EXE001 21, F401 19, BLE001 17, SIM115 16, ISC004 12, RUF012 10

by area: tests/ 298, bench/ (non-test) 142, scripts/ 130, bench/tests/ 87, workflows/test/tools/ 5
         -> 385 of 662 (58%) are in test code; shipped scripts/ carries 130

$ ruff format --check
62 of 79 files would be reformatted (17 already clean): bench 26, tests 21, scripts 14, workflows/test/tools 1

$ ruff check --select ALL      # informational upper bound only, never a proposed config
13,739 findings. Top: PT009 3421, ANN201 1909, ANN001 1469, D102 1399, E501 1289,
ANN202 470, COM812 462, SLF001 183, D101 174, D205 154, ARG001 130, PTH118 128,
RUF059 125, D209 123, PTH123 122
```

- 2026-07-30: ruff 0.16.0 (released 2026-07-23) expanded the no-config default from 59 rules to 413 available with a much wider default-enabled subset — which is why the 662 already includes bugbear/pyupgrade/simplify/pylint/refurb/security rules.
- 2026-07-30, per-code area splits driving the per-file-ignores: RUF059 is 118/125 in `tests/` (plus 4 bench/tests/, 2 bench/, 1 scripts/); PLW1510 is 30/48 in `tests/`; UP031 is 60 scripts/ + 37 tests/; UP032 is 85 bench/ + 25 bench/tests/.
- 2026-07-30, mypy at defaults: `scripts/` 6 errors in 2 of 15 files (all `var-annotated` — `merge_findings.py:92`, `merge_findings.py:231`, `post_review.py:90`); `tests/` 19 errors in 7 of 22 files (9 `import-not-found`, 4 `var-annotated` duplicated from scripts via import-following); `bench/` 13 errors in 5 of 34 files (5 attr-defined/operator on `object`-typed vars, 3 import-not-found, 3 None-to-str assignment — `profile_run.py:240`, `profile_run.py:254`, `invoke.py:137`).
- 2026-07-30, mypy required flags: plain `mypy bench` fails outright with `Duplicate module named build_context` (from throwaway `bench/workspace/runs/**/build_context.py` smoke artifacts); adding the excludes but not `--explicit-package-bases` then fails with `Source file found twice under different module names: runner.score vs bench.runner.score`.
- 2026-07-30: mypy ran under Python **3.14** because no `python_version` is set anywhere (confirmed via `--verbose` config dump), while `ci.yml`'s test matrix floor is **3.10** — local and CI currently check different language levels.
- 2026-07-30: 46 `# noqa` directives already exist in the tree (`E402` and `BLE001` only); RUF100 reports **38 of them unused** under the default rule set. Seven `# noqa: BLE001` sites carry contract comments: `await_workflow.py:948`, `verify_findings.py:1276`, `materialize_artifacts.py:278/429/465/475`, `collect_project_rules.py:593/597`.
- 2026-07-30, parity-twin safety check: `grep -cE "%[sdrf]|\.format\("` returns **0** for all five JS-twinned scripts (`merge_findings.py`, `finding_dedup.py`, `filter_findings.py`, `apply_validations.py`, `apply_challenges.py`) — the 207 UP031/UP032 rewrites cannot reach them.
- 2026-07-30, presence checks at HEAD: `pyproject.toml`, `ruff.toml`, `.flake8`, `mypy.ini`, `setup.cfg`, `.python-version`, `.git-blame-ignore-revs`, and `.github/dependabot.yml` are all MISSING.
- 2026-07-30, gate location: `.github/workflows/ci.yml` lines 24-28 — the `lint` job installs only `pre-commit` (`pip install pre-commit`) and runs `pre-commit run --all-files`. The "install only pre-commit" trap is already documented at `.pre-commit-config.yaml` lines 69-71, where the `agent-instruction-layout` hook uses stdlib `unittest` rather than pytest for exactly this reason.
- 2026-07-30, repo size for context: `scripts/` 9,191 lines in 15 files; `tests/` 15,459 in 19 files; bench first-party Python 17,031 in 28 files; `workflows/test/tools/record_parity.py` 213 lines.

- **Tools:** ruff 0.16.1 (`astral-sh/ruff-pre-commit`, `rev: v0.16.1`, hooks `ruff-check` **without** `--fix`, and `ruff-format`) and mypy 2.3.0 (`pre-commit/mirrors-mypy`, `rev: v2.3.0`, `pass_filenames: false`, `args: ["--config-file", "pyproject.toml"]`, `additional_dependencies: []` — the runtime is stdlib-only so no stubs are needed). A rewriting *formatter* cannot change semantics; a rewriting *linter* can, which is why `ruff-check` reports and fails rather than fixing.
- **Config home:** root `pyproject.toml` with only the five tables named in requirement 2. Fallback if the owner wants the root free of anything packaging-shaped: `ruff.toml` + `mypy.ini` (two files, two edits to extend; mypy has no TOML home other than pyproject). Note that #102 may have already created `pyproject.toml` — if so, add to it.
- **Wave-1 select:** `["E4", "E7", "E9", "F", "I", "UP", "B", "SIM", "BLE", "RUF"]`. `E4/E7/E9` not full `E` because E501 is 1289 findings and line length belongs to the formatter. `BLE` because 7 existing `# noqa: BLE001` comments document the never-print-nothing contract — selecting BLE makes them load-bearing, not selecting it makes RUF100 call them dead. `RUF` because RUF100 is what keeps the 46 existing noqa directives honest.
- **per-file-ignores:** `"tests/**" = ["RUF059"]` and `"bench/tests/**" = ["RUF059"]` — 122 of the 125 RUF059 findings are unpacked helper returns in test assertions. Keep the ~18 `# noqa: E402` comments inline at the `sys.path`-then-import sites rather than a blanket per-file ignore: inline is self-documenting and RUF100 then verifies it.
- **Formatter left unconfigured** (88 cols, double quotes). Its value is that it is not configured; if a different line length is wanted, decide before the format commit lands — changing it afterwards costs a second whole-tree reflow and a second `.git-blame-ignore-revs` entry.
- **Excludes:** `bench/vendor` and `bench/workspace` for both tools. For mypy these are mandatory, not cosmetic (see Evidence). mypy also needs `explicit_package_bases = true` and `mypy_path = "scripts"` (the latter resolves the 9 import-not-found errors caused by scripts being imported as top-level modules via runtime `sys.path` insertion).
- **Ramp, four commits, each with the suites green:** (1) config only, tools not yet gated; (2) `ruff format` — 62 files, plus `.git-blame-ignore-revs`; (3) `ruff check --fix` safe fixes only, clearing ~249 — I001 (50), F401 (19), F841 (4), the `[*]`-marked UP032 (110), RUF022/046/015; (4) hand-fix the remainder (UP031 97, SIM117 32, SIM115 16, BLE001 17, RUF012 10, RUF100 38 stale directives deleted) and register the hooks last.
- **Never bulk-apply `--unsafe-fixes`** (239 beyond the safe 249). The measured mitigation is that all five parity-twinned scripts contain zero `%`-format and zero `.format(` sites, so UP031/UP032 provably cannot reach them — but re-run the parity fixtures *as committed* anyway, as the proof (requirement 13).
- Whether the ramp lands as one PR or a stacked series is a judgment call: one PR keeps the tree from sitting half-formatted; a stacked series makes the 62-file format diff separately revertable.
- Quick wins that stand alone: delete the 38 stale `# noqa` directives RUF100 flags (dead weight today, zero risk); add `.git-blame-ignore-revs` before or with the format commit (after a 62-file reflow lands without it, `git blame` and the GitHub blame view are permanently degraded for those files); pin `python_version = "3.10"` and `target-version = "py310"` (four lines, and without them 207 UP03x auto-rewrites could emit syntax the 3.10 CI job rejects).
- If this issue grows too large, the natural Wave-2 split is `PTH` (250+ findings) plus `PL` (48 PLW1510, 30 in tests where `check=` is deliberately unset), and a mypy per-module `disallow_untyped_defs` ratchet starting with the smallest stable modules (`scripts/finding_dedup.py` 57 lines, `scripts/validate_ndjson.py` 132, `scripts/review_marker.py` 409). Prefer keeping this issue single and filing that only once this one is merged and the real cost is known.

## Out of scope

- Rule families deliberately excluded, each with its measured size: PTH (250+ — an `os.path` → `Path` semantic rewrite inside scripts whose stdout is byte-exact-tested, not a lint fix), PL (48 PLW1510), ANN (3848), D (1850), PT (3421 PT009), EXE (21 EXE001 — shebangs on files correctly invoked as `python3 scripts/x.py`, mode 100644 is right). Each is a separate project.
- `select = ["ALL"]` as standing config. The `--select ALL` run stays a one-time inventory.
- Any mypy strictness beyond the named four warn-flags. No `disallow_untyped_defs`, no `strict`, no per-module strictness overrides.
- ty (0.0.65 beta, breaking diagnostics between any two versions) and pyrefly (GA and faster, but this codebase has under 30 unique mypy errors, so the trade is pure downside on troubleshooting corpus). Revisit pyrefly only if the mypy run becomes a felt cost.
- Coverage tooling (coverage.py, pytest-cov, `fail_under` thresholds) and the `[tool.coverage.*]` tables — #102 owns them, including in a `pyproject.toml` this issue may have created first.
- JavaScript lint/format for `workflows/` — #105 owns it.
- Swapping pre-commit for prek. A hook-runner change, not a Python-tooling change.
- Adding uv, a lockfile, or a `[dependency-groups]` table (see requirement 3).
- Adding `.github/dependabot.yml` to keep hook revs fresh — a real gap, and #106 owns the file. Note that Dependabot does not support the pre-commit ecosystem, so these `rev:` pins stay on the manual `pre-commit autoupdate` cadence either way.
- Any change to `.github/workflows/ci.yml`. If this issue needs one, the hook wiring is wrong.
- Refactoring, renaming, or behavior changes of any kind. Every commit in this series must be provably output-preserving.

## Verification

`ruff check`, `ruff format --check`, and `mypy` all exit clean; `pre-commit run --all-files` passes; `python -m pytest tests/ -q`, `python -m pytest bench/tests/ -q`, `node --test workflows/test/*.test.js`, and `node workflows/build.js && git diff --exit-code workflows/pipeline.js` all pass, with the `tests/fixtures/parity/` fixtures re-run as committed and not re-recorded. Measurement tier: `suites-only`.

leehopper:
## Summary
- Part B of #104: hand-fix remaining ruff/mypy findings, register pinned `ruff` 0.16.1 + `mypy` 2.3.0 hooks, document local commands in CONTRIBUTING, and add `.git-blame-ignore-revs` naming the **#128 squash SHA** (`1864baf`).
- Completes the two-PR split from the [#104 design addendum](https://github.com/liatrio-labs/claude-code-gauntlet/issues/104#issuecomment-5172603407): mechanical churn was PR-A; hand authorship stays attributable here.
- Hooks register only now that the tree is clean under the locked config.

## Test plan
- [x] `ruff check` / `ruff format --check` / `mypy` clean
- [x] `pre-commit run --all-files` green (includes new hooks)
- [x] `pytest tests/` + `pytest bench/tests/` + node workflow tests + bundle freshness
- [x] Parity fixtures as committed
- [x] `.git-blame-ignore-revs` contains only `1864baf336b57d4c27e5c7ab981c0832b66a93c9`
- [ ] CI green

Closes #104.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,14 @@ node workflows/build.js
 # Run full pre-commit checks across the repo
 pre-commit run --all-files
 
+# Run the pinned Python lint, format, and type checks
+pre-commit run ruff-check --all-files
+pre-commit run ruff-format --all-files
+pre-commit run mypy --all-files
+
+# Optional: ignore mass mechanical reflows in git blame
+git config blame.ignoreRevsFile .git-blame-ignore-revs
+
 # Run markdown linting only
 pre-commit run markdownlint-fix --all-files
 
--- a/bench/report.py
+++ b/bench/report.py
@@ -171,7 +171,7 @@ RELEASES = (
         ),
         "what_changed": (
             "Full workflow-native rewrite: an 8-stage pipeline running in the "
-            "Workflow runtime, −49% tokens vs v2."
+            "Workflow runtime, \u221249% tokens vs v2."
         ),
         "bar_check": False,
         "extra_note": "",
@@ -244,7 +244,7 @@ def git_short_sha(cwd=None):
         sha = out.stdout.strip()
         if out.returncode == 0 and sha:
             return sha
-    except Exception:
+    except (OSError, subprocess.SubprocessError):
         pass
     return "uncommitted"
 
@@ -267,7 +267,7 @@ def fmt_money(x):
 def fmt_int(x):
     if x is None:
         return "—"
-    return f"{int(round(x)):,}"
+    return f"{round(x):,}"
 
 
 def fmt_millions(x, digits=1):
@@ -285,7 +285,7 @@ def fmt_delta_pp(new, old):
 
 
 def fmt_delta_pct(new, old):
-    """Signed relative change, e.g. tokens −49%."""
+    """Signed relative change, e.g. tokens -49%."""
     if new is None or not old:
         return "—"
     return f"{(new - old) / old * 100:+.0f}%"
@@ -619,7 +619,9 @@ def _run_marker(x, y, tier, var, reverted, headline):
         oy = (-4.6, -1.8, 1.8, 4.6, 1.8, -1.8)
         d = (
             "M"
-            + " L".join(f"{x + dx:.1f},{y + dy:.1f}" for dx, dy in zip(ox, oy))
+            + " L".join(
+                f"{x + dx:.1f},{y + dy:.1f}" for dx, dy in zip(ox, oy, strict=True)
+            )
             + " Z"
         )
         g.append(f'<path class="mk mk-fill" d="{d}" style="fill:var({var})" />')
@@ -642,8 +644,8 @@ def _run_marker(x, y, tier, var, reverted, headline):
 def build_runs_svg(points, top_anchor, v2_base, ceiling):
     """Two stacked panels on a shared, time-ordered run axis.
 
-    Panel A is golden recall (0–100%) with the v2-baseline and top-anchor bars; panel
-    B is noise rate (0–40%) with the ceiling. Colour encodes the tool, marker shape
+    Panel A is golden recall (0-100%) with the v2-baseline and top-anchor bars; panel
+    B is noise rate (0-40%) with the ceiling. Colour encodes the tool, marker shape
     encodes the tier, reverted experiments are faded, and the two gate headlines carry
     a halo + direct label. Every marker has a ≥24px hover/focus target.
     """
@@ -695,7 +697,7 @@ def build_runs_svg(points, top_anchor, v2_base, ceiling):
             )
             parts.append(
                 f'<text class="axis" x="{m_left - 8:.1f}" y="{y + 3.5:.1f}" '
-                f'text-anchor="end">{int(round(t * 100))}%</text>'
+                f'text-anchor="end">{round(t * 100)}%</text>'
             )
 
     panel_grid((0.0, 0.25, 0.5, 0.75, 1.0), yA, A_DOM)
@@ -928,7 +930,8 @@ def build_explainer_html(top_anchor, v2_base, ceiling):
             ("Recall", "= goldens caught ÷ all goldens. The headline, gated metric."),
             (
                 "Noise rate",
-                "= noise ÷ all delivered comments — ‘how often are we wrong’. Gated: must stay under the ceiling.",
+                "= noise ÷ all delivered comments — \u2018how often are we "
+                "wrong\u2019. Gated: must stay under the ceiling.",
             ),
             (
                 "Valid-extra",
@@ -979,10 +982,10 @@ def build_explainer_html(top_anchor, v2_base, ceiling):
         "fail a gate; a mini run is the 6-PR paired cut; a subset run is gate-grade; "
         "a holdout run confirms it on fresh PRs; a custom run is an explicit --prs "
         "list (including pre-mini paired legs such as Gate-2). By owner "
-        "decision v3 delivers ~4× the comments of v2, so its lower precision is a "
+        "decision v3 delivers ~4\u00d7 the comments of v2, so its lower precision is a "
         "bigger denominator, not weaker findings. Gate-2 re-baselined its token target "
-        "to the v2 baseline on 2026-07-21: the original −20%-vs-Gate-1 target was "
-        "self-referential (it measured only −1.8%) and was retired. Two child-model "
+        "to the v2 baseline on 2026-07-21: the original \u221220%-vs-Gate-1 target was "
+        "self-referential (it measured only \u22121.8%) and was retired. Two child-model "
         "smokes render faded because they are confounded experiments — a subagent-model "
         "pin cascaded the [1m] context variant onto the pipeline agents — and were "
         "reverted.</p></div>"
@@ -1096,7 +1099,7 @@ def subset_comparison(rows, baselines):
 
 
 def _efficiency(row, n_goldens):
-    """Per-run value-efficiency figures. ``goldens`` is tp = round(recall × N) —
+    """Per-run value-efficiency figures. ``goldens`` is tp = round(recall x N) —
     distinct known issues caught, the numerator of golden_recall — NOT the
     per_bucket golden_matched comment count (a comment can match several goldens).
 
@@ -1203,10 +1206,10 @@ def build_verdict_html(rows, baselines, ceiling):
             "Comments delivered",
             fmt_int(a["delivered"]),
             fmt_int(b["delivered"]),
-            (f"{b['delivered'] / a['delivered']:.1f}×" if a["delivered"] else "—"),
+            (f"{b['delivered'] / a['delivered']:.1f}\u00d7" if a["delivered"] else "—"),
             arrow(b["delivered"], a["delivered"]),
             "context",
-            "~4× more by design — a bigger denominator, not worse findings",
+            "~4\u00d7 more by design — a bigger denominator, not worse findings",
         ),
     ]
 
@@ -1226,7 +1229,7 @@ def build_verdict_html(rows, baselines, ceiling):
 
     # Takeaway sentence, generated from the numbers above.
     ratio = (a["tok_per_gold"] / b["tok_per_gold"]) if b["tok_per_gold"] else None
-    ratio_txt = f"{ratio:.1f}× leaner" if ratio else "leaner"
+    ratio_txt = f"{ratio:.1f}\u00d7 leaner" if ratio else "leaner"
     noise_dir = "rises" if (b["noise"] or 0) > (a["noise"] or 0) else "falls"
     takeaway = (
         f"v3 catches {b['goldens']} of the {n_goldens} known issues to v2's "
@@ -1482,7 +1485,7 @@ def build_anchor_section_html(baselines, v3_row, v3_holdout_row):
     caption = (
         f"Every tool judged on {gate_desc}. <em>deep-review v3</em> leads on "
         "recall while holding noise far below the external tools; <em>claude</em> is "
-        "the published Claude Code CLI row. Anchors are the upstream tools’ stored "
+        "the published Claude Code CLI row. Anchors are the upstream tools\u2019 stored "
         "candidates re-judged under our pinned judge." + corroboration
     )
     return (
@@ -1516,7 +1519,7 @@ def build_table_html(groups, top_anchor, ceiling):
         run = html.escape(truncate_middle(rid))
         note = ""
         if grp["count"] > 1 and grp["identical"]:
-            note = f' <span class="note">×{grp["count"]} identical</span>'
+            note = f' <span class="note">\u00d7{grp["count"]} identical</span>'
         pb = r.get("per_bucket") or {}
         delivered = sum(pb.values()) if pb else r.get("total_candidates")
         kind, glyph, desc = classify(r, top_anchor, ceiling)
@@ -1602,9 +1605,9 @@ def build_footnotes_html(baselines, rows=None):
         "<b>† Precision (strict)</b> is reported, not gated: it counts valid-extras "
         "— real issues outside the golden answer key — as misses, so it drops as a "
         "run delivers more comments even when quality holds. Noise rate is the gated "
-        "‘how often are we wrong’ metric.",
+        "\u2018how often are we wrong\u2019 metric.",
         pin_line,
-        "Owner-amended N=1 protocol: runs 2–3, baseline extension, and full-50 "
+        "Owner-amended N=1 protocol: runs 2\u20133, baseline extension, and full-50 "
         "tracking were dropped under the budget cap.",
         f"δ_noise = {dn_txt} proposed {html.escape(str(dn.get('proposed', '')))}, "
         "pending owner sign-off.",
--- a/bench/report.py
+++ b/bench/report.py
@@ -245,6 +245,7 @@ def git_short_sha(cwd=None):
         if out.returncode == 0 and sha:
             return sha
     except (OSError, subprocess.SubprocessError):
+        # git missing, not a repo, or timeout — degrade to a stable label
         pass
     return "uncommitted"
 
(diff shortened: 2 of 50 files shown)


CONTRIBUTING.md
```diff
@@ -73,6 +73,14 @@ node workflows/build.js
 # Run full pre-commit checks across the repo
 pre-commit run --all-files
 
+# Run the pinned Python lint, format, and type checks
+pre-commit run ruff-check --all-files
+pre-commit run ruff-format --all-files
+pre-commit run mypy --all-files
+
+# Optional: ignore mass mechanical reflows in git blame
```
leehopper, line 81: **💡 [LOW] New ruff-check/ruff-format/mypy pre-commit commands added to CONTRIBUTING.md with no drift test against .pre-commit-config.yaml hook ids**

This PR adds three new documented commands (`pre-commit run ruff-check --all-files`, `ruff-format`, `mypy`) plus a `.git-blame-ignore-revs` git-config tip to CONTRIBUTING.md, and adds the corresponding `ruff-check`/`ruff-format`/`mypy` hooks to `.pre-commit-config.yaml`, but no test asserts the two stay in sync the way the repo already does for the cspell hook.

**Suggested fix:**
Add a test mirroring test_cspell_scope_includes_the_new_public_docs that parses `.pre-commit-config.yaml` for `id: ruff-check`, `id: ruff-format`, and `id: mypy`, and asserts each id string documented in CONTRIBUTING.md's `pre-commit run <id> --all-files` commands is actually declared as a hook id in `.pre-commit-config.yaml` (and vice versa), so renaming or removing a hook id fails CI instead of silently breaking the documented contributor workflow.

**Cited rule:** tests/test_contribution_surface.py covers the issue-form schema, .github/labels.json, and the CI commands quoted in CONTRIBUTING.md (scripts/CLAUDE.md / AGENTS.md, Contribution surface section)

scripts/merge_findings.py
```diff
@@ -191,7 +191,10 @@ def _try_parse_json_at(text: str, start: int) -> dict | None:
             if depth == 0:
                 candidate = text[start : i + 1]
                 try:
-                    return json.loads(candidate)
+                    parsed = json.loads(candidate)
+                    if isinstance(parsed, dict):
+                        return parsed
+                    return None
                 except json.JSONDecodeError:
                     return None
```
leehopper, line 196: **💡 [LOW] _try_parse_json_at mixes two distinct None-return reasons inside one try block**

The rewritten body wraps json.loads, an isinstance check, and its own return-None branch all inside the try, alongside a second except-driven return-None; separating the JSON-parse failure path from the type-narrowing (dict vs other) path into two sequential steps would make the two distinct reasons for returning None easier to tell apart at a glance.

**Suggested fix:**
Before: try: parsed = json.loads(candidate); if isinstance(parsed, dict): return parsed; return None; except json.JSONDecodeError: return None. After: try: parsed = json.loads(candidate); except json.JSONDecodeError: return None; return parsed if isinstance(parsed, dict) else None — same mypy-narrowing guarantee (dict | None) with the parse-exception path and the type-check path visually separated and a single return list.

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: 525398517239f14d3f431914270b04ca49909869

bench/report.py
```diff
@@ -244,7 +244,7 @@ def git_short_sha(cwd=None):
         sha = out.stdout.strip()
         if out.returncode == 0 and sha:
             return sha
-    except Exception:
+    except (OSError, subprocess.SubprocessError):
```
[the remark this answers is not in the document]
leehopper: Addressed in 0030a55 — added a comment above the `pass` clarifying that git missing / not a repo / timeout all degrade to `"uncommitted"`.

CONTRIBUTING.md
leehopper: Addressed in 0030a55 — added `TestContributingDocs.test_contributing_python_lint_hook_ids_match_pre_commit_config`, which asserts CONTRIBUTING documents exactly `{ruff-check, ruff-format, mypy}` via `pre-commit run <id> --all-files` and that those ids are declared in `.pre-commit-config.yaml`. Verified red on a one-line CONTRIBUTING rename.

scripts/merge_findings.py
leehopper: Addressed in 0030a55 — `json.loads` alone sits in the try; type-narrow (`return parsed if isinstance(parsed, dict) else None`) runs after a successful parse.

leehopper:
## Review follow-up

Pushed [0030a55](https://github.com/liatrio-labs/claude-code-gauntlet/commit/0030a55) addressing all three review threads:

1. **`bench/report.py`** — comment on intentional empty `except` in `git_short_sha`
2. **`scripts/merge_findings.py`** — separate JSON-parse failure from type-narrow in `_try_parse_json_at`
3. **`tests/test_contribution_surface.py`** — CONTRIBUTING ↔ pre-commit hook-id drift test for `ruff-check` / `ruff-format` / `mypy`

Issue #134, opened 2026-08-04
verify_findings.py's parse_diff_lines shares two of post_review.py's pre-fix diff-parsing defects


leehopper:
While implementing #127, `scripts/post_review.py::parse_diff_lines` was rewritten to track each hunk's declared old/new line budgets (`old_rem`/`new_rem`) so that file- and hunk-header regexes are only matched in the "between hunks" zone, never against hunk-body content, and to split diff text on `"\n"` only rather than `str.splitlines()` (`scripts/post_review.py:280-375`; zone-tracking rationale at lines 285-346, splitlines rationale at lines 291-296).

`scripts/verify_findings.py` has its own, independent `parse_diff_lines` (`scripts/verify_findings.py:256-302`), untouched by #127, that shares part of the same defect class:

- **Stale `current_file` across files — applies, and is the clearest hit.** The loop only resets `current_file` on a `+++ b/...` match (lines 276-280) and only skips lines when `current_file is None` (lines 288-289); there is no per-file "have I seen this file's own header yet" tracking. Any file in the diff that never emits a `+++ b/...` line — a binary file (`diff --git a/img.png b/img.png` / `index ...` / `Binary files a/img.png and b/img.png differ`, with no `---`/`+++` pair at all) — leaves `current_file` pointing at whatever file preceded it, and the binary file's `diff --git`/`index`/`Binary files ... differ` lines fall through to the final `elif not raw_line.startswith("\\")` branch (line 297) and get recorded as context lines of the PREVIOUS file at its stale `new_line` counter. `get_diff` (`scripts/verify_findings.py:200-224`) runs a plain `git diff {base}...HEAD` over the whole repo, so this triggers on any real multi-file diff that contains a binary file — not a contrived case. No `seen_hunk`/budget mechanism exists here at all, unlike the fixed `post_review.py`.
- **`str.splitlines()` — applies, direct one-line match.** Line 274: `for raw_line in diff_text.splitlines():`. This is exactly the call `post_review.py` now avoids on purpose (see its comment at `scripts/post_review.py:291-296`): `splitlines()` also breaks on `\x0c`/`\x0b`/`\x85`/U+2028/U+2029, which git treats as ordinary line content, desyncing the `new_line` counter for the rest of the file whenever one appears in a hunk body.
[...]

Issue #133, opened 2026-08-04
glab mr diff output shape is undocumented and load-bearing for parse_diff_lines — no fixture, no contract note


leehopper:
Two of the three #127 defects are direct consequences of the diff text `glab mr diff` emits — no `a/`/`b/` prefixes, the same path on both sides for added files, no `new file mode` line. The plain command differs from `glab mr diff --raw`, which does emit `--- /dev/null`. That contract lives only in test fixtures today, and a glab release that changes it would break line validation silently — findings quietly dropped as "line not found in diff" — rather than loudly.

Proposal: capture a checked-in fixture of real `glab mr diff` output for a modified file, an added file, a deleted file, a rename, and a binary file, and assert `parse_diff_lines` against it, plus a note in `scripts/AGENTS.md` that the plain (non-`--raw`) form is what the parser targets and why.

A further gap in the same undocumented contract: plain `glab mr diff` also cannot distinguish an added file from a pre-existing EMPTY file that gains content — both emit the same-path headers plus `@@ -0,0 +1,N @@`, with no `new file mode` line and no `/dev/null` anywhere to disambiguate. #127's fix deliberately prefers the added-file reading for this case — a documented limitation, pinned by `test_empty_file_gaining_content_is_treated_as_added_a_known_limitation` in `tests/test_post_review.py`. Notably, real `git` emits no `---`/`+++` headers at all for a genuinely empty added file, which is undetectable in the plain diff either way — the ambiguity is structural to the format, not a parsing bug.

**Provenance:** Found during the #127 fix (adversarial design/review); out of scope there to keep the fix reviewable.

Issue #132, opened 2026-08-04
Rerunning post_review.py after a partial GitLab delivery duplicates inline discussions — no dedup on retry


leehopper:
#127 made the GitLab summary note idempotent (skipped when this SHA's review marker is already on the MR) and made the per-finding inline-discussion loop fault-tolerant. The combination means the intended recovery from a partial delivery is "just rerun" — but nothing dedupes the inline discussions, so every finding that succeeded on the first attempt gets posted a second time on rerun.

Options:

- Fetch `.../discussions` and match on `(new_path, new_line, body)`, or on a per-finding marker embedded in the comment body.
- Record delivered findings in the output directory and skip them on rerun.

A per-finding marker is the more durable shape — it survives a fresh clone and reuses the `review_marker.py` machinery that already owns "what has this tool already written."

Related: #127 also reports that GitLab may create a discussion record before rejecting the request, so a failed POST cannot be assumed to be a no-op — any dedup mechanism must tolerate a dangling record.

**Provenance:** Found during the #127 fix (adversarial design/review); out of scope there to keep the fix reviewable.

Issue #131, opened 2026-08-04
--dry-run does not validate GitLab position shape — malformed payloads pass dry-run then fail live with HTTP 400


leehopper:
Named in #127's Additional Context but never filed separately.

`post_json`/`try_post_json` short-circuit into `_CAPTURED` before a payload reaches the network, so a dry run can report "12 inline discussion(s) captured", exit 0 — immediately before a live run fails on all twelve with HTTP 400. Dry-run is the pre-flight the pipeline actually runs, and today it validates nothing about payload shape.

Proposal: a pure `validate_position(position)` invoked on every captured GitLab discussion in dry-run, enforcing the platform contract #127 established:

- `new_line` always present
- `old_line` present for a context line and absent for an added line
- `old_path` absent for an added file
- `line_code` never present
- `base_sha` / `head_sha` / `start_sha` all non-empty

Violations would surface through `_SKIP_WARNINGS` into the payload's `skipped` array. Because this is pure and operates on the already-captured payload, it costs no network call — and it would have caught #127's D1 and D2 before the live post.

**Provenance:** Found during the #127 fix (adversarial design/review); out of scope there to keep the fix reviewable.

PR #135, opened 2026-08-04
fix(scripts): repair GitLab inline MR discussion delivery (#127, #130)

### LINKED ISSUE #127: [Bug]: GitLab inline discussions fail with 400 line_code — post_gitlab omits old_line for unchanged lines; new-file detection dead on glab mr diff
> **Maintainer intake note:** reported from a live run on a self-hosted GitLab MR (plugin v3.3.5 via a marketplace mirror pinning the upstream tag, byte-identical to upstream). All three defects verified still present on current `main`: position build at `scripts/post_review.py:609-617`, context-line admission at `:277`/`:284`, `--- /dev/null`-only detection at `:245`. Internal hostnames, project paths, SHAs, and MR ids are scrubbed; the position JSON structure and the verbatim GitLab error string are preserved as the diagnostic payload.

## Summary

On GitLab MRs, Phase 8 delivery posts the summary note but **every inline per-finding discussion fails** with:

```
glab: 400 Bad request - Note {:line_code=>["can't be blank", "must be a valid line code"]} (HTTP 400)
```

`die()` aborts the loop on the first failure, so **0 of 12 inline comments landed** and the run hard-exited after the summary note had already posted (non-idempotent partial state). The full review had to be pasted into chat instead of landing on the MR.

Three independent defects in `scripts/post_review.py`, all GitLab-path only. None is caught by the existing suite, and `--dry-run` cannot surface any of them.

## Environment

| Field | Value |
|---|---|
| Plugin version | 3.3.5 (upstream tag `v3.3.5`; defects confirmed still on `main`) |
| Install method | Marketplace (`claude plugin install`, private mirror pinning upstream tag) |
| Claude Code | 2.1.220 |
| Python | 3.14.6 (also reproduces on 3.11) |
| OS | macOS (Darwin 25.6.0, arm64) |
| `glab` | 1.111.0 |
| Platform | GitLab MR — self-hosted (`gitlab.example.com`) |
| Component | Pipeline script (`scripts/`) — `post_gitlab()` + `parse_diff_lines()` |
| Phase | Phase 8 — Report & deliver |

## Defect 1 (primary cause of the 400) — `old_line` is never sent, so unchanged/context lines are unaddressable

GitLab's rules for `POST /projects/:id/merge_requests/:iid/discussions` (<https://docs.gitlab.com/api/discussions/>):

| Line kind in diff | Required position fields |
|---|---|
| Added (`+`) | `new_line` only |
| Removed (`-`) | `old_line` only |
| **Unchanged / context** | **both `new_line` and `old_line`** |

`post_gitlab()` builds a position that only ever contains `new_line` (`scripts/post_review.py:604-617`), while `parse_diff_lines()` deliberately admits context lines as valid targets (`:283-285`) but throws away the old-side number it would need — the parser doesn't even track an old-side counter. So any finding on an unchanged line passes validation, gets posted, and is rejected. GitLab surfaces the underspecified position as `line_code can't be blank`, which is misleading: **`line_code` is a derived server-side field the client must not send.** (Two experiments synthesizing `line_code` as `sha1(path)_<old>_<new>` — as a position sibling and inside `line_range` — both failed with the identical 400.)

On the failing MR, 3 of the 12 findings that passed line validation targeted context lines. Note that docstring/comment-accuracy findings — which the `conventions-and-intent` dimension produces in volume — land on unchanged lines *by nature* (the code changed, the comment didn't). This is the common case for a whole dimension, not an edge case.

## Defect 2 — new-file detection is inert on GitLab, re-arming the HTTP 500 the code already documents

`parse_diff_lines()` detects added files by matching `--- /dev/null` (`scripts/post_review.py:249`), and `post_gitlab()` guards `old_path` with a comment explaining GitLab returns HTTP 500 (after silently creating the discussion) when `old_path` points into a new file (`:611-617`).

**`glab mr diff` never emits `--- /dev/null`.** It prints the same path on both sides for added files (real output from the failing MR):

```
--- src/app/clients/api/__init__.py
+++ src/app/clients/api/__init__.py
@@ -0,0 +1,16 @@
```

(`glab mr diff --raw` *does* emit `new file mode` + `--- /dev/null`; the plain `glab mr diff` the script calls at `:196-198` does not. glab 1.111.0.)

Reproduced against the live MR with pristine v3.3.5:

```python
import sys; sys.path.insert(0, 'scripts')
import post_review as pr
vl, nf = pr.parse_diff_lines('gitlab', 'group/subgroup/repo', 'repo', 123)
print(len(vl))    # 3270
print(sorted(nf)) # []   <-- MR adds 8 files; should be 8 entries
```

`new_files` comes back empty, `old_path` is set unconditionally, and 3 findings aimed at a new file were headed straight for the documented HTTP 500 (latent in this run only because Defect 1 aborted first). **On GitLab, that guard has never fired.**

Why the suite misses it: `tests/test_post_review.py::test_new_file_detected_with_no_prefix_headers` keeps a synthetic `--- /dev/null` marker real `glab mr diff` output never contains, and `test_new_file_position_omits_old_path` injects `new_files` directly, bypassing the parser. The `@@ -0,0` hunk header — the reliable added-file signal in `glab mr diff` output — is unused.

## Defect 3 — first failure aborts the batch, after non-idempotent state already exists

`post_json()` calls `die()` → `sys.exit(1)` on any non-zero `glab` exit (`scripts/post_review.py:143-149`, `98-100`). In `post_gitlab()` the summary note posts **before** the per-finding loop (`:565-575`). Observed chain: summary posted → finding 1 (added line) posted → finding 2 (context line) → 400 → process exits → findings 3-12 never attempted. Re-running to retry **re-posts the summary note** — there is no skip, so a partial failure leaves the run un-retryable without duplicating it. A single malformed position should `warn_skip` + `skipped += 1` (the loop already has exactly that shape for line-validation misses) and let the rest through.

## Steps to Reproduce

```
1. GitLab MR target (self-hosted; glab authenticated). The MR must (a) add at
   least one new file, and (b) have at least one finding on an unchanged/context
   line of the diff.
2. Run the code-gauntlet skill against the MR.
3. Let Phase 8 deliver via scripts/post_review.py.
4. Observe: "MR summary note posted." then an immediate hard failure.

# Minimal confirmation of Defect 2 alone, no posting required:
5. python3 -c "
import sys; sys.path.insert(0,'scripts')
import post_review as pr
vl, nf = pr.parse_diff_lines('gitlab','<group/repo>','<repo>',<mr_iid>)
print(len(vl), sorted(nf))"
   -> new_files is [] even though the MR adds files.
```

## Expected Behavior

- Findings on unchanged/context lines post successfully, with both `old_line` and `new_line` in `position`.
- `new_files` is populated correctly from `glab mr diff` output, so `old_path` is omitted for added files and the documented HTTP 500 is actually avoided.
- One rejected position skips that finding with a warning; the remaining findings still post.
- A retry after partial failure does not duplicate the summary note.

## Actual Behavior

- Every inline discussion POST fails with HTTP 400 `line_code can't be blank / must be a valid line code`.
- `new_files` is always empty on GitLab, so `old_path` is always sent — including for added files.
- The first 400 exits the process; 0 of 12 inline comments delivered; the already-posted summary note makes retry non-idempotent.

## Logs / Output

```
Detected platform: gitlab (from git remote: gitlab.example.com)
MR summary note posted.
ERROR: API call failed (exit 1).
Command: glab api --method POST --header Content-Type: application/json \
  projects/<group>%2F<repo>/merge_requests/123/discussions --input /tmp/tmpXXXX.json
stderr: glab: 400 Bad request - Note {:line_code=>["can't be blank", "must be a valid line code"]} (HTTP 400)
```

Rejected payload (position verbatim in structure; `new_line: 61` is an unchanged line whose old-side number is 50 — no `old_line` present):

```json
{
  "body": "**[LOW] Class docstring still describes the old model** ...",
  "position": {
    "position_type": "text",
    "base_sha":  "<base-sha>",
    "head_sha":  "<head-sha>",
    "start_sha": "<start-sha>",
    "new_path":  "src/app/models/evidence.py",
    "old_path":  "src/app/models/evidence.py",
    "new_line":  61
  }
}
```

For contrast, a position GitLab itself stored for a working inline comment on the same MR — `line_code` absent, `old_line` explicitly `null` for an added line:

```json
{
  "base_sha": "<base-sha>", "start_sha": "<start-sha>", "head_sha": "<head-sha>",
  "old_path": null,
  "new_path": "src/app/fetch_cases.py",
  "position_type": "text",
  "old_line": null, "new_line": 619, "line_range": null
}
```

## Suggested Fix

1. **Track and emit `old_line`.** Have `parse_diff_lines()` return line metadata instead of a bare set — e.g. `(filepath, new_line) -> old_line | None` — by maintaining an old-side counter from the `@@ -old,… +new,… @@` header (the old-side group is currently parsed and discarded at `:267-271`). In `post_gitlab()`: set `position["old_line"]` when non-None; keep sending `new_line`; never synthesize `line_code`.
2. **Detect added files from the hunk header.** Treat `@@ -0,0 +N,M @@` as an added-file marker in addition to `--- /dev/null`, so detection works for both `gh pr diff` and `glab mr diff`; fix the tests to assert against a fixture captured from real `glab mr diff` output.
3. **Make the per-finding loop fault-tolerant.** Non-fatal mode for the discussion POST: `warn_skip` + `skipped += 1` on rejection, continue, report `posted/skipped/failed`, exit non-zero only if everything failed.
4. **Make the post idempotent enough to retry.** Skip the summary note when this SHA's review marker is already on the MR (`detect_prior_review.py` / `review_marker.py` own that signal).
5. **Test the GitLab position contract directly.** Context line → `old_line` present and correct; added line → no `old_line`; added file → no `old_path`; `line_code` never sent. No current test inspects `position` for `old_line`.

## Additional Context

- **`--dry-run` cannot catch any of this**: `post_json()` short-circuits into `_CAPTURED` (`:139-141`) before the payload reaches GitLab; the dry run reported `12 inline discussion(s) captured`, exit 0, immediately before the live run failed on every comment. A position-shape self-check in dry-run mode would close that gap.
- **Line validation is not implicated** — it correctly skipped 6 of 18 findings outside the diff with the designed diagnostics. The 400s are entirely about position shape for the 12 that passed.
- **Possible dangling discussions**: the MR's `updated_at` advanced to match the last failed POST while no corresponding note is visible via `/notes` or `/discussions` — consistent with the code's own note that GitLab "silently creat[es] the discussion" before rejecting. Unverified, but retry logic should not assume a failed call was a no-op.

### LINKED ISSUE #130: post_gitlab sends the new path as old_path for renamed files — parse_diff_lines never captures the old-side path
`post_gitlab` sets `position["old_path"] = filepath` — the NEW path — for every file that is not newly added. For a renamed file, GitLab expects `old_path` to be the pre-rename path; sending the new path on both sides is a mis-specified position, the same class of defect as the missing `old_line` fixed in #127.

`parse_diff_lines` currently discards the old-side path entirely — the `--- ` header's path is read only to test it against `/dev/null` — so the data needed to fix this isn't even captured yet.

Fix: record the old-side path per file alongside the added-file flag, and emit it as `old_path` only when it differs from the new path.

Out of scope for #127, which was confined to the three defects reproduced on the reporting MR; a rename in the diff would be a fourth.

**Provenance:** Found during the #127 fix (adversarial design/review); out of scope there to keep the fix reviewable.

leehopper:
Closes #127. Closes #130.

## What was broken

Live GitLab usage (issue #127, reported 2026-08-03) hit a total breakage of inline MR delivery: every per-finding discussion POST failed with `400 line_code can't be blank`, the first failure aborted the batch after the summary note had already posted, and a rerun duplicated that note. Three verified defects, plus a fourth (#130) found during review:

1. **`old_line` was never sent.** GitLab requires *both* `old_line` and `new_line` to address an unchanged/context line; `parse_diff_lines` discarded the old-side counter entirely, so every context-line finding (the natural home of docstring/comment findings) was rejected.
2. **Added-file detection was dead on GitLab.** Detection matched only `--- /dev/null`, which plain `glab mr diff` never emits — it repeats the same path on both sides. `new_files` was permanently empty, so `old_path` was always sent, re-arming the HTTP 500 the code already documented.
3. **One rejection stranded the batch.** `post_json` → `die()` on the first 400, after non-idempotent state existed.
4. **Renamed files sent the new path as `old_path`** (#130) — the same mis-specified-position class; the parser never captured the old-side path.

## What changed (`scripts/post_review.py`, `scripts/detect_prior_review.py`)

- **Old-side tracking.** `valid_lines` is now a mapping `(filepath, new_line) → old_line | None`; positions carry `old_line` for context lines, omit it for added lines, and never synthesize `line_code`.
- **Budget-gated diff parsing.** The parser tracks each hunk's declared old/new line counts and only matches file headers *between* hunks. This fixes added-file detection via `@@ -0,0` (the only signal glab emits), and hardens against three silent-desync classes found by adversarial review: hunk lines whose content starts with `--`/`++` (previously swallowed by the header regexes), pre-hunk noise (`index`, `Binary files … differ`) admitted as fake comment targets, and `str.splitlines()` splitting on form-feed/U+2028 characters git treats as line content.
- **Rename anchoring** (#130). The parser returns `old_paths` (new path → `---` header path); `position.old_path` anchors there, so renamed files use the pre-rename path GitLab requires. Added files still omit `old_path` entirely.
- **Fault-tolerant delivery.** A rejected discussion warns (with the verbatim glab error), skips that finding, and continues; final counters report `posted / skipped / rejected`. Exit 1 only when findings existed and *every* attempted post was rejected. Summary-note failure is still immediately fatal (nothing behind it can succeed).
- **Retry-safe summary note.** Before posting, `post_gitlab` asks `detect_prior_review.gitlab_note_exists_for_sha` (the marker's only reader — `post_review` stays write-only) whether this SHA's review marker is already on the MR, and skips the note if so. Dry-run performs no such read; an unresolvable SHA or failed fetch degrades to posting.
- **Platform-faithful path spelling.** `glab mr diff` emits paths verbatim, so a leading `a/`/`b/` is a real directory there, not diff syntax: synthetic prefixes are stripped only for `gh` output, and the position ships whichever spelling the parsed diff actually recorded (raw passthrough when validation was skipped). A repo with a genuine top-level `a/` directory keeps working on GitLab.

GitHub-path behavior is byte-identical (payloads, stdout, stderr).

## Verification

- **Suites:** `tests/` 1108 passed (+413 subtests), `bench/tests/` 529 passed, JS 608 pass; `workflows/pipeline.js` bundle unchanged. Boundary-parity tests green as committed (post_review.py has no JS twin; no parity fixture re-record needed).
- **Coverage gates:** scripts+.github 92.79% (floor 91.3), bench 88.29% (floor 87), JS 98.61/83.10/97.99 — all verbatim AGENTS.md commands. `pre-commit run --all-files` green.
- **Mutation-verified regression tests** (AGENTS.md rule): a 24-mutation catalogue (M1–M5 families for the four fixes; M6–M17 for the parser hardening, rename anchoring, and path-spelling mechanisms) was applied whole-mechanism by an independent verifier; every mutation turned exactly its named tests red, and the two misleading tests #127 called out were rewritten against real tool output shapes.
- **Process:** five implement/verify rounds, each gated by adversarial review (one round was blocked and re-fixed: the `--`-content header collision that would have shipped *wrong* `old_line` values silently; a later round caught and fixed the branch's own `a/`-directory path truncation regression).
- Position contract pinned end-to-end from realistic `glab mr diff` fixtures: context line → both sides; added line → `new_line` only; added file → no `old_path`; renamed file → pre-rename `old_path`; `line_code` never present anywhere in the payload.

## Deliberate deferrals (all filed and threaded into #101)

- **#132 — rerun after partial delivery duplicates inline discussions.** Deferred by design tonight: it's an enhancement that only bites after a partial failure *and* a rerun, and a durable fix wants a per-finding marker design, not an overnight patch. The summary note is already rerun-safe.
- **#131** dry-run position-shape validation (would have caught #127 pre-live), **#133** the undocumented load-bearing `glab mr diff` output contract (including the empty-file/added-file ambiguity this PR documents in a pinned limitation test), **#134** `verify_findings.py`'s independent diff parser sharing two pre-fix defect classes.

## Notes for the reviewer

- Known limitation (pinned by test): a pre-existing *empty* file that gains content is indistinguishable from an added file in plain `glab mr diff` output; the added-file reading is preferred because the alternative is the documented HTTP 500.
- Python scripts coverage floor raised 91.3 → 91.7 in the final commit, per the AGENTS.md policy: this PR's first CI run measured 92.69 on 3.12 (1.39 pp headroom > the 1.0 pp trigger). Bench stays at 87 — its 1.08 pp headroom drifted in on `main` before this branch.

https://claude.ai/code/session_019auKVBEyXrRbDz5T5Z8FV2

---

> [!NOTE]
> **Medium Risk**
> Changes live GitLab posting and diff parsing on real MRs (positions, idempotency, partial-failure retries); GitHub path is unchanged and coverage is heavily tested.
> 
> **Overview**
> Repairs **GitLab inline MR discussion delivery** (#127, #130): context-line findings were rejected without `old_line`, added files were mis-detected on `glab mr diff`, one API failure aborted the whole batch after the summary note posted, and renames used the wrong `old_path`.
> 
> **`parse_diff_lines`** now returns `valid_lines` as a map from `(path, new_line)` to old-side line (or `None` for additions), plus `new_files` and `old_paths`. Parsing is **platform-specific** (`gh` strips `a/`/`b/`; `glab` keeps paths verbatim), uses **hunk line budgets** so body lines like `--- comment` are not mistaken for headers, detects new files via `@@ -0,0` when `/dev/null` is absent, and splits on `\n` only.
> 
> **`post_gitlab`** sends `old_line` on context lines, omits `old_path` for new files, anchors renames via `old_paths`, resolves paths with `diff_path_spelling`, posts discussions through **`try_post_json`** (warn/skip/continue on rejection; exit 1 only if every inline post failed), and **skips duplicate summary notes** on rerun by delegating to `detect_prior_review.gitlab_note_exists_for_sha` / `entries_carry_sha` (exact SHA match, paginated notes fetch).
> 
> **Tests and docs:** large regression suite with realistic `glab` fixtures; CI/AGENTS.md coverage floor **91.3 → 91.7**; mypy override for `detect_prior_review`; bench adapter comment only. GitHub behavior unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85586e084ef9d2a6c0887b649ac25c55bfd67182. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/AGENTS.md
```diff
@@ -15,6 +15,8 @@ assume a language in the reviewed codebase.
   `post_review.py` writes and parses what `detect_prior_review.py` reads back. Readers never branch
   on the payload's `version` field — both token generations carry `"version":"3.0"` despite being
   different wire shapes. `tests/test_review_marker.py::TestRoundTrip` guards the agreement.
+  `post_review.py` never parses the signal it writes — its GitLab summary-note idempotency check
+  calls `detect_prior_review.gitlab_note_exists_for_sha`, so the reader stays the only reader.
```
leehopper: **💡 [LOW] AGENTS.md addition duplicates a rule already stated by code comments at the site**

REVIEW.md instructs reviewers to flag additions to CLAUDE.md/AGENTS.md that are derivable from the code or already stated in a comment at the site. The new bullet ('post_review.py never parses the signal it writes ... so the reader stays the only reader') restates, almost verbatim, the comment already added directly above the `from detect_prior_review import gitlab_note_exists_for_sha` line in scripts/post_review.py and the docstring of gitlab_note_exists_for_sha in scripts/detect_prior_review.py, both added in this same diff.

**Suggested fix:**
Drop the new scripts/AGENTS.md/CLAUDE.md bullet and rely on the code comments already present at both sites (or trim the code comments to a single canonical location and reference it from the other), per AGENTS.md's own 'build the mechanism, not the instruction' principle.

**Cited rule:** REVIEW.md Rules: 'Instruction files are loaded on every turn. Flag additions to CLAUDE.md or AGENTS.md that are derivable from the code, already stated in a comment at the site, or that record history, version numbers, or past incidents.'

leehopper (commented):
Code Gauntlet review (2 high-confidence findings; 1 main, 1 suggestion). Full report available on request.

---
Generated by code-gauntlet | Reviewed up to: 1d26d40c37e9be8271404312ec37ba9bfc1c1b73

scripts/AGENTS.md
leehopper: Fixed in 85586e0 — dropped the AGENTS.md tail sentence, kept the non-duplicated `review_marker.py` single-source-of-truth fact. Also fixed the other high-confidence finding from this same review (`bug-1`, not delivered inline — `is_new_file`'s stripped-prefix fallback could misreport a modified real-`a/`-rooted file as new when it collided with an unrelated added file's stripped basename).

Issue #106, opened 2026-07-31
Harden the Actions supply chain: SHA-pin every uses:, add dependabot.yml, and gate .github/workflows with zizmor


leehopper:
Filed from the #55 engineering-quality audit (see docs/engineering-audit-2026-07.md).

## Problem / Goal

All 7 workflow files pin their actions by mutable tag rather than commit SHA: zizmor 1.28.0 reports 23 High `unpinned-uses` findings covering 23 of the 24 `uses:` references in the repo (10 distinct actions; the sole SHA-pinned exception is the reusable-workflow call in `publish-marketplace.yml`). Alongside those sit 6 Medium + 1 Low `artipacked` findings — only `release.yml` sets `persist-credentials: false`, so the ephemeral `GITHUB_TOKEN` is written to `.git/config` and lingers for the rest of every other job — and 1 Low `adhoc-packages` for an unpinned global npm install. There is no `.github/dependabot.yml` at all, and `dependabot_security_updates` is `disabled`, so nothing in the repo moves a dependency pin forward.

Leaving this costs the repo a live tj-actions-class exposure (CVE-2025-30066, ~23k repos, CISA alert, March 2025): a repointed tag on `octo-sts/action@v1.1.1` executes with a `contents: write` federated token on every push to `main`, and repointed tags on `amannn/action-semantic-pull-request@v5` or `marocchino/sticky-pull-request-comment@v2` execute with `pull-requests: write` + `statuses: write` on every same-repo PR. It also leaves the release path — the one automation nobody watches interactively — dependent on upstream tags nobody controls. Fixed looks like: every `uses:` is a 40-char SHA with a version comment, Dependabot opens one grouped PR a week to refresh those SHAs, zizmor runs inside the existing `pre-commit run --all-files` gate so a tag pin can never come back, and every package installed inside a workflow has an exact version.

## Requirements

1. Every `uses:` reference under `.github/workflows/**` resolves to a 40-character commit SHA followed by a comment naming the human-readable version it corresponds to (e.g. `uses: actions/checkout@<sha> # v4.2.2`). Baseline: 23 of 24 references are tag-pinned today.
2. `.github/workflows/publish-marketplace.yml`'s reusable-workflow reference (`liatrio-labs/claude-plugins/.github/workflows/publish-to-marketplace.yml@6b89d0b0…`) and the comment block explaining that deliberate diagnostic pin are unchanged.
[...]

PR #136, opened 2026-08-04
ci: gate workflows JS on checksum-pinned Biome lint (#105)

### LINKED ISSUE #105: Gate workflows/*.js on a pinned CI-only Biome lint job with no repo-tree npm footprint

## Problem / Goal

The JS subsystem — 5,638 lines in `workflows/src/*.js`, 111 in `workflows/build.js`, 7,787 in `workflows/test/` — has no static analysis of any kind, because the repo carries no npm artifacts and every mainstream JS linter has historically assumed a `package.json`. Meanwhile `workflows/AGENTS.md` documents that the workflow sandbox provides none of `structuredClone`, `Buffer`, `TextEncoder`/`TextDecoder`, `URL`, `setTimeout`, `queueMicrotask`, `process`, or `console`, and that a stray reference "throws on the first live dispatch while every test stays green"; `applyChallenges.js:16` records `structuredClone` crashing a live smoke. That invariant is enforced today only dynamically, by `pipeline_run.test.js` deleting those globals from `globalThis` — which catches a violation only on a covered path.

Leaving it costs a class of failure the deterministic suites structurally cannot see: a reference on an uncovered branch ships green and fails on a live customer run, which is exactly how it failed before. It also leaves confirmed dead bindings (5 unused variables, 3 unused catch bindings) accumulating with no gate. Fixed looks like: a `js-lint` job in `ci.yml` runs a pinned, checksum-verified linter binary from `$RUNNER_TEMP` against `workflows/`, the banned-globals deny-list is a static error rather than a comment, the workflow-host injected globals (`args`/`agent`/`parallel`/`pipeline`) are declared in config so they never produce a false positive, and the repo tree still contains no `package.json`, no lockfile, and no `node_modules`.

## Requirements

1. A `js-lint` job exists in `.github/workflows/ci.yml`, triggered on the same events as the existing jobs, and it fails the build when a lint error is present in `workflows/**/*.js`.
2. The job obtains its linter as a version-pinned binary whose SHA-256 is verified against a literal committed in the workflow file; the build fails if the digest does not match. No package manager installs the linter.
3. After a full CI run of the new job, the repo tree contains no `package.json`, no lockfile, and no `node_modules` directory, and `git status --porcelain` is empty.
4. The linter configuration declares `args`, `agent`, `parallel`, and `pipeline` as known globals, such that no diagnostic is emitted for any reference to them.
5. `workflows/pipeline.js` (generated) and `workflows/src/pipeline_entry.js` (top-level `return` entry contract; `PIPELINE_VERSION` read out-of-band) are excluded from the gate, and the run emits zero parse errors.
6. A restricted-globals rule makes it a lint error for any file under `workflows/src/` to reference `structuredClone`, `Buffer`, `TextEncoder`, `TextDecoder`, `URL`, `setTimeout`, `setInterval`, `clearTimeout`, `queueMicrotask`, `process`, or `console`, and that rule does not apply to `workflows/test/**` or `workflows/build.js`.
7. The restricted-globals rule is proven to actually bite: adding one such reference to a file in `workflows/src/` makes the job fail, verified by mutation rather than by reading the config.
8. Unused-variable detection is enabled and the gate passes at HEAD — meaning the 5 unused variables and the 3 unused catch bindings at `workflows/src/args.js:238`, `workflows/src/mergeFindings.js:131`, and `workflows/src/stages.js:225` are resolved in this change.
9. The formatter is disabled, and the configuration records why (default `lineWidth` wrapping would break `build.js`'s single-line import strip).
10. Rules deferred rather than adopted each carry their measured hit count as the recorded reason, so a later re-evaluation does not re-derive the numbers.
11. `node workflows/build.js && git diff --exit-code workflows/pipeline.js` passes, and `node --test workflows/test/*.test.js`, `python -m pytest tests/ -q`, and `python -m pytest bench/tests/ -q` are all green.
12. `AGENTS.md` states the no-npm boundary in the terms the change relies on: no npm artifacts tracked in the tree, zero dependencies in the shipped pipeline runtime, pinned static binaries permitted in CI. `python3 scripts/sync_agent_rules.py` has been re-run so any generated twin matches.
13. **Owner action (GitHub settings):** the new `js-lint` job's check-run name is added to the required status-check contexts on ruleset `protect-default-branch` (id 16049246) — a new job is not a required check until that click happens, so without it the gate ships non-blocking. If #108 has already landed, its frozen check-name list in `tests/` is updated in the same PR that adds the job, and the mutation test for that list is re-run.

## Evidence

All measurements 2026-07-30, HEAD `ebf399d` (v3.3.4).

```text
$ biome check workflows/src/*.js workflows/build.js workflows/test/*.js --reporter=json   # Biome 2.5.6, 32 files
180 diagnostics — 66 errors, 113 warnings, 1 info
by rule: noUnusedFunctionParameters 55, useOptionalChain 51, format 32 (one per file
that would be reformatted), organizeImports 24, noAssignInExpressions 7,
noUnusedVariables 5, noControlCharactersInRegex 2, parse 1, noPrototypeBuiltins 1,
useArrowFunction 1, useTemplate 1

$ oxlint workflows/... --format json     # oxlint 1.76.0, 32 files, 96 rules loaded
23 diagnostics — eslint(no-unused-vars) 11, unicorn(prefer-string-starts-ends-with) 7,
unicorn(no-useless-fallback-in-spread) 3, eslint(no-control-regex) 1, unicorn(no-useless-spread) 1
(distinct-from-Biome signal: 11 findings, all stylistic)

$ tsc --noEmit --allowJs --checkJs --target es2023 --module nodenext \
      --moduleResolution nodenext workflows/src/*.js workflows/build.js     # tsc 7.0.2
453 errors. TS7006 (implicit-any parameter) 343 = 76%; TS7053 23, TS7005 17,
TS7031 13, TS7034 13; TS2339 18; TS2591 6 (absent @types/node)
```

- 2026-07-30: grep for `@param`/`@returns`/`@type`/`@typedef` across all 9 `workflows/src/*.js` returns **0** — `checkJs` is verifying no author-stated contract, which is why type checking is out of scope here.
- 2026-07-30, known false-positive set, 10 diagnostics, all from workflow-host injected globals and the entry contract: Biome `parse` ×1 at `pipeline_entry.js:39`; tsc TS2304 ×6 (`args` at `pipeline_entry.js:38`, `parallel`/`pipeline` at `stages.js:25-26`), TS2552 ×2 (`agent` at `stages.js:24`), TS1108 ×1. Documented at `workflows/AGENTS.md` and inline at `stages.js:5-7,20-21` and `pipeline_entry.js:5-8`.
- 2026-07-30, eleventh false positive, distinct class: oxlint flags `PIPELINE_VERSION` unused at `pipeline_entry.js:22`, but it is a version stamp read out-of-band by `.releaserc.toml:5` (semantic-release rewrites it on every release), `tests/test_bundle_fresh.py:29`, `skills/code-gauntlet/SKILL.md:493`, and `skills/code-gauntlet/references/headless-mode.md:112`.
- 2026-07-30, banned-globals baseline: `grep -nE '\b(structuredClone|Buffer|TextEncoder|TextDecoder|setTimeout|queueMicrotask|process\.|console\.)' workflows/src/*.js` returns 2 hits, both inside comments (`applyChallenges.js:16`, `stages.js:2338`). A restricted-globals rule therefore lands green at HEAD — a ratchet, not a cleanup.
- 2026-07-30, formatter hazard: `awk '/^import /{print length}' workflows/src/*.js` — `stages.js:18` is 163 chars and `stages.js:13` is 104 chars; 221 of 5,638 src lines exceed 100 chars. Biome's default `lineWidth: 80` wraps them, and `build.js` `strip()` uses `/^\s*import\s.+from\s.+;?\s*$/`, which requires `from` on the same line. A wrapped import therefore survives into `workflows/pipeline.js` and fails `tests/test_bundle_fresh.py`'s import-free assertion.
- 2026-07-30, boundary precedent: `.github/workflows/validate.yml` runs `npm install -g @anthropic-ai/claude-code` on every `pull_request` and on push to `main`. npm already executes on the CI runner; the no-npm rule governs the tree and the shipped runtime.
- 2026-07-30, tree-cleanliness proof: after Biome, oxlint, and tsc all ran against the repo by absolute path from an isolated scratchpad (npm cache redirected, repo never used as cwd), `git -C <repo> status --porcelain` was empty.
- 2026-07-30, current gates (`.github/workflows/ci.yml`): `lint` (pre-commit only), `test` (pytest `tests/` on 3.10/3.11/3.12), `workflow-tests` (`node --test workflows/test/*.test.js` on Node 24, plus `node workflows/build.js && git diff --exit-code workflows/pipeline.js`), `bench-tests` (pytest `bench/tests/` on 3.11/3.12). No JS lint job exists; `.pre-commit-config.yaml` has no JS hook.

- **Biome 2.5.6, lint only**, downloaded from `https://github.com/biomejs/biome/releases/download/@biomejs/biome@2.5.6/biome-linux-x64` into `$RUNNER_TEMP`, `chmod +x`, invoked as `biome check --config-path=workflows --error-on-warnings workflows/`. Biome ships no project-signed checksum manifest, so take the SHA-256 once from GitHub's asset-digest API (`gh api repos/biomejs/biome/releases/assets/<id>`) and pin it as a literal in the workflow file.
- The job needs **no `actions/setup-node` step**, unlike the adjacent `workflow-tests` job — the binary is static, which is the whole point.
- Put the config at `workflows/biome.json` rather than repo root, matching the repo's directory-scoped-rules convention and keeping the root clean. It is the only new tracked repo-tree artifact under this approach.
- `files.includes: ["workflows/**/*.js", "!workflows/pipeline.js", "!workflows/src/pipeline_entry.js"]`; `formatter.enabled: false`; `assist.enabled: false`; `javascript.globals: ["agent", "parallel", "pipeline", "args"]`. Excluding the 39-line entry file is load-bearing: it kills the Biome `parse` error, the tsc TS1108 and TS2304 `args` hits, **and** the oxlint `PIPELINE_VERSION` unused hit in one line.
- `linter.rules.recommended: false` with an explicit ON list — `correctness/noUnusedVariables`, `correctness/noUndeclaredVariables`, `complexity/noUselessCatch`, `suspicious/noPrototypeBuiltins`, and `style/noRestrictedGlobals` with the deny-list, plus an `overrides` entry disabling the last for `workflows/test/**` and `workflows/build.js` (both legitimately use `process`/`console`). Confirm the `deniedGlobals` option schema against the shipped 2.5.6 binary rather than assuming it.
- Explicitly OFF with counts recorded as the reason: `complexity/useOptionalChain` (51), `correctness/noUnusedFunctionParameters` (55 — dominated by positional mock callback params in tests; oxlint's default config reported only 11 for the same class, so 55 is a config default, not a defect density), `suspicious/noAssignInExpressions` (7 — 6 are the intentional `args.js:195-200` pattern), `suspicious/noControlCharactersInRegex` (2 — intentional U+0000/U+001F sanitization at `args.js:419`), `assist/source/organizeImports` (24).
- Convert the 3 unused catch bindings to optional catch binding (`catch {}`) rather than renaming them: zero behavior change, and the bundle diff should then show only the removals.
- Alternatives already walked: **oxlint** (23 diagnostics, 11 distinct and all stylistic, costs two pinned binaries and the exercised artifact was a Node wrapper plus an N-API addon, not a standalone executable); **ESLint** (its maintainers state `package.json` is effectively required even under flat config, so it cannot satisfy the no-npm shape at all); **quick-lint-js** (true single binary but narrowly scoped and pre-release by its own docs); **deno check** (satisfies the boundary as a single pinned binary but uses the same `checkJs` mechanism as tsc and inherits the same 343-implicit-any distribution — the blocker is missing JSDoc, not a missing runner); a **pre-commit local hook** instead of CI (the binary is not guaranteed present on a contributor machine, and CI's lint job installs only pre-commit).
- Restating the no-npm rule in `AGENTS.md` (requirement 12) is what makes this land as a clarification rather than a relaxation: the constraint as practiced has always governed the tree and the shipped runtime, and `validate.yml` already runs npm on the runner.

## Out of scope

- Enabling the Biome formatter. It would rewrite 32 of 32 files and requires first making `build.js`'s `strip()` matcher multi-line-aware — file that as its own follow-up.
- Any type checking (`tsc --checkJs`, `tsgo`, `deno check`) and any JSDoc annotation work. 453 errors, 76% bare implicit-any, against zero annotations in all 9 src files.
- `complexity/useOptionalChain` (51 sites) and `correctness/noUnusedFunctionParameters` (55 sites) cleanups, including any `_`-prefix rename pass over test mock callbacks.
- Adopting oxlint or tsgolint alongside Biome.
- Linting or formatting Python under `scripts/`, `tests/`, or `bench/` — #104 owns that axis.
- Adding a `package.json`, lockfile, or `node_modules` anywhere in the tree, or relaxing the no-dependency rule for the shipped pipeline runtime.
- Running the linter from a pre-commit hook.
- Changing `.github/workflows/validate.yml`'s existing `npm install -g @anthropic-ai/claude-code` step — #106 pins its version.
- Editing the ruleset itself beyond adding this job's context (requirement 13); every other ruleset rule belongs to #108.
- Consolidating duplicated logic inside `workflows/src/mergeFindings.js` or `stages.js` — #110 owns that, and its diff must stay lint-clean under whatever rule set lands here.

## Verification

`node --test workflows/test/*.test.js`, `node workflows/build.js && git diff --exit-code workflows/pipeline.js`, `python -m pytest tests/ -q`, `python -m pytest bench/tests/ -q`, and `pre-commit run --all-files` all pass; the new job exits 0 at HEAD; the restricted-globals rule is proven by mutation (add one banned reference under `workflows/src/`, observe the job fail, revert); and `git status --porcelain` is empty after a full run. Measurement tier: `suites-only`.

leehopper:
## Summary

- Add CI-only Biome 2.5.6 `js-lint` job (`Run Workflow JS Lint`) with SHA-256-pinned `biome-linux-x64` binary in `$RUNNER_TEMP` — no `package.json`, lockfile, or `node_modules`.
- Add `workflows/biome.json` (lint-only, formatter off, explicit ON rules including sandbox `noRestrictedGlobals`) and restate the no-npm / CI-binary boundary in root `AGENTS.md`; deferred-rule hit counts live in `workflows/AGENTS.md`.
- Clean unused catch bindings (`catch {}`) and one `Object.hasOwn` site so the gate is green; freeze the new check name in `REQUIRED_PR_CHECK_CONTEXTS`.

Closes #105 (Wave 2 of #101). Owner decision: option (ii) / Approach 1.

### Invocation note

CI and local verification use `working-directory: workflows` + `biome check --error-on-warnings .`. Biome 2.5.6 rejects `--config-path=workflows` from the repo root (nested-root detection). Comment in `ci.yml` records why.

### Mutation proof (req 7)

During implementation: added a `structuredClone` reference under `workflows/src/`, observed `noRestrictedGlobals` failure, reverted. Not retained as a standing CI self-test.

### Catch bindings / req 8

Live HEAD had four unused `catch (e)` sites (plus one `noPrototypeBuiltins` in tests), all fixed. On Biome 2.5.6, `noUnusedVariables` **does** flag unused catch bindings — they are gate-enforced, not fixed-but-unenforced.

## Owner follow-up (after this PR’s CI shows the new check, or right after merge)

- [ ] Add required status-check context `Run Workflow JS Lint` to ruleset `protect-default-branch` (id 16049246). Do **not** add it before the job exists on a run — open PRs would block on a check that never reports.

## Test plan

- [x] `(cd workflows && biome check --error-on-warnings .)` — 30 files, 0 parse errors, exit 0
- [x] `node workflows/build.js && git diff --exit-code workflows/pipeline.js`
- [x] JS tests with coverage floors + presence check
- [x] `python -m pytest tests/ -q` and `python -m pytest bench/tests/ -q`
- [x] `pre-commit run --all-files`
- [x] No npm artifacts in tree
- [ ] CI job `Run Workflow JS Lint` green on this PR
- [ ] Owner ruleset click (checkbox above)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -22,7 +22,9 @@ takes coordinated edits across N files, fix the shape rather than documenting th
 
 - No `package.json`, lockfile, or `node_modules` is tracked in the repository.
 - The shipped pipeline runtime (`workflows/src`, `workflows/pipeline.js`) has zero
-  npm dependencies — Node builtins and language globals only.
+  dependencies — language globals plus the host-injected `agent`/`parallel`/
+  `pipeline`/`args` only. No import survives into the bundle (`build.js` strips
+  import lines); see `workflows/AGENTS.md` for the sandbox surface.
 - Pinned static binaries in CI are permitted (e.g. Biome for the `js-lint` job).
   Bumps are manual and deliberate; do not add a package manifest to get Dependabot
   coverage for those pins.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,6 +18,17 @@ wrong — change the shape.
 **Extending should cost one edit.** A new dimension, field, or agent belongs in one place. If it
 takes coordinated edits across N files, fix the shape rather than documenting the ritual.
 
+## Tooling boundary (no npm in the tree)
+
+- No `package.json`, lockfile, or `node_modules` is tracked in the repository.
+- The shipped pipeline runtime (`workflows/src`, `workflows/pipeline.js`) has zero
+  npm dependencies — Node builtins and language globals only.
+- Pinned static binaries in CI are permitted (e.g. Biome for the `js-lint` job).
+  Bumps are manual and deliberate; do not add a package manifest to get Dependabot
+  coverage for those pins.
+- This does not forbid npm on the CI runner (`validate.yml` already installs the
+  Claude Code CLI globally).
+
 ## Scripts
 
 - **stdlib-only Python.** No pip dependencies in shipped `scripts/` runtime —
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,40 @@ jobs:
           COVERAGE_FILE="$(mktemp -d)/.coverage" python -m pytest tests/ -q \
             --cov=scripts --cov=.github --cov-fail-under=91.7
 
+  js-lint:
+    name: Run Workflow JS Lint
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Download and verify Biome
+        env:
+          # Biome pin is manual — Dependabot cannot see a raw binary URL.
+          # To bump: identify the new release asset id, then:
+          #   gh api repos/biomejs/biome/releases/assets/<id> --jq .digest
+          # Update BIOME_VERSION, URL, and BIOME_SHA256 together. Do not add
+          # package.json / lockfile to get auto-bumps; that reopens #105.
+          BIOME_VERSION: "2.5.6"
+          BIOME_SHA256: "3cc9a0c3fa26ac26a89e8a3b203c010c9ae88e36f69a2679e79981f267ce9d57"
+        run: |
+          set -euo pipefail
+          url="https://github.com/biomejs/biome/releases/download/@biomejs/biome@${BIOME_VERSION}/biome-linux-x64"
+          out="${RUNNER_TEMP}/biome"
+          curl -fsSL "$url" -o "$out"
+          echo "${BIOME_SHA256}  ${out}" | sha256sum -c -
+          chmod +x "$out"
+          echo "BIOME=${out}" >> "$GITHUB_ENV"
+
+      # cwd is workflows/ because --config-path from the repo root trips Biome 2.5.6's nested-root-configuration detection.
+      - name: Lint workflows JS
+        working-directory: workflows
+        run: |
+          set -euo pipefail
+          "${BIOME}" check --error-on-warnings .
+
   workflow-tests:
     name: Run Workflow JS Tests
     runs-on: ubuntu-latest
--- a/tests/test_agent_instruction_layout.py
+++ b/tests/test_agent_instruction_layout.py
@@ -80,7 +80,7 @@ CODEX_CAP_BYTES = 32_768
 # 3.12 CI reading of 92.69, per the floor policy in the same paragraph) — a floor whose
 # stated provenance no longer matches its value is exactly the unverifiable claim the
 # repo's own rules prohibit.
-AGENTS_SET_BUDGET_BYTES = 17_513
+AGENTS_SET_BUDGET_BYTES = 18_916
 CLAUDE_MD_MAX_BYTES = 856
 
 # Root CLAUDE.md is a pointer, not a document. The line cap is a shape bound and keeps its
@@ -397,6 +397,7 @@ class TestClaimsResolve(unittest.TestCase):
         "TextDecoder",
         "package.json",
         "node_modules",
+        "noControlCharactersInRegex",  # Biome rule id in deferred-rules table
     }
 
     def repo_files(self):
--- a/tests/test_agent_instruction_layout.py
+++ b/tests/test_agent_instruction_layout.py
@@ -80,6 +80,8 @@ CODEX_CAP_BYTES = 32_768
 # 3.12 CI reading of 92.69, per the floor policy in the same paragraph) — a floor whose
 # stated provenance no longer matches its value is exactly the unverifiable claim the
 # repo's own rules prohibit.
+# Raised 17_513 -> 18_916 (2026-08-04, #105): workflows/AGENTS.md JS lint
+# section (Biome pin, formatter-off rationale) and deferred-rules table.
 AGENTS_SET_BUDGET_BYTES = 18_916
 CLAUDE_MD_MAX_BYTES = 856
 
--- a/tests/test_agent_instruction_layout.py
+++ b/tests/test_agent_instruction_layout.py
@@ -82,7 +82,9 @@ CODEX_CAP_BYTES = 32_768
 # repo's own rules prohibit.
 # Raised 17_513 -> 18_916 (2026-08-04, #105): workflows/AGENTS.md JS lint
 # section (Biome pin, formatter-off rationale) and deferred-rules table.
-AGENTS_SET_BUDGET_BYTES = 18_916
+# Raised 18_916 -> 19_079 (2026-08-04, #105): reword tooling-boundary bullet —
+# shipped runtime is language globals + host-injected args only, not Node builtins.
+AGENTS_SET_BUDGET_BYTES = 19_079
 CLAUDE_MD_MAX_BYTES = 856
 
 # Root CLAUDE.md is a pointer, not a document. The line cap is a shape bound and keeps its
--- a/tests/test_contribution_surface.py
+++ b/tests/test_contribution_surface.py
@@ -925,6 +925,7 @@ REQUIRED_PR_CHECK_CONTEXTS = (
     "Run Tests (3.10)",
     "Run Tests (3.11)",
     "Run Tests (3.12)",
+    "Run Workflow JS Lint",
     "Run Workflow JS Tests",
     "Run Bench Self-Tests (3.11)",
     "Run Bench Self-Tests (3.12)",
--- a/workflows/AGENTS.md
+++ b/workflows/AGENTS.md
@@ -17,6 +17,26 @@ The v3 review pipeline. `pipeline.js` is invoked by the skill through the `Workf
 - **No wall-clock.** Timestamps arrive through the args waist (`generatedAt`); environment values
   are read by the skill and passed in `policy`.
 
+## JS lint (CI)
+
+CI job `js-lint` runs Biome 2.5.6 against this directory via `workflows/biome.json`
+(lint-only). Formatter is **off**: default `lineWidth: 80` would wrap long
+`import … from` lines and break `build.js`'s single-line import `strip()` regex,
+failing `tests/test_bundle_fresh.py`. `noRestrictedGlobals` applies to `src/`
+only; `test/**` and `build.js` override it (they legitimately use `process` /
+`console`).
+
+Deferred rules (measured 2026-07-30, HEAD `ebf399d`) — hit counts are why they
+stay off so a later re-evaluation need not re-derive them:
+
+| Rule | Hits | Why deferred |
+| --- | --- | --- |
+| `useOptionalChain` | 51 | Style churn; no reliability delta |
+| `noUnusedFunctionParameters` | 55 | Dominated by positional mock callback params in tests |
+| `noAssignInExpressions` | 7 | Six are intentional `args.js` pattern |
+| `noControlCharactersInRegex` | 2 | Intentional U+0000/U+001F sanitization in `args.js` |
+| `organizeImports` | 24 | Assist/format-adjacent; out of scope with formatter off |
+
 ## The bundle
 
 - `pipeline.js` is **generated** — never hand-edit it. Source is `src/*.js`, which may use ESM
--- a/workflows/CLAUDE.md
+++ b/workflows/CLAUDE.md
@@ -21,6 +21,26 @@ The v3 review pipeline. `pipeline.js` is invoked by the skill through the `Workf
 - **No wall-clock.** Timestamps arrive through the args waist (`generatedAt`); environment values
   are read by the skill and passed in `policy`.
 
+## JS lint (CI)
+
+CI job `js-lint` runs Biome 2.5.6 against this directory via `workflows/biome.json`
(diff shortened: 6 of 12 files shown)


AGENTS.md
```diff
@@ -18,6 +18,17 @@ wrong — change the shape.
 **Extending should cost one edit.** A new dimension, field, or agent belongs in one place. If it
 takes coordinated edits across N files, fix the shape rather than documenting the ritual.
 
+## Tooling boundary (no npm in the tree)
+
+- No `package.json`, lockfile, or `node_modules` is tracked in the repository.
+- The shipped pipeline runtime (`workflows/src`, `workflows/pipeline.js`) has zero
+  npm dependencies — Node builtins and language globals only.
```
leehopper: **Must fix — "Node builtins" is false for the shipped runtime, and the phrasing licenses the exact failure this PR closes.**

`workflows/src/*.js` imports zero Node builtins today (every import is relative `./`; only the unshipped `build.js` uses `node:*`), and the workflow sandbox provides none of them — that's the documented premise of the `noRestrictedGlobals` deny-list this PR adds. As written, this bullet tells a future author that Node builtins are fair game in `src/`. If they act on it (`import { inspect } from 'node:util'`), lint passes (the binding is declared in the src file), `build.js` strips the import line during bundling, and the bundle ships an undefined reference that throws only on a live dispatch of that path — the coverage-dependent live-crash class #105 exists to eliminate.

Suggested replacement for this bullet:

```markdown
- The shipped pipeline runtime (`workflows/src`, `workflows/pipeline.js`) has zero
  dependencies — language globals plus the host-injected `agent`/`parallel`/
  `pipeline`/`args` only. No import survives into the bundle (`build.js` strips
  import lines); see `workflows/AGENTS.md` for the sandbox surface.
```

This edit changes byte counts, so in the same commit: re-run `python3 scripts/sync_agent_rules.py`, and re-pin `AGENTS_SET_BUDGET_BYTES` in `tests/test_agent_instruction_layout.py` to the new exact measured total (the ratchet is pinned AT measured size — current branch total is 18,916; recompute after the edit and update the raise comment's value).

The structural guard for the same gap (a `build.js` assertion that every stripped import specifier starts with `./`) is filed separately — out of scope here.

leehopper:
Advisor review (independent verification, not just a read-through). **One must-fix, everything else verified correct.**

**Verified on this branch:**

- CI fully green including `Run Workflow JS Lint` (8s).
- The pinned SHA-256 matches GitHub's asset-digest API for `biome-linux-x64` at `@biomejs/biome@2.5.6` — independently re-fetched.
- `AGENTS_SET_BUDGET_BYTES = 18_916` equals the exact measured total of the four `AGENTS.md` files (7183 + 3106 + 2782 + 5845); twins report current under `sync_agent_rules.py`.
- The `pipeline.js` diff contains exactly the four `catch {}` conversions from `src/` and nothing else; the `parity.test.js` `Object.hasOwn` change is correctly absent from the bundle.
- All 13 requirements of #105 trace: full 11-name deny list, `test/**` + `build.js` override, both exclusions empirically load-bearing (30 files, 0 parse errors), frozen tuple updated in-PR, deferred-rules hit-count table in `workflows/AGENTS.md`, owner ruleset click correctly sequenced as post-CI follow-up.

**Must fix (inline comment on `AGENTS.md` lines 24–25):** the "Node builtins and language globals only" clause in the tooling-boundary restatement is false for the shipped runtime and licenses the exact coverage-dependent live-crash class this PR closes. Replacement text in the inline comment; the edit requires re-pinning `AGENTS_SET_BUDGET_BYTES` to the new measured total and re-running `sync_agent_rules.py` in the same commit.

**Filed as follow-up, out of scope here:** #137 — `build.js` assertion that every stripped import specifier is relative, turning the reworded prose into a mechanism.

**Noted, no action:** requirement 3 (tree clean after run) is verified manually rather than asserted in the job — acceptable with formatter and assist off; `actions/checkout@v4` tag-pinning is correctly left to #106; the audit-count drift (5 vars + 3 catch → 4 catch + 1 `noPrototypeBuiltins`) is honestly documented in the PR body and the green gate at HEAD is the arbiter.

AGENTS.md
leehopper: Fixed in the follow-up commit: reworded the tooling-boundary bullet to language globals + host-injected `agent`/`parallel`/`pipeline`/`args` only (no Node builtins), re-ran `sync_agent_rules.py` (twins current), and re-pinned `AGENTS_SET_BUDGET_BYTES` to the new measured total 19_079 with a raise comment.

Issue #148, opened 2026-08-07
Persist could enforce outputDir-prefix on planned artifact paths


leehopper:
## Context

Deferred from #86 (Approach 3). That issue fixed containment at Phase 1 (`ensure_output_dir.py`) and the args waist (absolute `outputDir`). Pipeline Persist already builds paths via `plannedArtifactPaths(outputDir, sha)` — every planned path is already prefixed by `outputDir` today.

## Proposal

Optionally harden Persist / `writeArtifacts` so any planned or writer path that is not under the absolute `outputDir` fails closed (gap or hard reject), as a belt-and-braces fence if a future stamp regresses.

## Why deferred

#86's delivery leak was skill-side (root markdown write), not Persist. Suites-only; no smoke required if implemented.

## Out of scope

- `repoRoot` absolute validation / removal (#81)
- Changing the default output directory name

leehopper:
## Design for #148 (Wave 2) — approved §1–§2

Session scratch only — not committed to the tree (AGENTS.md: design memos stay in the issue/PR, not tracked docs).

### Goal

Belt-and-braces Persist fence: planned artifact paths that escape absolute `outputDir` hard-fail as a programming error; writer-echo / off-plan paths that escape degrade with a named `path-escape` gap. Suites-only; no smoke.

### Locked decisions

| Topic | Decision |
| --- | --- |
| Failure split (Q1) | Hard reject for stamp/invariant; named gap for writer-echo / off-plan |
| Hard-reject placement (Q2) | Helper throws + stamp before `writeArtifacts` outer `try` (same predicate, no duplicate) |
| Under-predicate (Q3) | Normalize (`/.` collapse only); reject `\`, any `..` segment; absolute root required; trailing-slash strip; under iff `path === root \|\| path.startsWith(root + '/')` |
| `//` | Deliberately **not** collapsed (still under fence; comment in code) |
| Relative fallback | Delete `\|\| '.code-gauntlet'`; missing/relative `outputDir` → hard reject |
| Approach | **1** — single `pathUnderOutputDir` + stamp-at-entry |
| Python twin | Out of scope (`materialize` already `_confined` via realpath) |
| `#81` repoRoot | Out of scope |

### Ownership

| Piece | Owns |
| --- | --- |
[...]

Issue #81, opened 2026-07-29
repoRoot is stamped into the args waist but read by nothing: agent 'read the project's CLAUDE.md' instructions have no path anchor


leehopper:
## Problem / Goal

Several agent contracts instruct the agent to open a file by a **repo-relative name**, but nothing in the dispatch path ever tells the agent what the repo root is.

`workflows/src/args.js:11-14` says so outright:

> "`mode` and `repoRoot` are NOT read anywhere in workflows/src (mode is only ever re-checked against its own enum below; **repoRoot is unread entirely**) — they're required as provenance/telemetry the skill always stamps, not because any stage consumes them."

The only path-shaped datum any discovery/validate prompt carries is the shared context file path, built by `sharedContextLine` (`workflows/src/stages.js:104-107`) from `inp.contextPath`. Meanwhile:

- `agents/bug-detector.md:27` — "Read the project's CLAUDE.md first (if it exists)."
- `agents/code-simplifier.md:15,28,183` — "Read the project's CLAUDE.md first." / "…CLAUDE.md files before suggesting changes." / "Use Read to load CLAUDE.md…"
- `agents/security-reviewer.md` — "Use Read, Grep, and Glob to explore the full repository, not just the changed files."

Whether those resolve correctly depends entirely on the ambient working directory a subagent dispatched through the workflow runtime's `agent()` global inherits — which is **not** established anywhere in this repo's source or tests, and is a different mechanism from a top-level Task-tool subagent.

## Why it matters

This is the load-bearing assumption under several dimensions, and nothing verifies it:

- If the anchor holds, these instructions work and the gap is only that it is undocumented and untested.
- If it does not hold, then `security-reviewer`'s and `cross-file-impact`'s "explore the full repository" instruction is quietly a no-op in some configurations, and their cross-file findings are weaker than the contracts claim.

Either way the current state is that a behaviour the pipeline depends on is unverified. Note this is **not** hypothetical scaremongering: #49 found a directly analogous case where an instruction everyone assumed worked (reading a repo's CLAUDE.md) returned an 11-byte pointer on 3 of 5 real repos for years, silently.

PR #79 reduces the blast radius slightly — `conventions-and-intent`'s self-read of AGENTS.md/QODO.md is retired in favour of pre-loaded context — but leaves every CLAUDE.md self-read and every "explore the repository" instruction untouched.

## Scope of the question (deliberately open)

Establish the fact before designing anything:

[...]

PR #149, opened 2026-08-07
fix: contain review artifacts in a single ignored output dir (#86)

### LINKED ISSUE #86: Review artifacts are not reliably contained in a single gitignored output folder
## Problem / Goal

A review run is supposed to leave the reviewed repo's working tree clean aside from one gitignored output directory (`.code-gauntlet/` by default, overridable with `$CODE_GAUNTLET_OUTPUT_DIR`). Operators report that runs instead create a lot of visible files in the repo under review — the "single folder you can gitignore" contract feels broken.

Investigation of the current skill/pipeline shows the *intent* is still that design, but the containment + ignore guarantee is soft (model-executed path interpolation and a best-effort `info/exclude` append) and has documented holes that write outside the folder. Fixed means: every review-created byte lands under the resolved absolute output directory, that directory is ignored without editing the tracked `.gitignore`, and a run that cannot honor the contract discloses the failure instead of silently littering `git status`.

## Requirements

1. After a successful interactive or headless review against a clean clone that does not already ignore the output dir, `git status --porcelain` shows **no** untracked or modified paths for review artifacts — every artifact path is under the resolved output directory and is ignored (via existing ignore rules or `.git/info/exclude`, never via an edit to the tracked `.gitignore`).
2. The Phase 1 / Phase 2 path that creates the output directory resolves it to an **absolute** path under the reviewed repo (`git rev-parse --show-toplevel`) before any artifact is written, and stamps that same absolute value as `args.outputDir` / every `{output_dir}/…` write site. A relative `.code-gauntlet` echo alone is not enough.
3. The gitignore step keys off the **resolved** output directory (not a hardcoded `.code-gauntlet` literal). When `$CODE_GAUNTLET_OUTPUT_DIR` overrides the default, the same ignore guarantee still applies to that override path (today phase2-triage.md says "skip if using env var override", which leaves custom dirs fully visible).
4. Markdown delivery (`./code-gauntlet-{date}.md` in `phase8-delivery.md` / `delivery-guide.md`) either moves under the output directory or is an explicit, disclosed user-chosen write — it must not silently contradict PRIVACY.md's claim that "the only files created are within the `.code-gauntlet/` directory".
5. A regression pin covers the contract: at minimum, (a) SKILL.md's gitignore section still uses `.git/info/exclude` and never `>> .gitignore` (already partially pinned), and (b) a new pin that every skill/workflow write site for review artifacts is prefixed by the output-dir token / `outputDir` field — root-level `code-gauntlet-*` writes are forbidden except any explicitly allowed delivery path from requirement 4.

## Evidence

Taken 2026-07-29 against `main` at HEAD `3806f44002ee974b8d58d0be70dd033db99c8d3a` (detached checkout in the investigation environment).

**Design still says "one folder":**

- `PRIVACY.md` Data Storage: "The only files created are within the `.code-gauntlet/` directory… The directory is gitignored by default."
- `CLAUDE.md` Output directory convention: `{output_dir}` defaults to `.code-gauntlet/` (repo-local, gitignored).
- `skills/code-gauntlet/references/phase2-triage.md` Args table: `outputDir` = "resolved `{output_dir}` (absolute)".

**Ignore mechanism (SKILL.md Phase 2 Composite A `gitignore` section) — hardcoded default only:**

```bash
if git check-ignore -q .code-gauntlet 2>/dev/null; then
  echo "already-ignored"
elif echo "/.code-gauntlet/" >> "$(git rev-parse --git-common-dir)/info/exclude"; then
  echo "added"
else
  echo "unwritable"
fi
```

`phase2-triage.md` §2b-post step 2 explicitly says to skip this when using the env-var override — so `$CODE_GAUNTLET_OUTPUT_DIR=/tmp/foo` (or any custom path) gets no ignore entry by design.

**Phase 1 still emits a relative path — absolute resolution is prose-only:**

```bash
OUTPUT_DIR="${CODE_GAUNTLET_OUTPUT_DIR:-.code-gauntlet}"
echo "$OUTPUT_DIR"
mkdir -p "$OUTPUT_DIR" && echo "mkdir: ok" || echo "mkdir: FAILED"
```

There is no `realpath` / `git rev-parse --show-toplevel` join in that composite. `workflows/src/args.js` type-checks `outputDir` as a non-empty string but does **not** require it to be absolute (comment claims the skill stamps an absolute value).

**Documented write outside the folder (contradicts PRIVACY.md):**

- `skills/code-gauntlet/references/phase8-delivery.md` Step C: "write to `./code-gauntlet-{date}.md`".
- `skills/code-gauntlet/references/delivery-guide.md`: "Write the full report to `./code-gauntlet-<date>.md` or a user-specified path."

**Root-level `code-gauntlet-*` is not covered by the exclude pattern** (sandbox reproduction, same date):

```
$ git init && mkdir -p .code-gauntlet && echo x > .code-gauntlet/foo.json
$ echo '/.code-gauntlet/' >> "$(git rev-parse --git-common-dir)/info/exclude"
$ git check-ignore -v .code-gauntlet/foo.json
# → .git/info/exclude:…:/.code-gauntlet/  .code-gauntlet/foo.json   (ignored)
$ echo y > code-gauntlet-findings-abc12345.json
$ git check-ignore -v code-gauntlet-findings-abc12345.json || echo NOT IGNORED
# → NOT IGNORED
$ git status -u --porcelain
# → ?? code-gauntlet-findings-abc12345.json
```

So any skill/writer path that drops the `{output_dir}/` prefix (or stamps `outputDir` as `.` / empty-with-fallback quirks) lands files that the exclude entry can never hide. A single review already plans many paths under the folder (context, diff, files, project-rules, findings, report, post-review, checkpoint, persist-plan, plus `phase4-input/output-*.sliceN.json` per verify slice) — when containment or ignore fails, `git status` lights up with a pile of files.

**Existing issues checked (2026-07-29) — none owns this:**

| Issue | Why it is not this |
|---|---|
| #53 / #69 | Artifact-writer *fidelity* / parse integrity, not placement or ignore |
| #50 | Persistence floor / double-typing, not working-tree hygiene |
| #81 | `repoRoot` unused for agent *reads* — write-side containment out of scope |
| #65 | Stale NDJSON emission docs vs by-value agents |
| #37 | v2→v3 residue audit; its triage rule says behavior/design gaps become their own issues |

Title + body search across open/closed issues for gitignore / output-dir / working-tree pollution / `.code-gauntlet` containment found no owner.

## Candidate directions (non-binding, from investigation)

1. **Harden Phase 1 composite:** resolve `OUTPUT_DIR` to `$(git rev-parse --show-toplevel)/.code-gauntlet` (or `realpath` of the env override), `mkdir -p` that absolute path, echo the absolute path, and use it everywhere `{output_dir}` is substituted — including the gitignore section's `check-ignore` target and exclude pattern.
2. **Generalize gitignore:** append a pattern derived from the resolved output dir (repo-relative form anchored with `/`), including when `$CODE_GAUNTLET_OUTPUT_DIR` is set *and* that path is inside the repo; if the override is outside the repo, disclose "artifacts are outside the working tree" and skip exclude.
3. **Move markdown delivery under `{output_dir}/`** (e.g. `{output_dir}/code-gauntlet-{date}.md`) and update PRIVACY.md / delivery docs to one story; or keep root delivery but only when the user picked "Markdown file" and disclose it as a deliberate non-ignored write.
4. **Pin with tests:** extend the existing `test_phase2_triage_never_appends_to_tracked_gitignore` style pins to (a) assert the gitignore section references `{output_dir}` / `$OUTPUT_DIR` rather than only the `.code-gauntlet` literal, and (b) forbid root-level `./code-gauntlet-*.md` delivery paths unless requirement 4 explicitly keeps them.

## Out of scope

- Artifact-writer content fidelity / checksum drift (#53, #69, #25).
- `repoRoot` anchoring for discovery-agent *reads* (#81) — related path hygiene, different surface; do not merge the issues.
- Changing the default directory name away from `.code-gauntlet/`, or adding a tracked `.gitignore` entry to *reviewed* repos.
- Bench harness output under `bench/` / `$CODE_GAUNTLET_OUTPUT_DIR` for children (already pointed at `{run_dir}/output`).

## Verification

Measurement tier: `suites-only`.

- `python -m pytest tests/ -q` (include any new containment / gitignore pins).
- `node --test workflows/test/*.test.js` if waist/path validation changes.
- Manual smoke (not a gate): run `/code-gauntlet` against a throwaway clone with a clean `git status`, then confirm `git status --porcelain` has no review-artifact paths and `git check-ignore -v <outputDir>/…` reports the exclude/ignore rule.

---
_Created via Liatrio Brain in Slack_

leehopper:
## Summary
- Add `scripts/ensure_output_dir.py` to resolve an absolute output dir, establish ignore via `git rev-parse --git-path info/exclude` (verify-after-append), and mkdir only after the gate passes — hard-stop when in-repo containment cannot be established
- Args waist rejects non-absolute `outputDir` (POSIX `/`-prefix); Phase 1 invokes the script; Composite A gitignore bash removed
- Markdown delivery (interactive + headless) surfaces `artifactPaths.report` — no default root `code-gauntlet-{date}.md` write; user-named path is the only outside-fence write

## Design decisions (from #86 thread)
- Approach 2 (mechanism script, not prose-pinned bash)
- Refined-A markdown: path-surface only (no date alias)
- In-repo disclose-and-continue intentionally unavailable; escape hatch is `$CODE_GAUNTLET_OUTPUT_DIR` outside the repo
- `repoRoot` absolute check deferred to #81
- Follow-up filed: Persist `outputDir`-prefix fence (#148)

## Lee veto window
Root `./code-gauntlet-{date}.md` as a “review is done” signal is replaced by the chat path to `artifactPaths.report`. Design assumes that is preferred (you filed #86; PRIVACY.md already claimed one folder). Veto on the issue/PR if you relied on the root file.

## Test plan
- [x] `tests/test_ensure_output_dir.py` (14) + mechanism mutations
- [x] Waist relative/absolute `outputDir` tests; JS suite green; bundle rebuild idempotent
- [x] DocContract pins + mutations (root write / gitignore section / script name)
- [x] CI-identical coverage: Python 92.30 (≥91.7), JS 98.63/83.01/98.00, bench 88.45 (≥87)
- [ ] Suites-only CI on this PR

Closes #86


leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: d545e523eb0f4438294a550bb6f378285c3ea64a

scripts/ensure_output_dir.py
```diff
@@ -0,0 +1,269 @@
...
+        exclude_path = git_exclude_path(cwd)
+        if exclude_path is None:
+            return (
+                1,
+                "",
+                f"cannot establish gitignore for {abs_dir} "
+                "(info/exclude unresolvable via `git rev-parse --git-path`, "
+                "not otherwise ignored); "
+                "set CODE_GAUNTLET_OUTPUT_DIR to a path outside the repo and re-run.\n",
+            )
```
leehopper, line 217: **🟡 [MEDIUM] git_exclude_path() returning None (info/exclude unresolvable) is never exercised**

When check-ignore reports not-ignored, run() calls git_exclude_path(cwd) and hard-stops with exit 1 if it returns None (git rev-parse --git-path info/exclude fails or returns empty). tests/test_ensure_output_dir.py only tests the sibling 'exclude file unwritable' branch (exclude_writable() False) by chmod-locking a real, resolved exclude path; it never mocks or otherwise triggers git_exclude_path itself returning None.

**Suggested fix:**
Add a test that patches eod.git_exclude_path to return None (mirroring the existing unittest.mock.patch.object pattern used for git_check_ignore and append_exclude_pattern) and asserts exit 1, empty stdout, and the 'info/exclude unresolvable' message in stderr; also assert no mkdir occurred.

scripts/ensure_output_dir.py
```diff
@@ -0,0 +1,269 @@
...
+                f"({DEFAULT_OUTPUT_DIR})."
+            )
+            return 2, "", msg + "\n"
+        raw = raw.strip()
+    else:
+        raw = DEFAULT_OUTPUT_DIR
+
+    repo_root = git_repo_root(cwd)
+    if repo_root is None:
+        return 2, "", "not a git repository (git rev-parse --show-toplevel failed)\n"
```
leehopper, line 177: **🟡 [MEDIUM] No test exercises the 'not a git repository' exit-2 path**

run() returns exit 2 with a specific stderr message when git_repo_root(cwd) returns None (git rev-parse --show-toplevel fails), a case documented in the script's own docstring and in the SKILL.md hard-stop table. Every test in tests/test_ensure_output_dir.py calls setUp's _init_repo(), so all 14 tests operate inside a real git repo; none constructs a plain non-git directory to exercise this branch.

**Suggested fix:**
Add a test that runs eod.run(argv=['--cwd', <plain tempdir, no git init>]) and asserts exit code 2, empty stdout, and 'not a git repository' in stderr.

**Cited rule:** AGENTS.md: 'A regression test must fail against the bug it names' — this documented exit path has no regression test at all to fail.

scripts/ensure_output_dir.py
```diff
@@ -0,0 +1,269 @@
...
+        stderr_parts.append(f"exclude: added {pattern} via {exclude_path}")
+    else:
+        stderr_parts.append(f"exclude: already-ignored {abs_dir}")
+
+    try:
+        os.makedirs(abs_dir, exist_ok=True)
+    except OSError as exc:
+        return 1, "", f"mkdir failed for {abs_dir}: {exc}\n"
+
+    return 0, abs_dir + "\n", "".join(line + "\n" for line in stderr_parts)
```
leehopper: **🟡 [MEDIUM] stderr_parts list only ever holds one element before being joined**

`stderr_parts: list[str] = []` is declared to collect disclosure lines, but tracing every reachable path shows exactly one `.append()` call fires before either of the two final `"".join(line + "\n" for line in stderr_parts)` calls — out-of-repo skip appends once then returns immediately, and the in-repo branch's if/else each append exactly once before the shared tail. The list-plus-join pattern implies a multi-line case that the code never actually produces, forcing a reader to verify there is no second append rather than seeing a single string built directly.

**Suggested fix:**
Replace `stderr_parts: list[str] = []` plus `.append(...)` calls with a single `disclosure = f"...\n"` string assignment at each of the three sites, and change the two return statements from `"".join(line + "\n" for line in stderr_parts)` to plain `disclosure`. For example the out-of-repo branch becomes `disclosure = f"outside-repo: skip exclude for {abs_dir} (artifacts are outside the working tree)\n"` ... `return 0, abs_dir + "\n", disclosure`.

scripts/ensure_output_dir.py
leehopper: Addressed in 97a9509: added `test_git_exclude_path_none_hard_stops` covering the `git_exclude_path` → None hard-stop path.

scripts/ensure_output_dir.py
leehopper: Addressed in 97a9509: added `test_not_a_git_repository_exits_2` for the non-git exit-2 path.

scripts/ensure_output_dir.py
leehopper: Addressed in 97a9509: simplified to a single disclosure string instead of a one-element `stderr_parts` list.

tests/test_ensure_output_dir.py
```diff
@@ -0,0 +1,283 @@
...
+        _init_repo(self.repo)
+
+    def tearDown(self) -> None:
+        # Un-chmod anything we locked so cleanup can remove it.
+        for root, dirs, files in os.walk(self.repo):
+            for name in dirs + files:
+                p = os.path.join(root, name)
+                try:
+                    os.chmod(p, stat.S_IRWXU)
+                except OSError:
```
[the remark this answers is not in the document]
leehopper: Addressed in 97a9509: teardown now uses a ruff-clean `except OSError: pass` with an explanatory comment that cleanup errors are intentionally ignored.

PR #150, opened 2026-08-07
feat: persist outputDir-prefix fence for planned and writer paths (#148)

### LINKED ISSUE #148: Persist could enforce outputDir-prefix on planned artifact paths

leehopper:
## Summary
- Add a single `pathUnderOutputDir` predicate (normalize `/./`, reject `\`/`..`, absolute root, equality + prefix) used by stamp helpers and writer-echo checks
- `plannedArtifactPaths` / `persistPlanPath` / missing-relative `outputDir` hard-reject before `writeArtifacts`' outer try → `runWith` `{ ok: false }` (programming-error class)
- Writer-reported paths outside the fence degrade with one named `path-escape` gap (all-or-nothing `partial()`); token registered in `docs/machine-parsed-strings.md` (G3 still keys off `partial-artifacts`)

## Externally visible change
Omitting `outputDir` or passing a relative value no longer silently writes under a repo-relative default — it fails the run with a stable absolute-root message.

## Test plan
- [x] `node --test workflows/test/*.test.js` (626 pass)
- [x] `python3 -m pytest tests/test_machine_parsed_strings.py -q`
- [x] `node workflows/build.js && git diff --exit-code workflows/pipeline.js`
- [x] Mutation: relative stamp join → red; drop `..` rejection → red
- [ ] CI suites-only green

Closes #148


workflows/src/stages.js
```diff
@@ -2102,6 +2228,16 @@ async function attemptDerivedPersist(c, entries, planPath, plan, paths, policy,
...
+  // Issue #148: every receipt-reported path field (verified + written).
+  if (receipt && typeof receipt === 'object') {
+    const receiptFields = [];
+    const verified = Array.isArray(receipt.verified) ? receipt.verified : [];
+    const receiptWritten = Array.isArray(receipt.written) ? receipt.written : [];
+    verified.forEach((e, i) => receiptFields.push({ name: `verified[${i}]`, path: e && e.path, present: true }));
+    receiptWritten.forEach((e, i) => receiptFields.push({ name: `receipt.written[${i}]`, path: e && e.path, present: true }));
+    const receiptEscape = pathEscapeReason(outputDir, receiptFields);
+    if (receiptEscape) return fail(receiptEscape);
+  }
```
leehopper, line 2240: **🟡 [MEDIUM] Derived-persist path-fence checks (writtenEscape, receiptEscape) in attemptDerivedPersist have zero test coverage**

Issue #148 added two new outputDir-escape checks inside attemptDerivedPersist: one over the writer's echoed written[] list (writtenEscape) and one over the assemble receipt's verified[]/written[] path fields (receiptEscape). Neither is exercised by any test — the new stages_path_fence.test.js only drives the legacy writer path (writeArtifactsLegacy via ctx.agent), and the existing stages_persist.test.js derived-path tests (which do use an absolute OUT_DIR and assembleScriptPath) never supply an escaping path in written[] or in the receipt's verified/written arrays.

**Suggested fix:**
Add tests in stages_persist.test.js or stages_path_fence.test.js that drive the derived path (persist.assembleScriptPath set) and: (1) have the writer echo a written[] entry pointing outside outputDir (e.g. '/tmp/evil.json') and assert the result is { ok:false, reason: path-escape... } surfaced as a gap containing PATH_ESCAPE_TOKEN, and (2) have the assemble receipt's verified[] or written[] contain an out-of-fence path and assert the same escape-gap outcome rather than falling through to trustAssembleReceipt.

**Cited rule:** workflows/CLAUDE.md Persistence section: 'Both writer paths stay live and are not dead code: the derived one is the automatic fallback when the primaries exceed RETURN_CHAR_BUDGET' — the derived path is a live production path, not a legacy fallback only, so its new security-relevant fence logic warrants direct coverage.

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: df484beda9726f79ebe75a0bf22e047e90e90355

workflows/src/stages.js
leehopper: Fixed in 8e8da38. Added direct derived-path tests for escaping writer `written[]` entries and escaping assemble-receipt `verified[]`/`written[]` entries. Each test pins the single `path-escape` gap, and mutating either guard made its regression test fail.

workflows/pipeline.js
```diff
@@ -4279,14 +4382,18 @@ async function writeArtifacts(ctx, input) {
...
-    const outputDir = inp.outputDir || '.code-gauntlet';
-    const sha = inp.headShaShort || 'head';
-    const policy = inp.policy || {};
-    const paths = plannedArtifactPaths(outputDir, sha);
+  // Issue #148: absolute outputDir + planned-path stamp BEFORE the outer try so a
+  // programming-error / stamp regression throws to runWith ({ ok:false }) instead of
+  // being softened into a partial-artifacts gap. No relative fallback — omitting or
+  // passing a relative outputDir is the same class of failure as a stamp escape.
+  const c = ctx || defaultCtx();
+  const inp = typeof input === 'string' ? JSON.parse(input) : (input || {});
```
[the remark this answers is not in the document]
leehopper: Retaining `(input || {})` intentionally. `writeArtifacts(undefined)` is a supported directly-tested boundary for the new missing-outputDir behavior: the fallback allows `requireAbsoluteOutputDir(inp.outputDir)` to emit the stable `outputDir must be an absolute confined root` error. Removing it would instead throw a native `Cannot read properties of undefined` error, so the suggested change is not behavior-preserving. `pipeline.js` is generated from `workflows/src/stages.js`; no generated-only edit is appropriate.

PR #168, opened 2026-08-10
fix(post-review): gate GitLab position payloads against parser ground truth (#131)

### LINKED ISSUE #131: --dry-run does not validate GitLab position shape — malformed payloads pass dry-run then fail live with HTTP 400

leehopper:
## Summary

- Adds a `validate_position` gate in `scripts/post_review.py` that checks every GitLab inline-discussion position payload against the shape the parser's own ground truth (`valid_lines`, `new_files`, `old_paths`) implies, before it is sent (or written, in `--dry-run`).
- Both delivery paths run through the same unconditional gate: a malformed position is rejected in dry-run and in live delivery, with distinct counters and exit semantics for each (see Design).

## Design

- **SIGNATURE DEVIATION, flagged for owner sign-off.** The issue's literal ask was a single-arg `validate_position(position)`. That signature provably cannot catch the #127 D1 class: an added line's position dict and a D1-regressed context line's position dict are byte-identical — nothing distinguishes them from the dict alone. `validate_position` instead re-derives the expected shape from the parser's own `valid_lines`/`new_files`/`old_paths`, and diffs the actual position against that ground truth, rather than trusting the dict on its own.
- **Honest scope.** This gate catches D1-class assembly regressions (position shape drifting from what the parser established) and malformed inputs (non-integer line numbers; empty SHAs, caught by a hoisted pre-loop check). It does **not** catch D2 — D2 was a defect in the parser itself, and this gate consumes the parser's output as ground truth, so a parser that produces wrong ground truth will produce a gate that agrees with it. The issue's original claim is corrected here rather than silently repeated.
- **Shared gate, split semantics.** Dry-run and live delivery share one unconditional gate by construction — there is no code path that skips validation in either mode. Violations increment their own counter and drive exit-1 semantics per mode, while the dry-run payload is still written to disk (the write happens after `post_gitlab` returns, so a dry-run's output file reflects what was attempted even when the run subsequently reports failure).
- **Zero false positives** measured against every shipped GitLab fixture (added lines, context lines, renames, and directory-only diffs).
- **Merge order:** land after the #132 PR. Both PRs touch `post_gitlab`'s delivery loop; this is the smaller of the two, so it pays the rebase cost and re-runs its mutation catalogue after rebasing onto #132's changes.
- Closes #131.

## Mutation proofs

| mutation | expected red test | result |
| --- | --- | --- |
| M1: remove the `validate_position` call site | spy test + malformed-position e2e (dry-run and live) | RED — 3 failed as predicted |
| M2: revert the D1 guard (conditional `old_line` attach) | context-line contract tests | RED — cascade failure across the GitLab contract suites (verifier measured a 20-test blast radius, wider than the implementer's 12) |
| M3: weaken `old_path` to presence-only | `test_old_path_carries_the_post_rename_path` | RED — exactly that test, 1 failed / 197 passed |
| M4: remove the hoisted SHA check | `test_empty_sha_dies_before_the_summary_note` | RED — exactly that test, 1 failed / 197 passed |
| M5: route `invalid` into `skipped` | exit-semantics tests (dry-run + live malformed) | RED — both failed; `test_zero_posted_with_no_rejections_exits_zero` stayed green |

Gates at verification: suites green, coverage green (91.7 floor), pre-commit 14/14. Each mutation reverted and the suite re-greened before the next.

## Review

**Fixed (both should-fix findings + the two nits that were trivial):**

1. **`validate_position` now gates the whole position shape** (`scripts/post_review.py:488`). It builds the full expected position — `position_type`, all three SHAs, `new_path`, `new_line`, plus the conditional `old_line`/`old_path` — and diffs it against the actual in three directions: missing key, unexpected key, wrong value. The `line_code` special case is subsumed by the unexpected-key rule (so `line_range` and any future sibling are caught for free), and extending the position is now one edit. The `new_line` bool/float type check is retained separately, since `True`/`61.0` compare equal to the integer and would slip through a pure equality diff. The `fetch_gitlab_shas` triple is passed in as `shas`: whether the fetched values are *usable* stays answered once at the fetch (the hoisted check is untouched), but whether each position *carries* them is per-position structure only the gate can see.
2. **Live exit semantics** (`scripts/post_review.py`, end of `post_gitlab`). A malformed position no longer exits non-zero after a partially successful live delivery — inline discussions have no idempotency key, so that invited a rerun that double-posts everything that landed. It now follows the rule the GitLab-rejection branch already used: live exits non-zero only when nothing was posted inline (with a `die()` message mirroring the all-rejected one); `--dry-run` still fails on *any* malformed position.

https://claude.ai/code/session_015GdQDn1WVh7VX9hgAjHcMS

---

> [!NOTE]
> **Medium Risk**
> Touches GitLab MR comment delivery and exit codes; mistakes could skip valid findings or encourage risky reruns, but behavior is heavily tested and scoped to position assembly vs parser output.
> 
> **Overview**
> Adds **`validate_position`** in `post_review.py`, run on every GitLab inline discussion **before** capture or API post. Expected position shape is **re-derived** from the same diff facts as assembly (`valid_lines`, `new_files`, `old_paths`, MR SHAs)—missing/extra keys, wrong values, and non-integer `new_line` (e.g. `61.0`, `True`) are rejected.
> 
> **Delivery behavior:** Empty MR version SHAs fail **once** before the summary note. Malformed positions use a dedicated **`invalid`** counter (not “skipped” or GitLab rejections). **`--dry-run`** exits non-zero if **any** position is invalid but still writes `post-review-payload.json`; **live** exits non-zero only when **no** new inline discussions posted (partial success with warnings stays exit 0). `main()` propagates poster return codes after the dry-run file is written.
> 
> Docs: duplication register row for assembly↔gate mirror; delivery guide notes GitLab dry-run exit semantics. Tests cover unit branches, spy on the gate, and end-to-end malformed/SHA cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26ad365cac37071f84f90916ddd00ec974834e8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:
**Rebased onto `ee8a303` — and this PR now carries the fix for a red main.**

#167 and #169 merged without cross-rebasing: #169 pinned `NEW_FILE_KEY = 645f0e9e5cc472b3` over the fixture path `src/added.py`, which #167 (merged first) renamed to `src/app/clients/api/__init__.py`. Three `TestGitlabInlineDiscussionIdempotency` tests fail on pristine main (confirmed by the main-branch CI failure on the #169 merge commit). Commit `b731dfb` re-pins the literal to the independently recomputed value `a9cb7253f6710b82` — the literals stay hardcoded, preserving the tautology guard.

Post-rebase verification on this branch:
- Full mutation catalogue re-run: **M1–M5 all red** against their named tests, suite re-greened after each revert (details in the PR body's proof table; M2's wider 22-test sweep is index-shift collateral, with the unaffected-added-line claim carried by the four single-finding probes).
- #169's idempotency suites on the composed loop: **28/28 green**, plus a direct partition check — one finding per bucket prints all five buckets and sums to 5/5.
- Gates: 1229 passed; coverage 92.70% (floor 91.7); pre-commit 14/14.
- Register reconciled: 38 rows (29/9/0), ten tool-invisible relationships (arithmetic verified by counting, not by trusting either hunk).

Merging this PR returns main to green.

https://claude.ai/code/session_015GdQDn1WVh7VX9hgAjHcMS

Issue #91, opened 2026-07-30
The E2E resume test covers evidenceIsFresh's verifySlicesDegraded suppression but not its dimensionsLost suppression


leehopper:
Found by a design review of PR #87 while mutation-testing the health banner.

`workflows/test/stages_health_banner.test.js`'s end-to-end resume test runs `runWith` twice, feeding run 1's persisted checkpoint into run 2 — a genuine exercise of the replay path, not a mocked one.

But when the `evidenceIsFresh` gate on `dimensionsLost` was mutated away (so a resumed run would report a *rediscovered* population's dimension loss over a replayed delivered set), **the E2E test did not go red** — only the unit test did. Its fixture never re-loses a dimension on the replayed rediscovery, so end-to-end it exercises only the `verifySlicesDegraded` half of the suppression.

The guard is not missing — the unit test catches this mutation, and the suppression is correct. The issue is that the E2E test's coverage is narrower than its name and docstring imply, which is precisely the shape CLAUDE.md's mutation-check rule exists to catch: a test that reads as covering something it does not.

**Fix:** extend the resume fixture so the replayed run's fresh discover phase loses a dimension the original run did not, and assert the replayed banner does not mention it. Verify by mutation — remove the `evidenceIsFresh` gate on `dimensionsLost` and confirm the E2E test itself goes red, not just its unit-level sibling.

leehopper:
Closing (2026-07-30 triage): a test-gap finding against the discarded PR #87 branch's health-banner tests — owner decision 2026-07-30: that branch is thrown away and the work restarts fresh, tracked under #69. The requirement is recorded on #69: the fresh implementation's E2E resume fixture must go red when the dimensionsLost half of the evidence-freshness suppression gate is removed (mutation-verified per AGENTS.md), not only the unit-level sibling.

Issue #90, opened 2026-07-30
Verify slice-input ledger: 'rewritten' conflates attempted vs succeeded; unproven gap drops all but the last cause


leehopper:
Two low-severity diagnostic-fidelity issues in the slice-input proof ledger, found by a design review of PR #87 (issue #25 PR3). Neither loses a finding or misleads a report — both are internal/gap-text only.

**1. `inputProofStats.rewritten` conflates "attempted" with "succeeded."**
`verifySliceWithRetry` returns `rewritten: true` on the path where the re-materialize FAILED and the retry was therefore skipped, as well as on the paths where it succeeded. The counter's name reads as completed work.

Mitigating: `reviewHealth` surfaces only `recovered` / `unproven` / `degraded` from this ledger, never `rewritten`, so no report is misled — and a failed rewrite still increments `degraded`, so the pair `rewritten=1, degraded=1` does tell the true story to anyone reading both. Fix is either a rename (`rewriteAttempts`) or splitting the counter.

**2. The aggregated `verify-input-unproven` gap reports only the last cause.**
`unprovenDetail` is a single string overwritten per slice. `gradeInputProof` has two distinct unproven branches (no computable expectation vs. a missing receipt field); when slices in one run are unproven for both reasons, the aggregated message names only whichever came last, with no indication the other occurred.

No classification is lost — the count is right, only the explanation is non-representative. Untested for the multi-cause case (existing tests cover single-cause aggregation only).

Both deferred out of PR #87: they touch `workflows/src/stages.js`, and that PR's functional smoke had already launched against a pinned commit, so changing pipeline source there would have meant the pushed code was not the code that was smoked.

leehopper:
Closing (2026-07-30 triage): this describes ledger code that existed only on the discarded PR #87 branch — owner decision 2026-07-30: that branch is thrown away and the work restarts fresh, tracked under #69. Both fidelity requirements are recorded on #69 so the fresh implementation does not repeat them: `rewritten` must mean re-materialization succeeded (not attempted), and a multi-cause unproven gap must aggregate every cause, not keep the last.

Issue #88, opened 2026-07-30
coarsenLimits' lever heuristic omits VERIFY_REWRITES_PER_SLICE that worstCaseAgentCount includes


leehopper:
Found by a design review of PR #87 (issue #25 PR3).

`workflows/src/stages.js` `coarsenLimits` computes its lever-selection heuristic as:

```js
const verifyTerm = ceilDiv(...) * VERIFY_ATTEMPTS_PER_SLICE;
```

omitting the `+ VERIFY_REWRITES_PER_SLICE` term that `worstCaseAgentCount` correctly includes. The comment directly above it claims "Same VERIFY_ATTEMPTS_PER_SLICE scaling worstCaseAgentCount applies" — **that claim is false as written**, and PR3 is what made it false: it added `VERIFY_REWRITES_PER_SLICE` to `worstCaseAgentCount` (the re-materialize dispatch) without updating this heuristic.

**Not correctness-breaking.** The authoritative convergence guard calls the full, correct `worstCaseAgentCount`, so the loop still converges and stays under `AGENT_COUNT_GUARD`. The consequence is confined to *which lever* the heuristic picks: in a narrow window it can grow `validateBatch` / lower `challengeCap` when verify was actually the larger contributor.

**Why it wasn't caught:** no test constructs a case where the omitted term would flip the lever choice. The existing `coarsenLimits` tests in `stages_discover.test.js` don't reach that window.

Deliberately deferred out of PR #87 rather than fixed there: it is pipeline-affecting, and PR #87's functional smoke had already launched against a pinned commit. Fixing it there would have meant the pushed code was not the code that was smoked.

**Fix:** include the rewrite term in `verifyTerm`, and add a test that pins the lever choice in the window where the two terms disagree — a fix without that test would be unguarded in exactly the way this issue documents.

leehopper:
Closing (2026-07-30 triage, adversarially re-verified): the named asymmetry does not exist on main — `worstCaseAgentCount` and `coarsenLimits` compute the verify term with byte-identical expressions in workflows/src/stages.js, and `VERIFY_REWRITES_PER_SLICE` existed only on the discarded PR #87 branch. Per owner decision (2026-07-30) that branch is thrown away and the verify-boundary work restarts fresh under #69, whose body now carries the lesson: keep the lever heuristic and the worst-case term consistent, term for term.

PR #171, opened 2026-08-10
fix(verify): slice-input integrity — trailing-byte recovery + cross-runtime content proof (#69)

### LINKED ISSUE #69: One stray '}' from the artifact-writer costs the whole verify stage — 23 findings degraded to origin=unknown

leehopper:
## Summary

Closes #69 (the Wave 3 umbrella: fresh slice-input integrity, carrying #25's unlanded requirements 4–6 and the lessons from closed siblings #88/#90/#91).

One stray `}` appended by the artifact-writer after an otherwise complete slice-input document cost 23 intact findings their verification on `smoke-20260728-144630-a162ecd` (2 of 3 PRs degraded to `origin=unknown`). Two mechanisms close the class:

1. **Recovery** — `verify_findings.py` parses its input with `raw_decode` and accepts a trailing remainder of whitespace and/or unbalanced closing punctuation (`}`/`]`) only, pinned on the two byte-exact captures from that run (committed as fixtures). Recovered parses are disclosed loudly at every hop (`input_recovery` in the envelope, `RECOVERED:` on legacy stderr, a `verify: RECOVERED —` gap in the run report); everything outside the class — truncated value, duplicated document, trailing text, non-JSON whitespace — still hard-fails with the strict parser's message, byte-identical including the char offset.
2. **A real slice-input content proof** — the write proof used to confirm only that a path came back in the writer's echo. Now `verify_findings.py` echoes `receipt.input_checksum = fnv1a32(js_stringify_pretty(parsed doc))` (computed before numeric coercion), the workflow compares it against the same checksum computed over the content it dispatched, and a slice whose file does not carry the dispatched document degrades instead of being verified against a stranger's data. Recovery and proof compose: a recovered `}\n` slice still proves its leading document and stays trusted, with the recovery disclosed.

Mechanism details: a content-proof **mismatch** is deterministic (a retry re-reads the same bytes) and skips the retry; a **missing** proof is a sampled executor's transcription slip and retries. A document holding a number the two runtimes spell differently is **unprovable**: the check is skipped, counted, and the RECOVERED clause (if any) honestly drops its "proven against dispatch" claim. The per-run ledger `stats.inputProof = {slices, proven, recovered, mismatched, missing, unprovable}` is zero-populated on every path, disjoint by construction, and pinned by tests. `input_recovery` is shape-checked (`required: ['trailing_bytes']` + non-empty-string gate) so a schema-legal `{}` cannot fabricate a recovery disclosure. Cross-runtime checksum agreement is pinned by a new golden parity group `tests/fixtures/parity/slice_input_proof/` (5 cases: real slice shape, non-ASCII/astral/U+2028/U+2029/control, lone surrogate, safe-integer edges, empty findings).

## Requirements

| Requirement | Mechanism |
|---|---|
| #69 req 1 — `raw_decode`, accept trailing whitespace/closing punctuation | `load_input`'s recoverable-tail class (`_RECOVERABLE_TRAILING_RE`, JSON's exact whitespace set) |
| #69 req 2 — never silently; name the exact trailing bytes | `input_recovery.trailing_bytes` verbatim → `verify: RECOVERED —` gap; legacy `RECOVERED:` stderr |
| #69 req 3 — everything else hard-fails as today | strict-message reconstruction, byte-identical (incl. multi-line and non-JSON-whitespace cases) |
| #69 req 4 — pin on the byte-exact captures + gap fires + genuine corruption fails | `tests/fixtures/verify_slice_recovery/` (20,530 + 20,214 bytes, 11+12 findings), `TestSliceInputRecovery`, JS `(p1)`–`(p11)` |
| #69 req 5 — same treatment for assemble_artifacts.py? | **Decision: no** (see decision record item 3) |
| #25 req 4 — slice-input proof, mismatch degrades, deterministic and cheap | `receipt.input_checksum` vs dispatch-side expected; existing degrade path |
| #25 req 5 — length-vs-hash + writeArtifacts generalization | Decision record items 1–2 |
| #25 req 6 — writer contract line | Already present, `agents/artifact-writer.md:39` (see below) |
| #88 / #90 / #91 | N/A-with-reason (decision record) |

## Decision record

1. **Length vs hash (#25 req 5) — HASH.** `fnv1a32` over the canonical value form. A
   byte-length comparison misses every equal-length corruption (a swapped sibling slice of
   the same size is exactly the confusion `trustSlice` already documents defending
   against), and the checksum pair is already shared by both runtimes and already pinned
   across them by `tests/fixtures/parity/slice_input_proof/` (`workflows/test/parity.test.js`
   + `tests/test_parity_fixtures.py`) — a length would have been a second, weaker mechanism
   to maintain. It is a VALUE proof rather than a byte proof because the writer persists
   compact JSON (measured on both retained captures from `smoke-20260728-144630-a162ecd`),
   so a byte proof over the dispatched pretty form would fail 100% of real runs.

2. **`writeArtifacts` generalization (#25 req 5) — NOT generalized.** The proof kind
   follows the wire contract, and the two paths have different ones: the primaries reach
   disk through the workflow's own return and are then written by
   `materialize_artifacts.py`, which already computes content proofs over what it wrote
   (`workflows/AGENTS.md`, Persistence). Verify's slice inputs are the only remaining hop
   where a sampled agent's transcription is the *only* copy of the document a script then
   consumes. A shared abstraction over two different threat models would obscure both.

3. **`assemble_artifacts.py` (#69 requirement 5) — NO lenient recovery there.** The two
   hops are not symmetric. A recovered-but-corrupt verify slice input is consumed once,
   in-process, and its recovery is proven against what the stage dispatched. A recovered
   artifact would stay on disk as a *delivered primary* for every downstream consumer —
   bench, resume, Phase 8 posting — with no dispatch-side value to prove it against.
   Tolerating trailing bytes there would ship a file the pipeline has declared valid while
   it is not the file the pipeline produced. The structural failure stays structural
   (`workflows/AGENTS.md`: "missing file, unparseable JSON, duplicate id, bad plan
   checksum hard-fail").

4. **Deterministic-fault short-circuit.** A content-proof *mismatch* skips the retry (the
   file on disk is what it is); a *missing* proof still retries (a dropped field is a
   sampled agent's transcription failure and a fresh sample can fix it). No
   re-materialization machinery was added, so `worstCaseAgentCount` and `coarsenLimits`
   are untouched term for term — verified directly against this PR's diff:
   ```
   git diff main...HEAD -- workflows/src/stages.js | grep -n -A2 -B2 -i 'coarsen\|worstCase'
   ```
   shows only a new explanatory comment (naming both terms to say they don't change);
   no hunk touches either function's body, and `VERIFY_ATTEMPTS_PER_SLICE = 2`
   (`workflows/src/stages.js:589`) is untouched. The short-circuit only ever lowers the
   dispatch count, and a bound that is never exceeded is still a bound.

**N/A statements:**

- **#88** — N/A. No limits lever was touched. `VERIFY_ATTEMPTS_PER_SLICE`,
  `worstCaseAgentCount`'s verify term, and `coarsenLimits`' matching lever are unchanged;
  the short-circuit reduces actual dispatches below the bound without redefining it.
- **#90** — N/A. Describes discarded code that does not exist on `main`. This design has
  no `rewritten` counter (no re-materialization exists), `recovered` counts successes only
  by construction, and per-slice gaps are pushed into an array so two distinct causes
  cannot overwrite each other — the aggregation shape that caused the bug is absent, not
  patched.
- **#91** — N/A. Describes the discarded branch's evidence-freshness suppression gate
  (`dimensionsLost`), which does not exist on `main` and is not introduced here.

## Mutation proofs

Full per-row detail (exact edits, commands, before/after counts, restore confirmation) is
in `task-5-report.md`. All 16 rows were executed fresh in this session; every row produced
a detected (RED) result and every file was restored to a clean tree before the next row.

| # | Mutation | RED evidence |
|---|---|---|
| M1 | `load_input`: revert to `json.load`, drop tail handling | 10 failed / 143 passed (was 147/13) — all 4 named tests red |
| M2 | Delete the trailing-class guard | 8 failed / 145 passed — all 3 named tests red |
| M3 | Move checksum below numeric coercion | 1 failed / 146 passed — named test red |
| M4 | Always emit `input_recovery` (null when clean) | 2 failed / 145 passed — named test + 1 collateral red (see report for the 1 named test that didn't diverge on this exact edit site) |
| M5 | Let `JsSerializationError` propagate | 3 failed / 144 passed — named test + 2 collateral red |
| M6 | Delete `(4) INPUT PROOF` from `trustSlice` | 2 failed / 38 passed — `(p2)`, `(p3)` red |
| M7 | Gate RECOVERED gap on byte-count proxy | 3 failed / 37 passed — `(p4)` + 2 collateral red |
| M8 | Delete deterministic short-circuit | 1 failed / 39 passed — `(p3)`'s dispatch-count assertion red |
| M9 | Compact checksum in `materializeVerifySlices` | Full suite 25 failed / 631 passed (`(p1)`, `(p4)` + 11 collateral); `slice_input_proof parity: *` stayed green (108/108) exactly as expected — the parity check exercises the exported function, not this call site |
| M10 | Delete `firstUnsafeNumber` gate | 2 failed / 38 passed — `(p5)` + 1 collateral red |
| M11 | `recovered` increments regardless of recovery | 3 failed / 37 passed — `(p1)` + 2 collateral red (see report for the 1 named test that short-circuits before reaching this branch) |
| M12 | Remove `input_checksum` from schema + prompt | 1 failed / 39 passed — `(p7)` red |
| M13 | Remove `input_recovery` from schema + prompt | 1 failed / 39 passed — `(p7)` red |
| M14 | Recorder uses compact `json.dumps`, not the pinned pair | All 5 `slice_input_proof parity: *` cases red (103/108); goldens regenerated byte-identical after restore (`git status` clean), `TestGoldenFreshness` green |
| M15 | Emit `input_checksum` unconditionally (null, not omitted) | 1 failed / 146 passed — named test red |
| M16 | Always claim "proven against dispatch" | 1 failed / 39 passed — `(p8)` red |

Final state: `git status --porcelain` empty, `node workflows/build.js && git diff --exit-code
workflows/pipeline.js` clean. Post-catalogue verification ran the two directly-affected
Python files (155 passed / 121 subtests) plus the full JS suite (656/656); the full gates
were run separately in Task 6: `tests/` 1279 passed / 500 subtests, `bench/tests/` 537
passed / 22 subtests, JS 656/656.

## Writer contract

`agents/artifact-writer.md:39` (Protocol item 3) already reads:

> **The file contains the payload content exactly — nothing after the final byte.**
> No trailing commentary, no tool-call markup, no closing prose. A downstream script
> parses these files and checksums them; a single extra byte after the final `}` makes
> the file unparseable and fails the run's persistence.

This satisfies #25 requirement 6 and was not modified by this PR. The measured limitation
that motivates the mechanism this PR adds: the writer carried this exact instruction on
`smoke-20260728-144630-a162ecd` and appended `}\n` anyway on 2 of 3 PRs — prose cannot
enforce it, which is why the fix is a parser (the recoverable-tail class in `load_input`)
and a proof (the slice-input content checksum), not more instruction text.

## Verification

- Full gates (Task 6, this tree): `tests/` **1280 passed / 506 subtests**; `bench/tests/` **537 passed / 22 subtests** (bench untouched); JS **656/656**; bundle byte-stable (`node workflows/build.js && git diff --exit-code workflows/pipeline.js`); registries green (docs, machine-parsed strings, AGENTS.md set-size, contribution surface); `pre-commit run --all-files` 13/13.
- Coverage (local): scripts+.github **92.86%** (floor 91.7), bench **88.35%** (floor 87), JS **98.71 / 84.23 / 98.18** (floors 98 / 82.3 / 97). Four gates show >1.0 pp headroom locally; per the floor policy the raise keys off CI's measurement — if CI confirms, floors get ratcheted in a follow-up commit on this PR.
- 16/16 mutation rows RED (table below), executed fresh, tree restored clean after each.
- The RECOVERED gap cannot match bench's `_DEGRADE_RE` (checked against the live regex); `verify: UNVERIFIED` and retry-gap templates are byte-identical to main; `the deltas carry a checksum` byte-identical in all three registered locations.
- Biome js-lint is CI-only (pinned static binary, not available locally) — the one gate not run locally.

## Measurement policy

Suites-only per the #101 per-PR rule. This is the first Wave 3 item; it ships behind **Smoke A**, which runs after the last Wave 3 item merges. **Watch item for Smoke A: `stats.inputProof.mismatched` should be 0.** Both retained captures show the writer copies the prompt's compact JSON byte-for-byte (its only defect was the appended `}\n`, which now recovers), so the expected direction is fewer degrades, not more. A nonzero `mismatched` means the writer mangles slice inputs at a rate that needs a re-materialisation retry — deliberately out of scope here (decision record item 4).

## Residual risks (accepted, disclosed)

- **Key-order sensitivity**: the value proof hashes the canonical serialization without sorting keys, so a writer that re-serializes with transposed keys degrades the slice (deterministically, disclosed). Measured harmless at n=2 (both captures preserve key order byte-for-byte). Follow-up issue filed for a sorted-key canonical form if Smoke A ever shows it.
- **Disclosure transit**: the RECOVERED disclosure rides the executor's echo; the schema and prompt both name the fields (and `input_recovery` is required-shaped when present), but a sampled executor that drops the optional field loses the disclosure for that slice — the verdicts stay proven either way. Structural alternatives all route model-generated bytes through more model hops, not fewer.
- A byte proof over the compact form was considered and was available (the captures are byte-canonical): it would additionally catch key reordering and duplicate-key collapse. The value proof was chosen to reuse the plugin's one existing checksum pair and its parity infrastructure — one mechanism, one parity test.

https://claude.ai/code/session_017fxGhDyFf75U58uHN8fXHq

---

> [!NOTE]
> **Medium Risk**
> Changes core verify trust boundaries and parsing behavior for slice inputs; mistakes could wrongly trust or degrade findings, though extensive tests and deterministic-retry rules mitigate this.
> 
> **Overview**
> Closes the artifact-writer failure mode where a complete slice-input JSON document plus stray trailing `}`/`]` (or whitespace) used to fail verification and wipe intact findings. **`verify_findings.py`** now parses with `raw_decode`, recovers only that narrow tail class, surfaces **`input_recovery.trailing_bytes`** on the receipt envelope (and **`RECOVERED:`** on legacy stderr), and adds **`receipt.input_checksum`** (`fnv1a32` over canonical pretty JSON, computed before numeric coercion).
> 
> The **verify stage** gains guard (4): dispatch-time checksum vs echoed **`input_checksum`**, with mismatch treated as deterministic (no retry), missing proof retryable, and **`stats.inputProof`** ledger (`proven` / `recovered` / `mismatched` / `missing` / `unprovable`). Recovered-but-trusted slices emit **`verify: RECOVERED —`** gaps without UNVERIFIED degradation. Executor schema, prompts, and docs are updated to echo optional **`input_checksum`** and **`input_recovery`**; test fixtures and parity cases pin recovery bytes and JS/Python checksum agreement; CI coverage floors are bumped slightly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e11c7281c93c825b1ae4841100a552fc37636964. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


workflows/src/stages.js
```diff
@@ -743,12 +810,26 @@ function verifyDegradeGap(detail, k, n) {
...
+  if (first.deterministic) {
+    return {
+      ok: false,
+      reason: `${first.reason} — not retried: a re-dispatch reads the same slice-input file, so this fault is deterministic`,
+      inputFault: first.inputFault,
+    };
+  }
 
   const second = await attempt(`${nonce}.${i}.r1`, `verify-slice-${i}-retry`);
   if (second.ok) {
```
leehopper, line 835: **💡 [LOW] Verify-stage receipt/trust contract changed but two live skill reference docs still describe the old shape**

This diff adds a fourth trust guard (input_checksum/input_recovery), a new RECOVERED gap format, and a stats.inputProof ledger to the verify stage, and updated agents/executor.md and docs/machine-parsed-strings.md accordingly, but left skills/code-gauntlet/references/validation-pipeline.md and phase3-dispatch.md describing the old four-field receipt with no mention of input_checksum, input_recovery, or the new guard/gap.

**Suggested fix:**
Update validation-pipeline.md's receipt example and trust-guard list to include input_checksum/input_recovery and guard (4), and reword phase3-dispatch.md's 'four receipt fields' / 'status, the four receipt fields' phrasing to match the now-variable field set, mirroring the wording already applied to agents/executor.md and docs/machine-parsed-strings.md in this same PR.

workflows/src/stages.js
```diff
@@ -759,14 +840,19 @@ async function verifySliceWithRetry(c, inp, i, slice, { model, nonce, headShaSho
       ok: true,
       verified: second.verified,
       gap: `verify-slice-retry: slice ${i}'s first executor dispatch was untrusted (${first.reason}); a second dispatch was trusted and this slice's verified findings are from that attempt`,
+      recovery: second.recovery,
     };
```
leehopper, line 844: **💡 [LOW] No test exercises the 'missing input_checksum on attempt 1, valid on the retry' recovery path**

verifySliceWithRetry treats a missing input_checksum as retryable (unlike a mismatch, which is deterministic and skips the retry), and the retry branch forwards second.recovery through to the caller, but no test ever makes the retry attempt actually succeed with a correct input_checksum after the first attempt dropped it, so the 'a fresh sample is a plausible fix' claim behind the retryable classification is never exercised end-to-end, nor is second.recovery's wiring into the inputProof ledger on a retry-success path.

**Suggested fix:**
Add a test in workflows/test/stages_verify.test.js that builds an executor mock keyed on the `attempt` parameter: attempt 1 omits input_checksum (or sets it to null), attempt 2 returns a correctly-stamped envelope (optionally with input_recovery set too), then assert out.verified===true, the retry-disclosure gap is present, and out.inputProof correctly counts the slice as proven or recovered (not missing).

leehopper (commented):
Automated code-gauntlet review (v3, optimized tier). 5 findings discovered, 2 survived filtering + blind-challenge verification.

---
Generated by code-gauntlet | Reviewed up to: f04f87c8f1180e774d668a1de41311a99750c360

PR #175, opened 2026-08-11
fix(discover): close the finding item schema to undeclared fields (#53)

### LINKED ISSUE #53: Artifact-writer does not faithfully transcribe its payload (3 of 3 live runs diverged)
## Problem / Goal

`artifact-writer` (an agent, `agents/artifact-writer.md`, tools `Write`/`Read`) persists `findings.json` and `report.md` by hand-typing them into a `Write` call from the payload it was handed. Issue #38 added a content proof (`fnv1a32` checksum + char count, recorded in the persist plan and checked by the compact Workflow return) that measures, for the first time, whether the bytes that land on disk match the payload the writer was given. Today's 3-PR functional smoke (`bench/workspace/runs/smoke-20260727-205454-f99d948`, commit `f99d948`, 2026-07-27) is the first live measurement of that boundary, and it diverged on **all three** runs:

- grafana/grafana#80329 — `findings.json` landed 16 chars different from the payload (31,579 vs 31,595 chars; `fnv1a32:0x6554cf07` vs `fnv1a32:0xe43688ef`). Tolerated by the pipeline (findings preserved, gap recorded); artifacts persisted.
- calcom/cal.com#22345 — 8 chars different. Tolerated; artifacts persisted.
- ai-code-review-evaluation/discourse-graphite#6 — the writer produced **invalid JSON** (over-escaped `\"` as `\\"`, breaking the document at line 99). `scripts/assemble_artifacts.py` correctly refused to derive the post-review/checkpoint projections from unparseable input (structural failure, wrote nothing further); the pipeline returned `artifactPaths: null` for every artifact and `checkpoints.truncated: true`.

This is **not new** and was **not introduced by #38**. The pre-existing code has always had the same agent hand-typing the same file, with nothing checking whether the typed content matched what it was given — only the detection is new. Given all three runs in a single smoke diverged, silent artifact corruption on this boundary has evidently been routine.

Corroborating: re-serializing the *on-disk* `findings.json` through the exact pretty-printer `assemble_artifacts.py` uses (`json.dumps(obj, indent=2, ensure_ascii=False, allow_nan=False)`) reproduces the on-disk bytes exactly — so the writer's own formatting is canonical, and the divergence is in the *data* the writer transcribed, not in how it was serialized.

A second, related defect on the same boundary: `findingItemSchema()` (`workflows/src/stages.js`) never sets `additionalProperties: false` on the finding item schema it builds, so StructuredOutput leaves undeclared fields **permitted** rather than stripped. This is why `suggestion` (a field every discovery agent is instructed to emit but which is not declared in `FINDING_PROP_TYPES` or any dimension's `schemaExtra` — see issue #47) survived on 13 of 13 findings across two of today's three PRs and vanished entirely on the third: not a deterministic contract, a permitted-but-unenforced pass-through that a transcription boundary can drop at will. The same non-determinism already affects the `agent` field — `stages.js` documents (`:226-233`) that `agent` "survive[s] the echo only stochastically" when undeclared but present in payload, for exactly this reason. Cross-referencing #47: this makes "always absent" the wrong frame for that issue's missing-field analysis — the fields are inconsistently present, not uniformly dropped.

Goal: the artifact-writer's persisted `findings.json` is provably identical to the payload it was handed, on every run, not just measured-and-tolerated when it diverges.

## Requirements

1. `findings.json` (and any other primary the writer transcribes by value) is either derived programmatically from a source the pipeline already holds (as `assemble_artifacts.py` already does for `post-review.json`/`checkpoint.json`), or the writer's transcription is verified byte-for-byte against the payload before the run is allowed to report success — not merely measured and tolerated.
2. A structurally invalid write (discourse-graphite#6's shape) must not silently degrade to `artifactPaths: null` for every artifact without an unmissable, specific signal distinguishing "the writer mistyped valid-shaped-but-wrong content" from "the writer produced something that doesn't parse at all" — today both currently surface only as `gaps` text.
3. `findingItemSchema()`'s permissiveness on undeclared fields is resolved: either declare `additionalProperties: false` (forcing every field an agent contract instructs into `FINDING_PROP_TYPES`/`schemaExtra` explicitly, coordinated with #47's fix) or explicitly document that undeclared-field survival is accepted as non-deterministic and out of scope for any contract that depends on a field's presence.
4. Whatever remedy lands here does not regress the retry-once behavior `commit 3849e88` already shipped (report-JSON-unwrap + single deterministic retry before degrading) — this issue builds on that floor, it does not replace it.

## Evidence

Byte-length/checksum divergence, from today's smoke (`bench/workspace/runs/smoke-20260727-205454-f99d948`, 2026-07-27), read from the collected `workflows/wf_*.json` records' `result.gaps`:

```
grafana/grafana#80329 (wf_cb7cea68-9a4.json):
  "artifact-content-proof: .../code-gauntlet-findings-04cfa3bf.json bytes on
  disk differ from the payload handed to the writer (expected 31595
  chars/checksum fnv1a32:0xe43688ef, got 31579/fnv1a32:0x6554cf07)"

calcom/cal.com#22345 (wf_a6eaf1a7-e1f.json):
  "artifact-content-proof: .../code-gauntlet-findings-da09bc08.json bytes on
  disk differ from the payload handed to the writer (expected 38599
  chars/checksum fnv1a32:0x094cb97f, got 38591/fnv1a32:0xad3ff88f)"
  (8-char divergence: 38599 - 38591 = 8)

ai-code-review-evaluation/discourse-graphite#6 (wf_089793e7-ca8.json):
  "writeArtifacts: assemble script reported failure (expected artifact is
  not valid JSON: .../code-gauntlet-findings-267d8be1.json (Expecting ','
  delimiter: line 99 column 182 (char 12156))) — artifacts not persisted
  (partial-artifacts)"
  result.artifactPaths: { "findings": null, "report": null, "postReview":
  null, "checkpoints": null }
  result.checkpoints: { "completed": [...8 phases...], "truncated": true }
```

Re-serialization proof that the writer's formatting is canonical (grafana's on-disk `findings.json`, 2026-07-27):

```
$ python3 -c "
import json
p = 'bench/workspace/runs/smoke-20260727-205454-f99d948/pr-grafana-grafana-80329/code-gauntlet-findings-04cfa3bf.json'
raw = open(p, 'rb').read()
reser = json.dumps(json.loads(raw), indent=2, ensure_ascii=False, allow_nan=False).encode('utf-8')
print('on-disk bytes:', len(raw), 'reserialized bytes:', len(reser), 'identical:', raw == reser)
"
on-disk bytes: 31616 reserialized bytes: 31616 identical: True
```
(31,616 bytes / 31,579 chars — matches the "got 31579" figure in the gap above.)

`additionalProperties` is never set anywhere in the finding schema builder:

```
$ grep -n "additionalProperties" workflows/src/*.js
(no output — zero matches across all of workflows/src/)
```

`suggestion` field survival, counted directly from today's three persisted findings files (2026-07-27):

```
grafana#80329:              8 findings, 0 carry "suggestion"
cal.com#22345:               8 findings, 8 carry "suggestion"
discourse-graphite#6:        5 findings, 5 carry "suggestion"  (all 5 also carry "agent")
```
13 of 13 findings across cal.com + discourse-graphite carry `suggestion`; all 8 of grafana's do not — same field, same pipeline version, three different outcomes in one smoke.

`agent`'s documented non-determinism, `workflows/src/stages.js:226-233` (verbatim, 2026-07-27):

```
// `agent` is intentionally NOT declared (V3.1 item 4, REVERTED after mini-subset A):
// declaring it made the merge-injected agent identity survive the echo deterministically,
// which activated the filter's proximity-keyed cross-agent dedup everywhere (it requires
// 2+ distinct agents in a 5-line bucket — starved when agent is absent). Measured on the
// pre-registered 6-PR subset: dedup eliminations 7 -> 33 and same-6 recall 0.667 -> 0.433
// (-7 goldens), with dedup firing on exactly the Gate-1 PRs where agent happened to
// survive. ...
```

What #38 already shipped that this builds on (`scripts/assemble_artifacts.py`, `commit 3849e88`, both 2026-07-27 or earlier same-day):

```
$ ls -la scripts/assemble_artifacts.py
-rw-r--r--  1 lee  staff  31775 Jul 27 13:08 scripts/assemble_artifacts.py

$ git show 3849e88 --stat | head -5
commit 3849e884e3703104808437a53c4aba7463c1f3e9
fix(persist): unwrap JSON-wrapped reports and retry a refused derived persist once
```

- Cross-link #50: the durable direction — deriving `findings.json` itself (not just `post-review.json`/`checkpoint.json`) from a source the pipeline already holds rather than an agent re-typing it — is scoped there as "the only genuinely unique content the writer persists" today; that issue's remedy should probably absorb this one rather than the two being solved independently.
- For requirement 3 (`additionalProperties`), coordinate directly with #47 rather than deciding independently — the two issues are the same underlying gap (schema permissiveness vs. contract completeness) viewed from opposite ends (writer-side survival here, agent-contract-side declaration there).

## Out of scope

- Re-designing the delta-echo / per-slice mechanism for the *verify* boundary — that is Issue 3 and #25, a different agent (`executor`) and a different direction.
- Fixing the checker (`bench/runner/check.py`) that surfaced this — that is Issue 1, filed separately.
- Deciding the specific field list for `additionalProperties: false` — that decision belongs to whichever of #47 or this issue lands first; this issue only requires that the decision be made and recorded, not what it is.

## Verification

Tier: **smoke** — a writer-transcription fidelity fix is exactly the class of defect the 3-PR functional smoke exists to catch mechanically (pass/fail: does `findings.json` on disk match the payload, byte for byte, on every PR in the smoke). The always-on deterministic suites (including `tests/test_assemble_artifacts.py`, which already covers the derived-persistence round trip) are the baseline regression gate; a fresh smoke run after the fix lands should show zero `artifact-content-proof` gaps across all PRs, not merely tolerated ones.

leehopper:
Closes #53 (scoped by roadmap #101 to **requirement 3 only** — requirements 1/2 were resolved structurally by the harness-return persist architecture and their durable remainder is scoped under #50(b); requirement 4's retry-once floor is untouched by this diff).

## Decision (requirement 3): close the schema

`findingItemSchema()` now declares `additionalProperties: false`, so a discovery agent emitting a field no one declared is rejected by the platform's schema validator instead of passing through when it happens to feel like it. Issue #53 watched the same undeclared field survive 0/8, then 8/8, then 5/5 across three PRs of a single smoke — permitted-but-unenforced pass-through, never a contract.

Closing is safe now, and was not before:

- **The lockstep guarantee exists.** `FINDING_PROP_TYPES` + `FINDING_REQUIRED` + per-dimension `schemaExtra` all live in `workflows/src/registry.js`, and `tests/test_dimensions_registry.py` fails the build when the declared set drifts from what the agent .md contracts instruct (declared − instructed == {origin}). A closed schema therefore cannot reject a contract-instructed field.
- **The recall-sensitive dependency is gone.** The old hazard was `agent`: the filter's cross-agent dedup (`filterFindings.dedupCrossAgent`, re-run by `applyChallenges`) keyed on it, and it rode a verify echo that transcribed findings by value. Since PR #83 the verify executor returns per-id DELTAS — findings are rebuilt from the dispatched set at the join and `agent` is deleted there deterministically. Discovery agents never emit `agent` at all; it is stamped in code after StructuredOutput returns.
- **One boundary.** `findingItemSchema` feeds only the per-agent discovery dispatch (`findingSchema`). VALIDATE_SCHEMA returns per-id confidence adjustments, CHALLENGE_SCHEMA a score, VERIFY_SCHEMA a receipt plus deltas — none carries finding items. VERIFY_SCHEMA's receipt stays deliberately open for forward-compat and is untouched.

This is the mechanism instead of the instruction: undeclared-field emission becomes structurally impossible rather than stochastically tolerated.

Also corrects the comments that recorded the superseded truth: three stale pre-#83 clauses in `registry.js` ("the verify echo item schema unions them ALL"; "the executor drops the field when transcribing findings verbatim via the schema"; the FINDING_REQUIRED note's justification-by-verify-echo, echoed in a `finding_schema.test.js` comment), a false premise in `stages.js` `canonicalDeltas` ("an undeclared key does not survive StructuredOutput" — VERIFY_SCHEMA is deliberately open, so it can; the paragraph stands on `DELTA_VALUE_KEYS` alone), and three places in `tests/test_dimensions_registry.py` that told whoever trips the lockstep guard the undeclared field would be silently dropped. It is now rejected at dispatch.

## Tests

Two new pins in `workflows/test/finding_schema.test.js`:

- every discovery dispatch's item schema has `additionalProperties === false`
- a CLOSED schema declares every field it requires — `required ⊆ properties` was pinned nowhere, and a required-but-undeclared name is satisfiable while the schema is open and UNSATISFIABLE once it is closed (every emission would violate the schema and degrade the dimension)

**Mutation catalogue (all red):** removing `additionalProperties` fails the first; flipping it to `true` fails the first; adding a name to `FINDING_REQUIRED` that no property declares fails only the second.

## Verification

All gates green on this branch: JS 659/659 with coverage 98.71 / 84.07 / 98.18 vs floors 98 / 83.3 / 97.2 plus the presence check; `pytest tests/` 1280 passed @ 92.86% (floor 91.8); `bench/tests/` 537 passed @ 88.35% (floor 87); bundle byte-identical to a fresh build; `pre-commit run --all-files` green. Measurement note: JS branch coverage reads 84.07 vs main's last CI 84.23 — comment-shape churn, still 0.77 pp above the 83.3 floor, so no floor change is owed under the policy.

## Known risk → Smoke A watch item

`additionalProperties: false` is the FIRST use of a closed object in any dispatched schema in this repo, and the item schema pairs it with a PARTIAL required list (`FINDING_REQUIRED` is 8 of ~14 properties). Under a strict structured-output validator of the OpenAI kind — which rejects a closed schema that does not list every property in `required` — the failure would not be the extra retry it looks like: all seven discovery dispatches would fail schema validation before dispatch, every agent would return null, and `discover()` would degrade all nine dimensions. Suites cannot exercise the live platform validator, so this ships suites-only per roadmap #101 with the Smoke A watch item stated at that stake: **ALL SEVEN DISCOVERY AGENTS DISPATCH SUCCESSFULLY — zero null agents, zero degraded dimensions.** A wholesale discovery failure is the thing being watched for, not merely extra schema retries.

## Coordination (roadmap #101 Wave 3)

Lands alongside the sibling branch for #75 (`fix/75-verify-command-quoting`). Both edit `workflows/src/stages.js` but in different functions — #75 in `verifyCommand`/`assemblePrompt`, this PR in `findingItemSchema`/`canonicalDeltas` — and every other source file is single-owner; the generated `workflows/pipeline.js` bundle is the only shared surface. A trial merge of both final heads onto main is clean: 666/666 JS tests, 1280 pytest, and the merged bundle is byte-identical to a fresh `node workflows/build.js`. Suggested order: #75 first, then this PR. **After rebasing onto the merged #75: re-run `node workflows/build.js` and commit the regenerated bundle rather than trusting the textual auto-merge of `pipeline.js`, and re-run `pytest tests/ -q` — #75 edits `agents/executor.md` while this PR tightens `tests/test_dimensions_registry.py`'s contract scan (combined tree measured green here; re-verify after rebase).**

https://claude.ai/code/session_01PiWyAeUdUpC8X9DTYjV9ti

---

> [!NOTE]
> **Medium Risk**
> First closed object schema in dispatched workflows pairs partial `required` with `additionalProperties: false`; if the live validator is stricter than tests assume, all discovery agents could fail before dispatch (called out in the PR as Smoke A watch).
> 
> **Overview**
> Closes **#53 requirement 3** by making discovery finding items a **closed JSON Schema**: `findingItemSchema()` in `workflows/src/stages.js` (and the bundled `pipeline.js` / `registry.js` comments) now sets **`additionalProperties: false`**, so extra fields from agents are **schema violations** (retries / possible dimension failure) instead of silently passing through StructuredOutput.
> 
> **Docs and comments** across `REVIEW.md`, `agents/AGENTS.md`, `report-format.md`, and `tests/test_dimensions_registry.py` now describe **rejection at dispatch** rather than silent drops. Stale notes about verify echo unioning finding schemas or undeclared keys being stripped by StructuredOutput are corrected (verify uses **open** delta schemas; discovery is the only finding-by-value boundary).
> 
> **Tests** add pins in `workflows/test/finding_schema.test.js`: every discovery dispatch has `additionalProperties === false`, and **`required` ⊆ `properties`** so a closed schema cannot require undeclared names.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e205731c5a5034d0c5f458621c9815f22b44e0f1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/workflows/src/stages.js
+++ b/workflows/src/stages.js
@@ -300,11 +300,23 @@ export function agentActive(spec, agentFlags) {
 // descriptions for every downstream stage (validate/filter/challenge) and false-firing the
 // filter's short-description injection guard on high-confidence findings.
 //
-// An UNDECLARED property is dropped by the same mechanism, silently and by design — which is
-// why FINDING_PROP_TYPES and every dimension's `schemaExtra` live together in registry.js and
-// are pinned against the agent .md output contracts by tests/test_dimensions_registry.py
-// (issue #47: `suggestion`/`claude_md_rule`/`spec_text`/`criticality`/`failure_scenario` were
-// instructed by the contracts, declared by nothing, and dropped at this boundary on every run).
+// An UNDECLARED property is REJECTED at dispatch, not silently dropped: `additionalProperties:
+// false` makes a field no one declared a schema violation the platform retries, rather than
+// pass-through tolerated at random (issue #53 watched one undeclared field survive 0/8, then
+// 8/8, then 5/5 across three PRs of a single smoke — permitted, never a contract). Closing the
+// item schema is safe because the declaration is complete and STAYS complete: every field an
+// agent contract instructs is one entry in FINDING_PROP_TYPES or a dimension's `schemaExtra` in
+// registry.js, and tests/test_dimensions_registry.py fails the build when the declared set
+// drifts from what the .md contracts instruct (declared − instructed == {origin}), so the
+// closed schema cannot reject a field a contract asks for (issue #47: `suggestion`/
+// `claude_md_rule`/`spec_text`/`criticality`/`failure_scenario` were instructed by the
+// contracts, declared by nothing, and dropped at this boundary on every run). The blast radius
+// is ONE boundary — this item schema reaches only the per-agent discovery dispatch
+// (findingSchema below); verify echoes per-id deltas, validate per-id confidence adjustments,
+// challenge a score, none of them finding items. `agent`, the one field whose undeclared
+// survival ever mattered (the filter's cross-agent dedup keyed on it), is not emitted by
+// discovery at all: discover() stamps it in code after StructuredOutput returns and
+// joinVerifyDeltas deletes it — both deterministic.
 // Entries are EITHER a type-name shorthand string ({ k: 'string' } -> { type:'string' }) OR a
 // full JSON-Schema fragment used verbatim (how array-valued fields like cross_file_refs and
 // affected_consumers are declared); the shorthand keeps the common case terse while the
@@ -316,7 +328,7 @@ function findingItemSchema(schemaExtra) {
   for (const [k, t] of Object.entries({ ...FINDING_PROP_TYPES, ...(schemaExtra || {}) })) {
     props[k] = typeof t === 'string' ? { type: t } : t;
   }
-  return { type: 'object', properties: props, required: FINDING_REQUIRED };
+  return { type: 'object', properties: props, required: FINDING_REQUIRED, additionalProperties: false };
 }
 
 // Canonical finding schema (per-dimension schemaExtra unioned on top), wrapped in the
@@ -1113,10 +1125,10 @@ function verifySliceWriterPrompt(entries) {
 // verify_findings.py's build_deltas() emits exactly this shape in exactly this order.
 //
 // The rebuild therefore also makes the proof BLIND to any key outside DELTA_KEYS. That is
-// deliberate and not a hole: an undeclared key does not survive StructuredOutput, and
-// joinVerifyDeltas copies only DELTA_VALUE_KEYS, so a key the proof ignores is a key
-// nothing reads. Covering it would buy no protection and would cost the order- and
-// noise-tolerance that keeps a harmless echo quirk from degrading a slice.
+// deliberate and not a hole: joinVerifyDeltas copies only DELTA_VALUE_KEYS, so a key the
+// proof ignores is a key nothing reads — and VERIFY_SCHEMA is deliberately OPEN, so an
+// undeclared key genuinely can arrive. Covering it would buy no protection and would cost
+// the order- and noise-tolerance that keeps a harmless echo quirk from degrading a slice.
 function canonicalDeltas(ids, byId) {
   return ids.map((id) => {
     const src = byId.get(id) || {};
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -64,7 +64,8 @@ obvious from the diff alone; the full engineering rules live in `AGENTS.md`.
   `TextEncoder`, `URL`, `setTimeout`, `process` and `console` exist under `node --test` but not in
   the workflow sandbox, so a reference passes every test and throws on first live dispatch.
 - **A new finding field must appear in `workflows/src/registry.js`.** A field declared only in an
-  agent contract is dropped silently by StructuredOutput on every run.
+  agent contract is rejected at dispatch by the closed item schema (`additionalProperties: false`)
+  — schema retries, then a terminal agent failure that degrades the dimension.
 - **Instruction files are loaded on every turn.** Flag additions to `CLAUDE.md` or `AGENTS.md`
   that are derivable from the code, already stated in a comment at the site, or that record
   history, version numbers, or past incidents.
--- a/agents/AGENTS.md
+++ b/agents/AGENTS.md
@@ -7,9 +7,9 @@ Subagent contracts. Each `.md` is a system prompt with enforced YAML frontmatter
 - **LSP-first investigation.** Agents prefer `goToDefinition` / `findReferences` / `hover`, with
   Grep as fallback.
 - **A finding field is one entry in `workflows/src/registry.js` plus the owning agent's output
-  block.** A field no schema declares is **dropped silently** — StructuredOutput returns only
-  declared properties — so a field a contract instructs but the registry does not declare never
-  reaches merge, on every run. `tests/test_dimensions_registry.py` fails the build when the two drift.
+  block.** A field no schema declares is **rejected at dispatch** — the item schema is closed
+  (`additionalProperties: false`, issue #53) — so emitting it burns schema retries and, persisted,
+  fails the agent. `tests/test_dimensions_registry.py` fails the build when the two drift.
 - **Declared-but-optional fields are not nullable.** A not-applicable value is *omitted*, never
   `null`; the platform types each property to a single type and a null burns retries.
 - **`required` is one flat list shared by every dimension.** A field a contract calls required for
--- a/agents/CLAUDE.md
+++ b/agents/CLAUDE.md
@@ -11,9 +11,9 @@ Subagent contracts. Each `.md` is a system prompt with enforced YAML frontmatter
 - **LSP-first investigation.** Agents prefer `goToDefinition` / `findReferences` / `hover`, with
   Grep as fallback.
 - **A finding field is one entry in `workflows/src/registry.js` plus the owning agent's output
-  block.** A field no schema declares is **dropped silently** — StructuredOutput returns only
-  declared properties — so a field a contract instructs but the registry does not declare never
-  reaches merge, on every run. `tests/test_dimensions_registry.py` fails the build when the two drift.
+  block.** A field no schema declares is **rejected at dispatch** — the item schema is closed
+  (`additionalProperties: false`, issue #53) — so emitting it burns schema retries and, persisted,
+  fails the agent. `tests/test_dimensions_registry.py` fails the build when the two drift.
 - **Declared-but-optional fields are not nullable.** A not-applicable value is *omitted*, never
   `null`; the platform types each property to a single type and a null burns retries.
 - **`required` is one flat list shared by every dimension.** A field a contract calls required for
--- a/skills/code-gauntlet/references/report-format.md
+++ b/skills/code-gauntlet/references/report-format.md
@@ -34,7 +34,7 @@ For self-hosted instances, replace the hostname with the one detected from the g
 
 ## Finding Fields Reference
 
-The pipeline's declaration lives in `workflows/src/registry.js`. A field this table lists but the registry does not declare is dropped silently before any stage sees it — which is why `tests/test_dimensions_registry.py` pins this table, the registry, and the agent contracts to each other.
+The pipeline's declaration lives in `workflows/src/registry.js`. A field this table lists but the registry does not declare is rejected at the dispatch boundary (the item schema is closed — `additionalProperties: false`) before any stage sees it — which is why `tests/test_dimensions_registry.py` pins this table, the registry, and the agent contracts to each other.
 
 ### Canonical fields — every finding, every dimension
 
--- a/tests/test_dimensions_registry.py
+++ b/tests/test_dimensions_registry.py
@@ -7,10 +7,13 @@ nine dimension names in agents/AGENTS.md, the field lists there,
 and `references/report-format.md`'s Finding Fields Reference tables. A fourth — the seven
 discovery agents' `.md` output contracts — tells the models what to emit.
 
-Every one of those can drift from the registry silently, and drift is not cosmetic here:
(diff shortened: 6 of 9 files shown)


workflows/src/stages.js
```diff
@@ -316,7 +328,7 @@ function findingItemSchema(schemaExtra) {
   for (const [k, t] of Object.entries({ ...FINDING_PROP_TYPES, ...(schemaExtra || {}) })) {
     props[k] = typeof t === 'string' ? { type: t } : t;
   }
-  return { type: 'object', properties: props, required: FINDING_REQUIRED };
+  return { type: 'object', properties: props, required: FINDING_REQUIRED, additionalProperties: false };
```
leehopper, line 331: **🟡 [MEDIUM] Closed discovery item schema (additionalProperties:false) makes agents/AGENTS.md's 'dropped silently' claim stale**

stages.js's findingItemSchema now sets additionalProperties:false so an undeclared field is a schema violation that costs a retry and can fail the whole agent, but agents/AGENTS.md (and its generated twin agents/CLAUDE.md), REVIEW.md, and report-format.md still tell every reader and every dispatched subagent that an undeclared field is 'dropped silently' by StructuredOutput — the old, no-longer-true failure mode — because those docs were not updated in this diff even though the PR's own test/registry/stages comments were rewritten to describe the new rejection semantics.

**Suggested fix:**
Update agents/AGENTS.md's bullet (and re-run scripts/sync_agent_rules.py to refresh agents/CLAUDE.md), REVIEW.md's rule, and report-format.md's line 37 to describe rejection-and-retry (with possible terminal agent failure/dimension degradation) instead of silent drop, matching the new comment already written in workflows/src/stages.js and workflows/src/registry.js.

**Cited rule:** AGENTS.md: 'A rule stated in prose that could be enforced by code is a finding... the full engineering rules live in AGENTS.md' combined with 'Instruction files are loaded on every turn' (CLAUDE.md) — the prose here is not just unenforced, it now misdescribes the enforced mechanism.

leehopper (commented):
Code Gauntlet review: 2 doc-drift findings (both confirmed) — the discovery item schema was closed to reject undeclared fields (issue #53), but agents/AGENTS.md, REVIEW.md, and report-format.md still describe the old silent-drop behavior.

---
Generated by code-gauntlet | Reviewed up to: a4bc0cd23a26922382301a7637c29b73fc7548e1

workflows/src/stages.js
leehopper: Fixed in 290dc7b — all three surfaces (agents/AGENTS.md + regenerated CLAUDE.md twin, REVIEW.md, report-format.md line 37) now describe rejection-at-dispatch with schema retries and terminal agent failure, matching the stages.js/registry.js comments. Twin regenerated via scripts/sync_agent_rules.py; suites and pre-commit green.

PR #189, opened 2026-08-15
fix(pipeline): fail loud when every active discovery dimension degrades (#178)

### LINKED ISSUE #178: Run with every discovery dimension degraded still returns ok:true with an empty report
Observed in a Bedrock user transcript (2026-08-11, the same run that surfaced the first-party model-ID pinning bug): every agent dispatch failed with an invalid model identifier, all discovery dimensions landed in \`stats.degraded\` (\`bug, security, cross_file_impact, test_coverage, convention, intent, comment_accuracy, type_design, simplification\`), the run completed in 2 seconds with 0 findings — and the envelope still reported \`ok:true, phaseReached:"report"\`.

The orchestrating session happened to notice the degraded list and re-ran with a \`subagentModel\` override. A less attentive run (or a headless one) would have delivered "0 findings, clean review" on an unreviewed MR — the silent-success class the gap channel exists to prevent.

Proposal: when **all** active discovery dimensions degrade, the run should fail loud (\`ok:false\`, or at minimum a prominent dedicated gap + non-ok status) instead of producing an empty successful report. Partial degradation stays degraded-but-disclosed as today.

Found while fixing the model-ID pinning bug (policy.provider); out of scope for that change.

leehopper:
Closes #178.

## What

A run in which **every active discovery dimension degrades** now fails loud instead of returning `ok:true` with an empty "clean" report (live Bedrock incident 2026-08-11: invalid model ID null-isolated all 7 discovery agents, 2s run, 0 findings, `ok:true, phaseReached:"report"`). Partial degradation stays degraded-but-disclosed, byte-identical to before — pinned by test.

## Mechanism

- **`allActiveDimensionsDegraded(dispatched, degraded)`** (`workflows/src/stages.js`): dimension-level and scope-relative — expands the dispatched agentTypes through `agentSpecs()` into their owned dimensions, true only when every one of them is in `degraded`. A light-scope run whose two active agents both fail also fails loud. **Fails closed** on unresolvable scope: a replayed checkpoint whose agentTypes no longer resolve (registry rename / absent `dispatched`) with anything degraded reads as all-degraded, never as "nothing failed".
- **Guard in `runWith`**, immediately after the discover phase's gap merge, so it fires identically on a fresh dispatch and on a replayed checkpoint. On trip it returns a dedicated `ok:false` envelope (`error` prefixed `all-degraded:`, `failingPhase:'discover'`, `artifactPaths:{}`) that keeps the per-agent failure gaps and — unlike the top-level catch — carries `stats.degraded` and `resolvedPolicy`, the operator's diagnosis (the incident's root cause was exactly a model/provider mismatch). Nothing downstream dispatches; no empty clean-review artifact can ever reach disk on this path.
- **Resume semantics**: the degraded-to-nothing `discover` output is dropped from the returned resume checkpoints, so a retry re-dispatches discovery under (possibly corrected) policy instead of replaying the emptiness forever. Two conjuncts scope the guard so resume never loses real work product:
  - a replayable `challenge` checkpoint (the one phase the slim persisted artifact carries whole) means a prior attempt already delivered a review — a totally re-degraded re-discovery on that resume stays degraded-but-disclosed and the replayed delivery proceeds;
  - a findings-empty conjunct (vacuously true for same-version output) keeps the fail-closed arm from aborting a version-skew replay that still carries real findings.

## Consumers checked

- `SKILL.md` Phase 8 already branches on `ok:false` — routes into the existing failure path; both Phase-8 docs gained a short all-degraded paragraph.
- `scripts/await_workflow.py` terminal detection is value-indifferent (shape inventory comment updated; `ALL_DEGRADED_RETURN` fixture added to `tests/test_await_workflow.py`).
- bench never reads the envelope's `ok`/`gaps` (verified `invoke.py`/`check.py` end to end); bench classification of this new shape is deliberately out of scope here. The `all-degraded:` sentinel is registered in `docs/machine-parsed-strings.md` with parser `—`.
- Parity suite pins pure boundary functions only — no envelope coverage, no fixture changes.

## Verification

Two adversarial review rounds (correctness / test-quality / conventions lenses); 5 accepted findings applied in the second commit (resume-delivery preservation, fail-closed scope, `resolvedPolicy` value-fidelity assertion, inverse-boundary test, await_workflow shape inventory).

Mutation-verified per AGENTS.md — each mechanism reverted whole and watched go red on exactly the tests that name it (10 mutations across the two rounds): guard deleted → regression + checkpoint-replay tests red; helper forced false → same + unit tests red; `every→some` → partial-degradation pin red; resume-map deletion removed → checkpoint-exclusion assertion red; challenge-checkpoint conjunct removed → resume-delivery test red; fail-closed arm reverted → version-skew unit tests red; findings-empty conjunct removed → skew-replay-with-findings test red; `resolvedPolicy` hardcoded → value-fidelity assertion red; condition replaced with `findings.length===0` → inverse-boundary test red.

Suites (all run locally, byte-identical commands to CI):

- `tests/`: 1291 passed, 510 subtests
- `bench/tests/`: 537 passed, 22 subtests
- JS: 715/715; coverage 98.81 / 85.04 / 98.29 vs floors 98 / 83.3 / 97.2; presence check OK
- Python coverage: 92.89% (floor 91.8), bench 88.35% (floor 87)
- Bundle byte-fresh (`node workflows/build.js && git diff --exit-code workflows/pipeline.js`)
- `pre-commit run --all-files` green

Per the roadmap (#101), this is the first Wave 4 item: suites-only, merge on green CI, no smoke escalation condition applies.

---

> [!NOTE]
> **Medium Risk**
> Changes core orchestration outcomes for total discovery failure (ok:false vs ok:true) with nuanced resume/checkpoint exceptions; impact is mitigated by extensive tests and unchanged partial-degradation paths.
> 
> **Overview**
> Fixes a **silent success** class of failures: if every active discovery dimension degrades (e.g. invalid model ID null-isolating all agents), `runWith` now **aborts after discover** with `ok:false`, an `all-degraded:` `error`/`gaps` line, `stats.degraded`, and `resolvedPolicy` for diagnosis—**no merge/report/persist** and no empty “clean” review on disk.
> 
> The trip uses **`allActiveDimensionsDegraded(dispatched, degraded)`** (scope-relative to dispatched agents; **fail-closed** when `dispatched` no longer resolves after registry skew). It only fires when findings are empty and there is **no replayable `challenge` checkpoint**, so partial degradation, genuine zero-finding reviews, version-skew replays with findings, and **resume-after-prior-delivery** still behave as before. On abort, **`discover` is omitted from resume checkpoints** so retries re-dispatch discovery.
> 
> **`resolvedPolicyEnvelope`** deduplicates success vs failure envelopes. Skill/Phase 8 docs describe all-degraded handling; `docs/machine-parsed-strings.md` registers `all-degraded:`. Tests add `nullAgentLabels` mock support, await-workflow terminal fixture, and broad regression coverage (mirrored in `workflows/pipeline.js` bundle).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb0f3398891abb2b487988b344f8b4fd1bb433aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


workflows/src/stages.js
```diff
@@ -3451,6 +3478,73 @@ export async function runWith(ctx, rawArgs) {
...
+      const resumable = { ...phaseOutputs };
+      delete resumable.discover; // a degraded-to-nothing discover output must never replay on resume — a retry re-dispatches discovery under (possibly corrected) policy
+      return {
+        ok: false,
+        error: 'all-degraded: every active discovery dimension degraded — no review was performed',
+        phaseReached,
+        failingPhase: 'discover',
+        artifactPaths: {},
+        stats: { discovered: 0, degraded: discoverOut.degraded || [] },
+        resolvedPolicy: { subagentModel: policy.subagentModel || null, provider: policy.provider || 'firstParty' },
```
leehopper: **🟡 [MEDIUM] resolvedPolicy envelope construction duplicated verbatim across two return paths**

The new all-degraded return path builds resolvedPolicy as { subagentModel: policy.subagentModel || null, provider: policy.provider || 'firstParty' } at line 3542, which is the exact same expression the pre-existing success-path return builds at lines 3715-3720 (with an explanatory comment there). The two copies live in the same function and will silently diverge if the default ('firstParty') or null-fallback logic is ever changed in only one place.

**Suggested fix:**
Extract a small helper, e.g. `function resolvedPolicyEnvelope(policy) { return { subagentModel: policy.subagentModel || null, provider: policy.provider || 'firstParty' }; }`, and call `resolvedPolicy: resolvedPolicyEnvelope(policy)` at both return sites (line 3542 and ~3715). Before: two separate literal object expressions with the same fallback logic. After: one named helper called twice, so a future change to the default only needs one edit.

**Cited rule:** AGENTS.md 'Extending should cost one edit.' — a change to the provider/subagentModel default currently requires editing this expression in two places to stay correct.

leehopper (commented):
code-gauntlet review: 1 suggestion found (0 blocking issues). Full multi-agent pipeline (bug/security/tests/conventions/cross-file/simplification/type-design) ran to completion; this finding survived independent verification and blind challenge.

---
Generated by code-gauntlet | Reviewed up to: 18e7ee3c9e801b3774bfdba10996915f5238b049

workflows/src/stages.js
leehopper: Addressed in fb0f339 — extracted `resolvedPolicyEnvelope(policy)` and both return sites now call it; the 'firstParty' rationale comment moved onto the helper. Full suites + coverage gates re-run green, bundle rebuilt byte-fresh.

PR #191, opened 2026-08-15
fix(review-md): one REVIEW.md contract — spec, parsers, and generators agree (#94, #113, #118)

### LINKED ISSUE #94: REVIEW.md contract: reconcile the documented format with the parser (thresholds, per-dimension, ignore forms)
## Problem

`references/review-md-spec.md` and both templates document markdown-heading fields (`## Confidence Threshold`, blank line, `70`). Both parsers — `scripts/filter_findings.py:213-241` and `workflows/src/filterFindings.js:106-121` — only match snake_case `confidence_threshold: 70` inside a fenced ```` ```yaml # code-gauntlet ```` or `<!-- code-gauntlet-config -->` block.

A REVIEW.md written to the published spec therefore yields **zero parsed thresholds** from either implementation, and silently falls back to the built-in defaults (55 non-security / 70 security).

This is survivable today only because the live v3 path parses REVIEW.md by model in SKILL Phase 2 and hands the result across the args waist — the scripts are the retained/parity path. That makes it latent rather than absent.

## Why it matters now

We are about to author a REVIEW.md for this repo, and `skills/build-review-md/` generates files in the documented format. Anything we or a user writes to the spec is inert against the retained parsers.

## Scope

Decide which format is authoritative and make the other conform — or make the parsers accept both. The choice, and whether `build-review-md` output changes, is an implementation decision for whoever picks this up.

## Verification

Per `bench/MEASUREMENT.md`. Suites-only unless live parse behaviour changes.
## Consolidated scope (2026-07-30 triage, owner-approved)

#95, #96, and #97 are closed into this issue — all four defects share one root cause (the documented REVIEW.md dialect and the parser dialect are different languages) and one format decision settles them together:

- **#95 — `dimension:"pattern"` ignore entries are inert.** `applyExclusions` (workflows/src/filterFindings.js:939-968; scripts/filter_findings.py:1209-1239) regex-escapes each entry as one literal substring; nothing strips a dimension prefix anywhere in the codebase. Phase 8's dismissed-findings write-back template and bench/golden/review_md_fixture.md emit exactly this inert form — update both to whichever format wins.
- **#96 — the security floor claim is false twice over.** review-md-spec.md:345,357 claims security confidence "always uses a minimum of 60" that "cannot be set below 60"; the code default is 70 (DEFAULT_SECURITY_MIN_CONFIDENCE) and the real semantics are `min(confidence_threshold, security_min_confidence)` — a lower-value-taking ceiling, not a floor. Same template section this issue rewrites; land in the same hunk.
- **#97 — per-dimension thresholds are parsed by nothing.** review-md-spec.md:133-153 documents per-dimension confidence thresholds with merge/override rules; parseReviewMd extracts only the flat keys and applyThresholdFilter branches only on security-vs-not. Decide: implement per-dimension support, or retract the documentation.
- **Design input (not a filed defect):** the root scaffolding template ships `confidence_threshold: 70` while build-review-md ships 85 (skills/build-review-md/SKILL.md:96) — both silently raise the non-security bar above the built-in 55 default, the exact footgun SKILL.md:270 warns about. Resolve or document while deciding the contract.

Sequencing: #24's REVIEW.md-in-pipeline-parse sub-PR lands strictly after this issue — routing REVIEW.md through the current code parser as-is would make the documented forms definitively inert. Wave 4 milestone; unit/parity suites.

### LINKED ISSUE #113: Reconcile REVIEW.md Focus with pipeline agentFlags
## Origin
Filed from #37 inventory ID R-012 (v2→v3 residue audit).

## Problem
`skills/code-gauntlet/references/review-md-spec.md:20-94` says a REVIEW.md `## Focus` section means only listed dimensions run. The live parser reads only `confidence_threshold`, `security_min_confidence`, `severity_threshold`, and `ignore`; dimension gating is controlled by `agentFlags`, which `SKILL.md` derives only from the trivial-scope gate. No path maps Focus entries to agent flags.

## Why not fixed in #37
Both remedies are behavioral contract decisions: implement Focus-driven gating, or remove the documented Focus contract. This may affect which review agents run and therefore needs roadmap placement in #101 rather than a Wave 2 suites-only assignment.

## Acceptance
- Choose and document the canonical Focus behavior.
- Make the parser, args waist, agentFlags derivation, and REVIEW.md spec agree.
- Add regression coverage proving the chosen behavior.
- Place this issue in the appropriate #101 wave before implementation.

### LINKED ISSUE #118: Choose the fallback for omitted per-dimension confidence thresholds
## Origin
Filed from #37 inventory ID R-043 (v2→v3 residue audit; deferred design call discovered during Task 6).

## Problem
`skills/code-gauntlet/references/review-md-spec.md:151` and its matching template comment say dimensions omitted from a per-dimension block use the plain-number default, or 70 when no default is set. The live `DEFAULT_NONSECURITY_CONFIDENCE_THRESHOLD` is 55. This is the same prose-versus-constant class recorded in #37 as R-019, R-021, and R-022.

## Why not fixed in #37
Both values are plausible policy choices. Correcting the prose codifies 55; changing the constant changes live filtering and review recall. This needs a design decision and #101 wave placement, not a Wave 2 suites-only assignment.

## Acceptance
- Choose the canonical fallback for omitted non-security dimensions.
- Align the REVIEW.md spec and templates, parser/default constants, and both runtime implementations.
- Add focused tests that distinguish 55 from 70 for an omitted dimension.
- Place this issue in the appropriate #101 wave before implementation.

leehopper:
Closes #94. Closes #113. Closes #118.

## Decision record (one decision surface, per #94)

1. **Canonical machine config** is the block the parsers already accept — fenced ` ```yaml # code-gauntlet ` (or `<!-- code-gauntlet-config -->`) with exactly four snake_case keys: `confidence_threshold`, `security_min_confidence`, `severity_threshold`, `ignore`. The heading-based config dialect is retracted from the spec.
2. **#113 — `## Focus` retracted** as a gating contract (nothing ever parsed it; `phase2-triage.md` even referenced a `focus` key that never existed — removed, along with the twin in `phase1-preflight.md`). Dimension gating remains the automatic trivial-scope gate → `agentFlags`; the spec now says so. `## Skip` retracted on the same evidence (never a file filter). `## Rules` and free prose stay documented as advisory agent context — REVIEW.md text genuinely reaches agents by value.
3. **#118 / #97 — per-dimension thresholds retracted** (documented, parsed by nothing). Fallback codified: **non-security 55, security 70**; the self-contradicting "70 if no default is set" prose and both `#118` pointer comments are gone. A marker-anchored defaults statement in the spec is now asserted equal to the constants in *both* runtimes by a new test (`tests/test_review_md_contract.py`) — doc/code skew is a red suite from here on.
4. **#96 — security semantics corrected**: effective security threshold = `min(confidence_threshold, security_min_confidence)`, default 70 — a lower-value-taking ceiling, not a floor; the "minimum of 60" claim is deleted.
5. **#95 — `dimension:"pattern"` ignore form retracted**: entries are case-insensitive raw substrings vs title+description. Parsers now strip one pair of surrounding quotes; docs stop folding reasons/dates into the pattern (they made every written entry unmatchable).
6. **Generators stop pinning numbers**: `build-review-md` no longer writes 85 (or any value the user didn't choose), emits the canonical block, and never generates `## Skip`; the spec's scaffolding templates use digit-free commented placeholders.

## Behavior changes (all retained-path; live pipeline unchanged)

- `parseReviewMd` is dead code in the live JS runtime (no call sites outside parity tests; the sandbox cannot obtain REVIEW.md text — Phase 2 model-parses into `args.reviewConfig`). Verified in review.
- The retained Python CLI (`scripts/filter_findings.py`) now filters non-security at 55 (was 70) when the key is unset, matching the live JS Filter stage and SKILL.md's documented built-ins. Both parsers stopped pre-filling defaults into parse output — `applyThresholdFilter`'s `.get(key, DEFAULT)` fallback is now actually reachable and parity-covered.
- Parser hardening (both runtimes): key regexes are line-anchored so commented-out keys no longer parse as live config (previously the spec's own "commented" scaffolding template parsed to `confidence_threshold: 70, severity_threshold: medium` — silently re-pinning the exact values this contract removes).

## Bugs found and fixed along the way

- Phase 8 dismissed-findings write-back targeted a bare `## Ignore` heading the parsers can never read — every write-back through that template was dead config. Retargeted to the real config-block `ignore:` list, with a legacy-`deep-review`-block clause so an existing block is edited, not shadowed.
- Invalid CommonMark fence nesting (triple-backtick blocks inside triple-backtick blocks) in the spec and practices doc — outer fences widened to four backticks; every embedded example verified to parse to exactly what its prose claims via the real parser.

## Verification

- Two adversarial review rounds (7 verified findings in round 1, all fixed and re-verified in round 2; round 2 surfaced no new defects).
- Suites: pytest tests/ 1298 passed (coverage 92.93%, floor 91.8); bench 537 passed (88.35%, floor 87); node --test 718 passed (98.82/84.92/98.31 vs floors 98/83.3/97.2). Bundle rebuilt, byte-fresh. pre-commit all green.
- Every new regression mechanism mutation-verified red (Python 55 split, doc-constant guard, comment-anchor, quote-strip, no-pin parse — each reverted and observed failing the named tests).
- Suites-only per #94's Verification clause and the Wave 4 rule: no live parse behavior changes (verified above), so no smoke escalation.

---

> [!NOTE]
> **Medium Risk**
> Changes default confidence filtering for unset REVIEW.md config (non-security 70→55) and parser output shape; mitigated by parity tests and a doc/runtime constants contract test, but retained CLI users may see more findings.
> 
> **Overview**
> **Unifies REVIEW.md** so only the fenced `yaml # code-gauntlet` block is machine-readable; retracts misleading `## Focus`, `## Skip`, heading-based thresholds, per-dimension confidence, and `dimension:"pattern"` ignore syntax from the spec and wizard docs.
> 
> **Parser/filter behavior (Python + JS twins):** `parse_review_md` / `parseReviewMd` return only keys actually present in the file (plus `ignore`), so threshold filtering can apply the documented **55 non-security / 70 security** split when `confidence_threshold` is omitted; non-security filtering in the retained Python CLI moves from 70→55 in that case. Hardening adds line-anchored key regexes (commented examples no longer become live config) and strips one matching quote pair from `ignore` entries.
> 
> **Docs & tooling:** README and `build-review-md` stop pinning default thresholds or generating `## Skip`; dismissed-finding write-back targets the real `ignore:` list (including legacy `deep-review` blocks). New `tests/test_review_md_contract.py` ties spec default markers to constants in both filter implementations; parity fixtures cover commented keys, quoted ignores, and config-absent threshold split.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 869519a3dc95fa42e8e546f73bdc2927dec31bd5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_review_md_contract.py
```diff
@@ -0,0 +1,144 @@
...
+    )
+    assert m is not None, "defaults marker block not found"
+    block = m.group(1)
+    found = re.search(r"`severity_threshold`\s*\n?\s*\*\*(\w+)\*\*", block)
+    if not found:
+        raise AssertionError(
+            "Could not find a bolded value for `severity_threshold` inside the "
+            "code-gauntlet-defaults marker block."
+        )
+    return found.group(1)
```
leehopper: **🟡 [MEDIUM] Marker-block extraction regex duplicated across two helper functions**

_extract_doc_defaults and _extract_doc_severity both re-run the identical `<!-- code-gauntlet-defaults --> ... <!-- /code-gauntlet-defaults -->` regex search against the same text, and both call sites in the two test methods then stitch the results together with `doc["severity_threshold"] = _extract_doc_severity(...)`, working around a `None` placeholder left in the first function's return dict.

**Suggested fix:**
Merge _extract_doc_severity's logic into _extract_doc_defaults so the marker block is located once and all three values (including severity_threshold) are extracted and returned together. Before: doc = _extract_doc_defaults(text); doc["severity_threshold"] = _extract_doc_severity(text). After: doc = _extract_doc_defaults(text) returns {"confidence_threshold_nonsecurity": ..., "security_min_confidence": ..., "severity_threshold": ...} directly, with _extract_doc_severity's regex logic folded in as an internal step using the already-extracted `block`.

skills/code-gauntlet/references/delivery-guide.md
```diff
@@ -248,11 +254,22 @@ AskUserQuestion(
...
-If confirmed:
-
-- If no REVIEW.md exists → offer to create using scaffolding template from `references/review-md-spec.md`
-- If REVIEW.md exists without `## Ignore` → append the section
-- If `## Ignore` exists → append new entries
+If confirmed, the entries must land inside the parser's config block (`references/review-md-spec.md`
+→ Format) — a bare `## Ignore` heading with bullet items is never parsed, so writing there would
+silently produce dead configuration. **Check for an existing config block first, including a legacy
+one** — `parse_review_md`/`parseReviewMd` still recognize the pre-rename `` ```deep-review `` fenced
+form and `<!-- deep-review-config -->` comment form (`references/review-md-spec.md` → Format), and
```
leehopper, line 261: **💡 [LOW] Cross-reference to review-md-spec.md "Format" section for legacy marker support is inaccurate**

delivery-guide.md's newly-added guidance cites `references/review-md-spec.md` → Format as documenting that the parser still recognizes the legacy pre-rename ```deep-review``` fenced block and `<!-- deep-review-config -->` HTML comment block, but review-md-spec.md's rewritten Format section (also part of this diff) only describes the two current `code-gauntlet` block forms and never mentions `deep-review` anywhere in the file, so the citation points to content that isn't there.

**Suggested fix:**
Either add a short paragraph to review-md-spec.md's Format section describing the legacy `deep-review` fenced/comment forms (matching the code comment in filter_findings.py), or change the delivery-guide.md citation to point at the actual source (the parser code comments) instead of "references/review-md-spec.md -> Format".

leehopper (commented):
Code Gauntlet review: 3 findings survived full verification and blind challenge (1 main, 2 suggestions). See inline comments for details.

_Note: this run's verify stage reported a gap — the slice-input writer echo did not fully cover the dispatched slice paths, so origin tracking for all 8 candidate findings shows unknown (informational only; validation and challenge still ran normally)._

---
Generated by code-gauntlet | Reviewed up to: 1507c0464019e28a06120134a0d0c78626802655

tests/test_review_md_contract.py
leehopper: Addressed in 3fc5d82 — marker block is now located once in `_extract_doc_defaults`, which returns all three values; `_extract_doc_severity` and the call-site stitching are deleted. Mutation re-verified (spec 55→54 turns both contract tests red).

skills/code-gauntlet/references/delivery-guide.md
leehopper: Addressed in 3fc5d82 — review-md-spec.md's Format section now has a Legacy forms paragraph documenting the `deep-review` fenced/comment blocks (current-form-wins order), so the delivery-guide citation is accurate. Verified the added prose leaks no parseable values (`parse_review_md` on the full file still yields only `{'ignore': []}`).

Issue #198, opened 2026-08-17
Phase 1 eligibility check #4 over-fires: typo-only diffs stopped as 'trivially simple' before the light-scope gate


leehopper:
Observed live 2026-08-16 during the #24 verification tier (M4 trivial-PR fixture, first attempt; recorded in the #24 evidence comment).

## What happened

The M4 fixture's first attempt was a 2-line diff: a README typo fix + a code-comment typo fix. `references/phase1-preflight.md` check #4 reads:

> **Trivially simple?** — If ONLY lockfile/generated/auto-formatted changes with no logic modifications, stop.

Neither file is a lockfile, generated, or auto-formatted — but the headless child stopped anyway, reasoning "pure text/comment corrections with zero code-behavior impact… the review stops here." It never reached the Phase 2e light/full gate, dispatched nothing, and produced no artifacts.

## Why it matters

The abort condition and the light gate are adjacent but distinct (`phase2-triage.md` 2e separates them explicitly). Check #4's enumerated list is being read as an open-ended "too trivial to review" judgment — a prompt-adherence surface of exactly the kind #24 just removed from the scope decision. The light path (`deriveAgentFlags`) is deterministic once reached, but an over-broad check #4 starves it upstream: a docs/typo PR submitted for review gets a "nothing to do" response instead of the 2-agent light review the gate exists to offer.

## Fix directions (open)

(a) tighten the wording to an exhaustive, file-type-testable condition; (b) derive the abort mechanically from the changed-file list the Phase-1 composite already holds; (c) fold abort-vs-light-eligible-vs-full into the single riskTable classification #24 introduced, one contract. Per AGENTS.md, prefer the (b)/(c) mechanism shapes over more prose.

Natural home: adjacent to #35's question-surface pass (same Phase-1 gate territory) — absorb there if it fits, else the Wave 5 prompt-surface group.

## Verification

Suites-only for the change itself; the reusable M4 fixture (2-line and 15-line variants, scripts preserved from the 2026-08-16 session) is the regression check — the chosen contract should state the 2-line variant's expected outcome explicitly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DkugfQAn4UQTkbf3iu59cs

PR #199, opened 2026-08-17
refactor(skills): zero upfront questions — two-question Phase 8, state-based config gate (#35, #153)

### LINKED ISSUE #35: Re-evaluate and refactor the interactive question surface: schema drift, prompt count, and the enforcement-driven config gate
## Problem / Goal

The plugin's `AskUserQuestion` surface has accumulated **16 authoring sites covering 14 distinct decisions across five reference files** (plus 2 legacy-shape sites in `skills/build-review-md/SKILL.md`), guarded by **25 paired "Headless exception" blocks**. It has drifted in four independent ways.

**Schema drift.** 8 of 14 distinct headers exceed the 12-character chip limit — one is a 39-character sentence. Several option labels restate their own descriptions rather than staying concise. One site hand-authors a free-text escape hatch the UI already provides automatically. `multiSelect: true` is used at 1 of 14 decisions while three decisions that are genuinely multi-select — per-finding PR-comment selection, "pick from all findings" for the task board, and dismissed-findings picking — are modelled as single-select loops or free-form chat parsing. The `preview` affordance is unused plugin-wide despite two clear fits.

**Prompt volume, concentrated after the value is delivered.** The interactive floor for the happy path (REVIEW.md present, PR open, first review) is 3 blocking calls, two of which are pure confirmations. A "let me pick" delivery walkthrough asks one question *per finding*, unbounded: 20+ calls on a 12-finding review. Six of the sites live in Phase 8 — after the report already exists.

**An enforcement mechanism became a user-facing prompt.** Phase 2 halts the run if no `AskUserQuestion` was presented in Phase 1, so Phase 1 must ask a confirmation-only question even when REVIEW.md resolves everything — directly contradicting REVIEW.md's own pitch of "zero questions next time". The gate tests *that a question was asked* (self-attestation about a prior turn) rather than *that configuration is resolved*. Headless already models this correctly by gating on the printed config echo.

**The duplication has already drifted.** One decision ("no REVIEW.md found") now has three divergent treatments across three files; two templates are byte-identical duplicates; and a cross-file pointer is stale for prompts while still valid for scaffolding templates.

Goal: a designed, re-evaluated question surface — current against the tool schema, materially fewer interactive stops, single-sourced per decision, with the config gate resting on resolved state rather than on a prompt having occurred.

## Requirements

1. Every question conforms to the current `AskUserQuestion` schema: `header` ≤ 12 chars and written as a chip label rather than a sentence; option labels concise with detail carried in `description`; ≤4 questions per call and ≤4 options per question. No hand-authored option whose sole purpose is to collect free text the UI already collects (e.g. `phase1-preflight.md:47`); terminal options such as Cancel are retained.
2. No flow issues an unbounded per-item `AskUserQuestion` loop. Selection over N findings is bounded by a constant number of calls, using `multiSelect` where the decision is genuinely multi-select.
3. `preview` is used where the current design emits a separate chat message immediately before a question (the proposed REVIEW.md entries at `delivery-guide.md:189-196`; the per-finding detail crammed into the `question` string at `phase8-delivery.md:220`) — or the pickup session records why not.
4. Recommendation ordering is deliberate: where a recommended option exists it is first and labeled; where prose recommends one thing, option order must not contradict it.
5. **The Phase-2 configuration gate must be satisfiable by inspecting resolved configuration state, not by attesting that a prompt occurred in a prior turn.** The confirmation-only question then exists only if it earns its place on UX grounds.
6. The happy path (REVIEW.md present, PR open, first review) ends at **≤2** blocking calls, down from 3; the pickup session records a target for the no-REVIEW.md path in its spec before implementation. Phase 8's post-value prompt cluster is the primary reduction target.
7. Each decision is single-sourced: one canonical template per decision, referenced rather than re-authored. (The false-positive-exclusion duplication rationale in CLAUDE.md does not carry here — that guards against a failed file read across 7 independent agent contexts; these templates are read by one orchestrator.) The three-way "no REVIEW.md" divergence, the byte-identical duplicate, and the stale half of the cross-file pointer are fixed as part of this.
8. The headless contract is preserved exactly: every gate stays resolvable from the environment, no interactive prompt becomes reachable in a headless run, the 9 knobs and their precedence (`env pin > REVIEW.md explicit > headless default`) are unchanged, and the `Headless config:` echo keeps its current parsed form and byte-identical three-copy emission. If the headless-exception blocks are restructured, gate coverage must not regress.
9. Downstream consumers keep working: `args.delivery.tier`, `args.limits.deliveryCap`, `args.agentFlags.deep`, the delivery-method set driving Phase 8 Stage 1, and the Phase 8 `pr_comment_set` / `dismissed_set` handoffs. Removing the task-board offer (currently a MANDATORY GATE at `SKILL.md:268`) is a behavior change requiring an explicit decision, not incidental cleanup.
10. `skills/build-review-md/SKILL.md:46,:65` uses a fully obsolete call shape (top-level `question`, bare-string options, no `header`, no `multiSelect`) — in scope for the same sweep.

## Evidence

```text
Enforcement-driven prompt:
SKILL.md:79 — "Entry check: If no `AskUserQuestion` was presented during
  Phase 1, STOP — the configuration gate was missed. Return to Phase 1 and
  complete it before proceeding."
phase1-preflight.md:143 — "0 | Still present AskUserQuestion with a single
  confirmation question ... Never skip — Phase 2 checks that AskUserQuestion
  was called."
...contradicting phase1-preflight.md:189 — "No REVIEW.md found. Want to create
  one? It pre-configures review settings so you get zero questions next time."
Headless already gates correctly (headless-mode.md:79): passes if the
  `Headless config:` block was printed.
```

```text
Schema drift — headers over the 12-char limit (8 of 14 distinct):
  39  "{emoji} {Severity} — finding {M} of {N}"  phase8-delivery.md:221
  22  "Subdirectory REVIEW.md"    phase2-triage.md:173, review-md-spec.md:281
  20  "Review Configuration"      phase1-preflight.md:205
  19  "Previously Reviewed"       phase1-preflight.md:89, :107
  18  "Dismissed Findings"        delivery-guide.md:177
  17  "Save to REVIEW.md"         delivery-guide.md:206
  15  "PR Comment Tier"           phase1-preflight.md:174, :227
  15  "REVIEW.md Setup"           phase1-preflight.md:190, :231, review-md-spec.md:264
Hand-authored free-text escape hatch — phase1-preflight.md:47:
  { label: "Try a different number", description: "I'll provide the correct PR number" }
Label restates its description — phase1-preflight.md:92:
  { label: "Incremental — only changes since last review",
    description: "Review new commits only" }
multiSelect:true — 1 of 14 decisions (2 of 16 sites); `preview` — 0 sites.
```

```text
Prompt volume:
Happy path (REVIEW.md present, PR open, first review) = 3 blocking calls,
  2 of them pure confirmations. No REVIEW.md => 4-5.
"Let me pick" walkthrough = one call PER FINDING, unbounded (20+ on 12 findings):
  phase8-delivery.md:213 — "After showing the table, walk through each severity
  group one finding at a time."
Site distribution: 7 in Phase 1/2d, 3 in Phase 2c, 6 in Phase 8 (post-value).
Preview case — delivery-guide.md:189-196 prints "These entries would be added to
  REVIEW.md under ## Ignore:" as a separate chat message, then asks a bare
  question at :203-213 (question string at :205).
```

```text
Duplication that has already drifted — one decision, three treatments:
  phase2-triage.md:164     plain text, no question
  review-md-spec.md:263    full AskUserQuestion
  phase1-preflight.md:189  a third variant with different options
...while phase2-triage.md:185 points at review-md-spec.md "for the full prompts
  and scaffolding templates" — stale for the prompts half, still valid for
  scaffolding.
Byte-identical duplicate: phase2-triage.md:170 == review-md-spec.md:278.
Incomplete example: phase1-preflight.md:167 says omit the "PR comments" option
  for local targets, but the worked example at :217-230 is explicitly the
  "worst case" PR scenario, so the local-target variant is never illustrated.
```

```text
Load-bearing constraint (headless-mode.md:3):
  "In headless mode there is no user to answer an `AskUserQuestion`, so every
   interactive gate is resolved deterministically from the environment and the
   skill never prompts. A single `AskUserQuestion` call in a headless run
   deadlocks the process — the harness marks such runs invalid."
```

- Treat a resolved-config echo as the single gate artifact for both modes; interactive and headless would differ only in how values are sourced, not in whether an echo exists.
- Batch the Phase 8 walkthrough into one or few multi-select calls with `preview` carrying per-finding detail, replacing the per-finding loop.
- One canonical template per decision in a single file, with other references pointing at it.

The pickup session owns design (spec-driven, TDD), including the reduction targets and whether any prompt is removed outright.

## Out of scope

- Review-pipeline behavior: discovery, filter, verify, challenge, consolidation semantics.
- The headless knob set and its precedence (unchanged by this work).
- The severity-emoji mapping itself (`phase8-delivery.md:233`, `delivery-guide.md:34`) — owned by #36 (branding/output identity). This issue owns the `{emoji} {Severity} — finding {M} of {N}` header at `phase8-delivery.md:221`; if the walkthrough is restructured such that the header disappears, note it on that issue.

## Verification

Rung 2 of the four-tier ladder (suite-only → smoke → paired mini → full-15/holdout), per the ratcheted measurement policy (2026-07-23, canonically owned by #28 → `bench/README.md`). Tier: **smoke.**

Every item ships behind the always-on deterministic suites (pytest + `node --test`, parity and bundle-freshness included). This surface is prompt-side and interactive-only, but it borders the headless contract and the failure mode it risks is severe and mechanically detectable: an `AskUserQuestion` reachable in a headless run deadlocks the child, which the bench harness already detects and marks invalid. The per-sub-release functional smoke (2–3 PRs, mechanical pass/fail; asserts nothing about recall or noise) is therefore the regression gate — a headless prompt leak fails it immediately. That smoke is the release manager's run; this item does not gate on an individually-triggered one. Once #28's checker exists the smoke runs through it; until then a manual smoke suffices (soft dependency, not a blocker).

No paired benchmark: the interactive path is not exercised by scored runs, and any change touching `args.agentFlags.deep` or `args.delivery.tier` defaults must preserve current resolved values — proven by unit tests on the waist, not by a paid run.

Additional required coverage: a docs-parity test that extracts the `Headless config:` block from `references/headless-mode.md` and asserts every knob line still satisfies `bench/runner/invoke.py`'s line-anchored `EXPECTED_ECHO` pattern, plus a docs check that each headless gate row in `headless-mode.md` names an env knob (so gate coverage cannot silently regress).

### LINKED ISSUE #153: Phase 1 asks about model tier despite the fixed optimized policy and explicit never-ask prose
## Observed (live usage, 2026-08-07)

Kicking off a review, Phase 1 sometimes asks whether to run on the **optimized** or **frontier** tier — even though the frontier tier was eliminated, the policy is fixed (`policy.tier="optimized"`), and the repo's own REVIEW.md already says `optimized`.

## What this is not

There is no stale question text to delete. Current main and every cached plugin version (3.1.1 → 3.3.4) state the opposite: a non-`optimized` Model Tier **self-heals** — "never ask and never abort on this field" (`skills/code-gauntlet/SKILL.md` Phase 1, `references/phase1-preflight.md` Model policy). The only remaining "frontier" strings are that self-heal prose itself and an unrelated sentence in `review-md-practices.md`.

## Root-cause hypothesis

The orchestrating model improvises the question in violation of the prose prohibition. Two shipped ingredients make that likely:

1. `SKILL.md` instructs Phase 1 to "Check REVIEW.md for `model_tier`" and then explains the frontier/optimized distinction at length — handing the model the vocabulary for a question it is told not to ask.
2. Phase 2 halts the run if Phase 1 presented no `AskUserQuestion` ("never skip AskUserQuestion entirely"), so Phase 1 is forced to compose a question, and tier is sitting in its working context to pad it with.

Textbook instance of the project design rule: "never ask" is instruction-fallback where a mechanism should be. The model cannot ask about a field it never reads and the contract does not contain.

## Fix direction (structural, not more prose)

Belongs in the Wave 4 #94 / #35 pass:
- Remove `model_tier` from what Phase 1 reads — or remove `Model Tier` from the REVIEW.md contract entirely (#94 decides), keeping only the fail-loud `CODE_GAUNTLET_MODEL_TIER` env pin.
- Remove the Phase 2 asked-a-question halt that forces Phase 1 to compose a question (#35's "enforcement mechanism became a user-facing prompt").

Owner direction recorded on #35 (2026-08-07): the target interaction shape is **zero upfront questions** — run the review, then one delivery question at the end (post to PR/MR vs markdown-only), plus at most a final task-board question. Do not fix this issue by adding more never-ask text.

## Verification

Suites-only for any contract/prose mechanism change; the behavioral claim (question no longer appears) is confirmed via live usage, the standing bug channel between smoke gates.

leehopper:
Closes #35. Closes #153.

## Behavior changes (interactive mode)

- **Phase 1 asks nothing.** Removed: the delivery-preference question, the PR-comment-tier question, the REVIEW.md-setup question, and the confirmation-only question. Error-path questions (PR not found, draft PR, previously reviewed, trivial-scope) are unchanged — they fire on anomalies only. The PR-not-found question also drops its hand-authored free-text option (the UI collects free text itself).
- **Phase 8 asks exactly two questions, both after the report exists:** Delivery ("Post to PR/MR" vs "Markdown only") and Task board (Yes/No). The delivery question is skipped when there is nothing to post to (closed/merged PR, or a local diff).
- **Removed:** the per-finding delivery walkthrough (an unbounded question loop), the dismissed-findings suppression flow, and the save-to-REVIEW.md confirmation. This deletes the Phase-8 write-back path that appended dismissed findings to REVIEW.md's `ignore:` list — an explicit decision under the 2026-08-07 owner direction on #35; the list stays fully user-editable by hand per the #94 contract and `build-review-md` still scaffolds it.
- **"No REVIEW.md" / "subdirectory has no REVIEW.md" are non-blocking notices, not questions**, single-sourced in `review-md-spec.md` → Discovery (fixing the three-way divergence and the stale cross-file pointer).
- **Chat is no longer an interactive delivery destination** — the report is always on disk, chat gets a short completion summary, and the old full-report chat dump (the PR-310 regression class) has no remaining instruction path. `chat` remains a legal headless `CODE_GAUNTLET_DELIVERY` value with an explicit headless-only definition (requirement 8: knob semantics unchanged).

**Question counts:** happy path 5–6 → **2**; no-REVIEW.md path 5–6 → **2** (recorded target per requirement 6); trivial low-risk PR → **3** (Review Scope is anomaly-path). Worst case is no longer unbounded.

## #153, structurally

`model_tier` is no longer read from REVIEW.md anywhere. The `## Model Tier` spec section, the inheritance-table row, every `build-review-md` emission of the heading (SKILL + 7 practice examples + this repo's own REVIEW.md), and the frontier/optimized self-heal prose that handed the model the vocabulary to improvise a tier question are all gone. The only remaining pin is the fail-loud `CODE_GAUNTLET_MODEL_TIER` env knob, unchanged. An existing `## Model Tier` section in a user's REVIEW.md is inert prose — nothing reads it, nothing warns.

## Gate change

Phase 2's entry check no longer asks "was an AskUserQuestion presented in Phase 1" — it gates on a resolved-config echo. Headless keeps `Headless config:` byte-for-byte; interactive gains the sibling `Resolved config:` (4 lines, `key=value (source)`), registered in `docs/machine-parsed-strings.md` with the two producing copies pinned byte-identical.

## Headless contract

Unchanged: same nine echoed knobs, names, validation, and three-copy byte-identical emission; bench fixtures untouched and green. Two deliberate deltas, both issue-directed: `model_tier` lost its REVIEW.md precedence slot (#153 — env pin only now), and `CODE_GAUNTLET_DELIVERY_TIER` / `CODE_GAUNTLET_PR_COMMENT_CAP` now govern interactive runs too on the same precedence, since their questions are gone.

## New mechanism test

`tests/test_question_surface.py` extracts every `AskUserQuestion` block from `skills/**/*.md` and pins: the site count (**9** — 5 error-path in phase1-preflight, 2 in phase8-delivery, 2 in build-review-md), the schema (header ≤12 chars, 2–4 options each with label+description, explicit `multiSelect`), question-string uniqueness (single-sourcing), and the absence of prompt-history gate patterns ("Never skip AskUserQuestion" et al.). All six required mutations verified red (gate-text restoration, walkthrough restoration, long header, dropped multiSelect, duplicated question, broken fence walker), plus an independent 10th-site mutation in review round 1.

## Requirement 3 (preview) — why not

Both `preview` candidate sites named in #35 (the proposed-REVIEW.md-entries message and the per-finding detail question) lived in flows this pass deletes; nothing remains where `preview` would earn its place.

## Review process

Plan authored against a fresh line-verified survey; two adversarial review rounds (11 findings round 1 — all fixed, including README claim drift, orphaned step lettering, duplicated notice wording; 3 round 2 — all fixed) and a round-2 re-verification of every fix.

**Validation:** `pytest tests/` 1303 passed (+513 subtests); `pytest bench/tests/` 537 passed; `node --test workflows/test` 796 passed; `workflows/pipeline.js` byte-identical (no JS touched); coverage gate 92.93% ≥ 91.8 floor; pre-commit all green.

**Deferred:** #198 (Phase 1 check #4 over-firing) — distinct mechanism change, stays in the Wave 5 prompt-surface group.

---

> [!NOTE]
> **Medium Risk**
> Large orchestration and delivery contract changes across skill prose and user-facing review flow; pipeline JS is untouched but operator behavior and env precedence shifted materially.
> 
> **Overview**
> **Interactive UX is narrowed to anomaly prompts plus two end-of-run questions.** Phase 1 no longer asks about delivery, PR-comment tier, REVIEW.md setup, or confirmation; it prints a **`Resolved config:`** receipt and Phase 2 proceeds only when that block (or headless **`Headless config:`**) was emitted—not when any `AskUserQuestion` ran. Phase 8 becomes **Post to PR/MR vs Markdown only**, then a **Yes/No task board**; the per-finding walkthrough, chat as a delivery channel, and dismissed-findings write-back to `REVIEW.md` are removed.
> 
> **#153:** `## Model Tier` is dropped from specs, `build-review-md` output, and this repo’s `REVIEW.md`; nothing reads tier from REVIEW.md anymore (only the fail-loud `CODE_GAUNTLET_MODEL_TIER` env pin remains). **`CODE_GAUNTLET_DELIVERY_TIER`** and **`CODE_GAUNTLET_PR_COMMENT_CAP`** now apply in interactive runs the same way as headless, since their Phase 1 questions are gone.
> 
> **REVIEW.md discovery** is notice-only (canonical copy in `review-md-spec.md`); ignore lists are **hand-maintained**, not auto-appended on dismiss.
> 
> **Guardrail:** `tests/test_question_surface.py` pins **9** `AskUserQuestion` sites, schema rules, and bans prompt-history gate phrases. `docs/machine-parsed-strings.md` registers **`Resolved config:`** as a byte-identical two-producer invariant.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c910d18f69839bcff05d7a8ed534ea14b5fcb119. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/machine-parsed-strings.md
```diff
@@ -18,6 +18,7 @@ Tracking: issue #37. Sibling point-in-time audit:
 | string | producer path(s) | parser path(s) | notes |
 | --- | --- | --- | --- |
 | `Headless config:` | `skills/code-gauntlet/references/headless-mode.md`, `skills/code-gauntlet/SKILL.md` | `bench/runner/invoke.py`, `bench/tests/test_headless_echo_contract.py` | Three-copy emission must stay byte-identical across skill + headless-mode + any echoed block; see headless-mode.md |
+| `Resolved config:` | `skills/code-gauntlet/SKILL.md`, `skills/code-gauntlet/references/phase1-preflight.md` | `—` | Issue #35: the interactive twin of `Headless config:`. Phase 1 prints it once; Phase 2's entry gate keys on its presence, replacing the old "did an AskUserQuestion happen" check. No component parses it — a human and the model read it — so the row pins the two producing copies against each other, the same two-copy invariant shape as `HEADLESS CONFIG ERROR:` |
```
leehopper: **💡 [LOW] New 'Resolved config:' registry row cites the wrong sibling row for its 'two-copy invariant' analogy**

The new row's note claims 'Resolved config:' shares 'the same two-copy invariant shape as `HEADLESS CONFIG ERROR:`', but the HEADLESS CONFIG ERROR: row in this same file lists exactly one producer path (headless-mode.md only) and its own note describes it as a single-file fail-loud template, not a two-copy invariant at all. The row that actually models a multi-producer, byte-identical-copy invariant is `Headless config:` (two listed producers, explicitly 'the interactive twin' referenced one line earlier), so the comparison target looks like a copy-paste mix-up between the similarly named `Headless config:` and `HEADLESS CONFIG ERROR:` strings.

**Suggested fix:**
Point the analogy at `Headless config:` (the actual multi-copy, byte-identical echo block this row calls itself the twin of) instead of `HEADLESS CONFIG ERROR:`, or drop the comparison and just describe the invariant in the row's own terms.

skills/code-gauntlet/references/phase8-delivery.md
```diff
@@ -58,42 +60,65 @@ Always use the full 40-character SHA from `git rev-parse HEAD`.
...
+  persisted at `artifactPaths.report` (e.g. `{output_dir}/code-gauntlet-report-{head_sha_short}.md`). Tell
+  the user that absolute path in chat — that is the delivery. The only allowed write outside
+  `{output_dir}` is when the user explicitly names a destination path; then copy/write the report there.
+  Never invent a root-level `./code-gauntlet-{date}.md` (or any other default path outside
+  `{output_dir}`).
+
+Either way, print a short completion summary to chat: finding counts by severity, the report path, and the
+methodology pointer. Never dump the full report into the conversation unprompted.
+
+**Step 2 — Post via post_review.py**
```
leehopper: **💡 [LOW] 'Step 2' heading has no corresponding 'Step 1' after the Stage 1 rewrite**

Stage 1 used to number its sub-steps 'Step A / Step B / Step B.1'; the rewrite replaced the AskUserQuestion sub-step with 'Question 1 of 2 — Delivery' but left the following sub-step labeled 'Step 2 — Post via post_review.py', an orphaned numeral from the old scheme with no 'Step 1' anywhere in the file for a reader to match it against.

**Suggested fix:**
Either relabel the delivery question sub-step as 'Step 1' to match, or drop the 'Step 2' numeral entirely (e.g. rename to just 'Post via post_review.py') since the file no longer uses a lettered/numbered step scheme elsewhere.

tests/test_question_surface.py
```diff
@@ -0,0 +1,211 @@
...
+                offenders[key] = f"header {len(s['header'])} chars > {MAX_HEADER_CHARS}"
+            if s["multiSelect"] is None:
+                offenders[key + " multiSelect"] = "not stated explicitly"
+            n = len(s["options"])
+            if not MIN_OPTIONS <= n <= MAX_OPTIONS:
+                offenders[key + " options"] = f"{n} options"
+            for label, description in s["options"]:
+                if not label.strip() or not description.strip():
+                    offenders[key + " option"] = "empty label or description"
+        self.assertEqual(offenders, {}, f"schema failures: {offenders}")
```
leehopper, line 184: **💡 [LOW] offenders dict keys collide, silently dropping earlier failure messages**

In test_every_site_conforms_to_the_schema, the 'no question string' and 'no header'/'header too long' checks write to the same dict key (offenders[key]), so when a site fails both checks only the second message survives — the assertion still fails correctly, but the diagnostic the test prints on failure silently loses information about the first defect, making debugging a double-failure harder than it needs to be.

**Suggested fix:**
Accumulate messages per key instead of overwriting, e.g.: Before: `if not s["question"]: offenders[key] = "no question string"` ... `if not s["header"]: offenders[key] = "no header"`. After: `msgs = offenders.setdefault(key, [])` then `if not s["question"]: msgs.append("no question string")` and `if not s["header"]: msgs.append("no header")`, so every failure reason for a site is preserved in the assertion output.

leehopper (commented):
Automated code-gauntlet review — 3 findings (all low severity: 1 main, 2 suggestions). See inline comments.

---
Generated by code-gauntlet | Reviewed up to: 9156d44141d2abf4f70649ba6295c04ae1d76308

leehopper:
Addressed all review comments in c910d18:

- **Bugbot (high — posting ignores Markdown-only choice):** the `post_review.py` step is now explicitly gated on the "Post to PR/MR" answer in both `phase8-delivery.md` (the posting step opens with the gate; "Markdown only" ends Stage 1 at the completion summary) and `SKILL.md`'s Deliver section (posting runs only on that answer; headless only when `CODE_GAUNTLET_DELIVERY` includes `pr_comments`).
- **Registry-row analogy:** now cites the `Headless config:` row (the actual multi-producer byte-identical invariant), not `HEADLESS CONFIG ERROR:`.
- **Orphaned "Step 2" numeral:** dropped — the posting step is unnumbered, matching the rest of the file.
- **Offenders dict collision:** `test_every_site_conforms_to_the_schema` accumulates every failure reason per site via `setdefault` lists; the header-length mutation was re-verified red after the rework.

Issue #192, opened 2026-08-16
post_review: findings skipped by line-in-diff validation are silently dropped from PR delivery while the review body still counts them


leehopper:
## Problem

When `post_review.py`'s inline line validation (from the #127/#135 fix) rejects a finding because its anchor line is not in the diff, the finding is skipped with only a stderr warning and a stdout count line — it reaches the PR in no form at all. The review body, composed upstream by the pipeline, still claims the full finding count.

## Live sighting (2026-08-15, PR #191 review)

The pipeline delivered a review whose body said "3 findings survived full verification and blind challenge (1 main, 2 suggestions). See inline comments for details" — but only 2 inline comments were posted. The **main** finding (`cross-file-1`, MEDIUM: README.md's REVIEW.md example not updated to the #94 contract) was anchored at `scripts/filter_findings.py:216`, a file in the diff but a line outside the changed hunks:

```
WARNING: Skipping finding 'README.md's REVIEW.md example was not updated...' at scripts/filter_findings.py:216 — line not found in diff. Valid lines for this file: [70, 71, 72, 73, 74, 75, 76, 150, 151, 152]
Review posted: .../pull/191#pullrequestreview-4944880964
  2 inline comment(s) posted.
  1 finding(s) skipped (lines not in diff).
```

The reviewer reading the PR sees a count that doesn't match the comments, and the highest-severity finding of the run is the one that vanished. It was only recoverable from the on-disk report artifact.

## Contributing factor

The finding's own anchoring was poor — a README defect anchored to a parser file:line (the cross-file-impact agent anchored to the contract site rather than the stale doc site). Better anchoring would have dodged this instance, but delivery must still not silently drop: skip reasons are legitimate and will keep occurring (context lines, files outside the diff, renames).

## Direction

Skipped-by-validation findings should degrade into the review body (or a trailing section of it) instead of disappearing, and the body's stated count should reflect what was actually delivered inline vs appended. Both belong to `post_review.py` so the guarantee is structural — the model composing `review_body` upstream cannot know what validation will later skip.

Fits the Wave 5 delivery cluster (#22/#62/#66/#63 neighborhood) — it changes delivery surface, not discovery/filtering.

## Verification

Per bench/MEASUREMENT.md: suites-only (unit tests over the skip path; parity untouched).

Issue #84, opened 2026-07-29
Verify stage: `agent` is stripped on the trusted path but kept on the degraded path


leehopper:
## Problem

`agent` is withheld from findings on the verify stage's TRUSTED path and kept on its DEGRADED path. The two paths therefore hand the filter different data for the same finding, and which one a run gets depends on whether an executor dispatch happened to succeed.

Issue #25 requirement 1 is explicit that deterministic merge-injected `agent` reaching the filter is the measured recall-collapse mechanism: mini-subset A moved `dedupCrossAgent` eliminations 7 → 33 and same-6 recall 20/30 → 13/30 when `agent` survived the verify boundary. #25 PR2 (delta echo) made the withholding explicit and deterministic on the trusted path — `joinVerifyDeltas` deletes `agent` — because joining a delta onto findings the stage already holds would otherwise have made survival deterministic in the *wrong* direction.

`degradedSlice()` was deliberately left untouched by that PR. It re-emits the stage's own input findings with `origin='unknown'`, and those findings carry `agent` exactly as merge injected it. So:

| path | `agent` visible to filter |
|---|---|
| trusted (join) | never — explicit `delete` |
| degraded (any untrusted slice, write-proof failure, unusable ids) | always |

A degraded run is precisely a run that already lost its classification; giving it *additionally* the dedup behaviour the project measured as recall-collapsing is the opposite of conservative.

## Why it was not fixed in #25 PR2

Stripping `agent` on the degraded path changes findings content on that path, so it is a measurement-tier change — and PR2's justification rests on being finding-equivalent under the requirement-1 constraint. Changing the degraded path in the same PR would have widened the claim PR2 has to defend. `workflows/test/stages_verify_delta.test.js` pins today's degraded-path behaviour explicitly ("the degraded path is unchanged — it re-emits the dispatched findings as they were") so this cannot drift silently while the issue is open.

## Options

1. Strip `agent` in `degradedSlice()` too — one line, makes the stage's output uniform, needs the measurement tier that any dedup-affecting change needs.
2. Leave it and document the asymmetry as intentional (it is currently documented only as "not touched here").
3. Fold it into #22's cross-dimension consolidation redesign, which owns re-landing deterministic agent identity properly — at which point both paths get the same answer by construction.

Option 3 is probably right, but the asymmetry should not stay undecided and unfiled.

## Evidence

[...]

Issue #73, opened 2026-07-28
Mixed verified/degraded findings now share one consensus group: a never-verified finding can out-rank and displace a verified one from delivery


leehopper:
## Problem / Goal

`detectDisagreement` (`workflows/src/filterFindings.js`) groups findings by `(file, line_bucket(10))` and applies `CONSENSUS_BOOST = 10` to every member of a co-located group. It never inspects `origin`. Neither does `rankFindings` (`workflows/src/applyChallenges.js`), which orders on `(severity, -confidence, …)`.

That was harmless while verify degradation was all-or-nothing: a run's findings were **either** all script-verified (`origin` `new`/`surfaced`) **or** all degraded (`origin='unknown'`), so ranking only ever compared like with like. **#71 (issues #54 / #25 req 3) made degradation per-slice**, so a single run's array can now legitimately mix them — and the origin-blind consensus/ranking path mishandles that shape.

A finding whose own slice failed verification on **both** dispatch attempts — never fact-checked at all — can sit in the same location bucket as an unrelated finding from another agent that verified cleanly. It then receives `+10` confidence and a `corroborated_by` stamp, and can out-rank a genuinely verified finding, pushing that one past `limits.challengeCap` into the `unverified` bucket where it is never delivered.

Found by the adversarial review pass on #71 and reproduced.

## Evidence

Driving the real `detectDisagreement` and `rankFindings` exports (no mocks):

```
A origin=unknown conf=75 corroborated_by=["security-reviewer"]
B origin=new     conf=80 corroborated_by=["bug-detector"]
C origin=new     conf=74 corroborated_by=[]
ranked: B=80 > A=75 > C=74
challengeCap=2 -> delivered: B,A | dropped to unverified: C
```

Fixture: `A` (bug-detector, `x.js:12`, confidence 65, slice failed both attempts → `origin='unknown'`), `B` (security-reviewer, `x.js:13`, confidence 70, verified), `C` (bug-detector, `y.js:99`, confidence 74, verified, singleton). `A` and `B` share a bucket and both get `+10`. `C` — the only finding of the three whose content the script actually fact-checked — is the one that does not get delivered.

Repro script: <details><summary>poc.mjs</summary>

```js
import { detectDisagreement } from 'workflows/src/filterFindings.js';
import { rankFindings } from 'workflows/src/applyChallenges.js';
const F = [
[...]

PR #206, opened 2026-08-17
feat: consensus consolidation — boost-never-drop, agent-identity re-land, origin-aware ranking, payload consolidation (#22, #73, #84)

leehopper:
Closes-when-measured: #22, #73 (do not auto-close — the paired measurement gates the close; see Verification). Resolves the #84 asymmetry by construction.

## What this is

The Wave 5 consensus-consolidation change set, landed as one parity-locked PR per the owner triage on #22 (2026-07-30): cross-agent convergence now **boosts, never drops**; the deterministic merge-injected `agent` identity **re-lands** on the verify echo, uniformly on both trusted and degraded paths; consensus and ranking become **origin-aware** (#73); and consolidation now governs the **posted PR/MR payload** (the PR-310 near-duplicate inline-comment class), not only report-render.

## Design decisions

- **D1 — no drop-one-winner anywhere.** `dedupCrossAgent` is replaced by `consolidateCrossAgent` (both call sites: filter stage and the post-challenge re-run). Cross-agent proximity groups (5-line bucket, ≥2 distinct agents) are stamped with a shared `consolidation_key` and exactly one `consolidation_primary` (old winner key: isCore, confidence, description length — deliberately origin-blind, since nothing is dropped there). Nothing is eliminated; `eliminated_by: 'dedup:cross-agent'` no longer exists. Stamps are cleared at the start of every run so groups dissolved by challenge eliminations lose stale stamps.
- **D2 — consolidation governs the payload, not the findings array.** Findings stay distinct end-to-end (merging content is exactly the lossy shape that collapsed recall — distinct real bugs inside one 5-line bucket). Delivery groups them: one inline comment/discussion per consolidation group (primary rendered as today + corroborating sections, neutralized through the same #192 escaping); if the primary can't anchor, the whole group degrades into the #192 skipped section as individual entries. The report writer receives non-primaries folded under the primary as `corroborations` (documented in `agents/report-writer.md`); the findings artifact and dimensions table remain computed over the raw distinct array.
- **D3 — #73 decision (recorded on #73).** (a) The consensus boost is agent-agreement-derived but only among like origins: the `detectDisagreement` grouping key gains `degraded = (origin === 'unknown')`, so degraded and verified findings never corroborate each other. (b) `rankFindings` becomes `[severityIndex, degraded, -confidence, tertiary]` — within a severity, a verified finding always outranks one the pipeline could not verify. Both components are constant on uniform-origin runs, so all-verified and all-degraded behavior is unchanged (#73 req 2).
- **D4 — agent identity re-lands uniformly (#84 option 3).** `delete joined.agent` removed from `joinVerifyDeltas`; `degradedSlice` untouched; both paths now emit `agent` identically. Safe because no drop-semantics consumer of `agent` remains. The `stages_verify.test.js` (m5) and `stages_verify_delta.test.js` pins are flipped accordingly; `skills/code-gauntlet/references/validation-pipeline.md` updated to match.
- **D5 — #22 requirement 5 rider dropped:** zero `\'`/`\\'` escapes remain in the three named agent files on main (already fixed; consistent with the pickup triage note).
- **D6 — parity discipline:** JS twins and authoritative Python moved together; golden fixtures re-recorded via `record_parity.py` (fixture group renamed `dedup_cross_agent/` → `consolidate_cross_agent/`, three new mixed-origin cases); bundle regenerated and byte-clean.

## Adversarial review

Two independent passes. The opus adversarial pass produced 8 verified findings (3 high), all fixed in this PR: report-writer contract missing `corroborations`; a second `consolidation_primary` in a group silently dropping a finding (both twins); stale stamps surviving the post-challenge re-run (enabling same-agent payload merging); GitLab loss counters counting groups instead of findings; the line-is-None whole-group degrade fan-out unpinned on both platforms; stale skill-reference prose claiming `agent` is still stripped; skipped-section prose not covering group-degraded members; doc/comment skew (CONTRIBUTING fixture path, dedup wording).

## Mutation verification (every mechanism red)

| Mutation | Result |
|---|---|
| Remove `degraded` from consensus grouping key (JS / Py) | RED — 2 / 3 tests |
| Revert rank key to 3-tuple (JS / Py) | RED — 2 / 3 tests |
| Re-add `delete joined.agent` | RED — 3 tests |
| Reintroduce drop-one-winner in consolidate (Py) | RED — 21 failures |
| Drop corroborators from the report payload (JS) | RED — 3 tests |
| Remove stale-stamp clearing (JS / Py) | RED — both new tests |
| Remove line-is-None degrade fan-out (GitHub / GitLab) | RED — both new pins |

Python runs purged `__pycache__` per the stale-bytecode hazard; each mutation restored and re-verified green.

## Verification tier

**Tier: measurement** — this changes findings content, ranking, and delivery selection. This PR ships behind the deterministic suites only and makes **no recall/noise claims**. Completion of #22 gates on the owner-triggered paired mini-subset A run (6 pre-registered gate PRs vs baseline-of-record `custom-20260723-102149-381e9ff`, recall 19/30 = 0.6333, noise 0.2233). Note on #73 req 2: the uniform-origin pins in `consolidation.test.js` demonstrate the origin extensions are inert on uniform input; they are hand-written literals, not a byte-comparison against a recorded run of main — the paired run is the authoritative regression check.

Gates at HEAD: pytest 1346 passed + 516 subtests; node --test 822/822; bundle rebuild byte-clean; coverage 93.12% Py-scripts / 88.35% bench / 98.92-86.61-98.39 JS (all floors passed); pre-commit all hooks passed.

---

> [!NOTE]
> **High Risk**
> This changes which findings survive filtering, how they are ranked and posted, and rerun dedup behavior across the full review delivery path—recall impact is explicitly gated on measurement, not just unit tests.
> 
> **Overview**
> Replaces **cross-agent dedup** with **`consolidate_cross_agent`**: co-located findings from different agents keep a shared `consolidation_key` and one `consolidation_primary` instead of being eliminated (`dedup:cross-agent` is gone). The same stamping runs after challenges, with stamps cleared on re-run so dissolved groups do not leave stale metadata.
> 
> **Delivery and reports** group stamped findings: GitHub/GitLab post **one inline comment per group** (primary body plus corroborating sections), with fallbacks and counters that treat every group member as a finding. **`find_finding_markers`** collects all per-finding keys from a body so reruns do not repost (#132). The report writer and minimal fallback fold non-primaries into optional **`corroborations`** while dimension stats stay on the raw list.
> 
> **#73 origin rules:** consensus boosting only groups findings with the same verified vs `origin === 'unknown'` bucket; **`rank_findings`** prefers verified over degraded within the same severity. **Verify** no longer strips **`agent`** on join (safe now that dedup no longer drops on agent). Python/JS twins and parity fixtures are updated to match.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f8cb13d6c89db29e48517d675e4c09eaf205a94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/filter_findings.py
```diff
@@ -1041,9 +1048,9 @@ def group_by_proximity(findings, line_proximity=5):
...
 
-    This utility is shared between dedup_cross_agent and apply_challenges.py,
-    which re-runs that dedup after challenge scoring. (Challenge results are
-    matched to findings by id, not by proximity.)
+    This utility is shared between consolidate_cross_agent and apply_challenges.py,
+    which re-runs that consolidation after challenge scoring. (Challenge results
+    are matched to findings by id, not by proximity.)
     """
 
     def _bucket(line, proximity):
```
leehopper, line 1056: **💡 [LOW] New _line_bucket_proximity duplicates group_by_proximity's inner _bucket verbatim**

This diff adds a module-level _line_bucket_proximity (lines 1142-1149) whose body is byte-identical to the pre-existing nested _bucket closure inside group_by_proximity (lines 1056-1060) -- same try/round/except. The new function's own docstring even says it uses 'the same rounding rule group_by_proximity uses internally', acknowledging the duplication instead of removing it. The JS twin avoids this: groupByProximity and consolidateCrossAgent both already call one shared lineBucket helper (workflows/src/filterFindings.js:478, used at 774 and inside consolidateCrossAgent), so the Python side is now the odd one out with two copies of the same math that can silently drift if one is edited without the other.

**Suggested fix:**
Delete the nested `_bucket` closure in group_by_proximity and have it call the module-level `_line_bucket_proximity` instead, e.g. change `bucket = _bucket(finding.get("line_start", 0), line_proximity)` to `bucket = _line_bucket_proximity(finding.get("line_start", 0), line_proximity)`, matching how the JS twin already shares one `lineBucket` helper between groupByProximity and consolidateCrossAgent.

leehopper, line 1056: **💡 [LOW] consolidation_key derived from a second, independently-maintained bucket formula**

Encapsulation: 6/10, Expression: 5/10, Usefulness: 7/10, Enforcement: 4/10. group_by_proximity already computes and keys its returned dict by (file, bucket) using its local _bucket closure, but consolidate_cross_agent iterates groups.values() (discarding that key) and re-derives the bucket for consolidation_key via a second, module-level function, _line_bucket_proximity, reimplementing the identical round(int(line)/proximity)*proximity formula instead of reusing group_by_proximity's own key. The invariant that a finding's consolidation_key always matches the proximity bucket it was grouped under is guaranteed only by two formulas staying byte-identical, not by construction.

**Suggested fix:**
Have group_by_proximity return/iterate its (file, bucket) key alongside each group — e.g. for (fpath, bucket), group in groups.items(): — and build consolidation_key from that key directly, deleting _line_bucket_proximity so there is exactly one place that defines the bucketing formula, matching the JS twin's shared lineBucket design.

**Cited rule:**
> AGENTS.md > Design: "Extending should cost one edit. A new dimension, field, or agent belongs in one place. If it takes coordinated edits across N files, fix the shape rather than documenting the ritual."

leehopper (commented):
---

4 inline comment(s) were posted; the following 1 finding(s) reference lines outside this diff and are included here instead:

#### `scripts/apply_challenges.py:429`

**💡 [LOW] --help text still describes the redesigned step as 'cross-agent dedup'**

This PR rewrote apply_challenges.py's module docstring (lines 3-8, 42-44) to say the post-challenge step is 'cross-agent consolidation' where 'Nothing is dropped by this step,' but the ArgumentParser description shown by --help (line 431) still reads 're-runs cross-agent dedup,' preserving the old drop-one-winner terminology this PR replaced with consolidate_cross_agent.

**Suggested fix:**
Update the argparse description string to say 'cross-agent consolidation' to match the module docstring and actual (never-drops) behavior.

---
Generated by code-gauntlet | Reviewed up to: 2173e6067928a7b126df28659fdf7bb07a919988

scripts/filter_findings.py
```diff
@@ -1099,81 +1114,50 @@ def _winner_key(f):
...
-    kept = [
-        f for f in findings if f.get("id", "") in kept_finding_ids or not f.get("id")
-    ]
-    return kept, dropped
+        for f in group:
+            if not f.get("id", ""):
+                continue  # id-less findings stay unstamped
+            f["consolidation_key"] = consolidation_key
+            f["consolidation_primary"] = f is primary
+            consolidated_count += 1
```
leehopper, line 1138: **💡 [LOW] consolidate_cross_agent primary-selection skip-over-id-less-member logic has no assertion on the resulting consolidation_primary value**

consolidate_cross_agent must pick the primary by walking the priority-ranked list past any id-less member (an id-less finding cannot carry the consolidation_primary stamp), landing on the first ranked member that has a truthy id. The only test exercising a mixed id-less/has-id group (Python's test_no_truthy_id_passes_through_unstamped and the JS twin's 'findings without a truthy id pass through unstamped') checks that consolidation_key is set/absent but never asserts the value of consolidation_primary on the surviving id-full finding, so the 'walk past the id-less top-ranked member' behavior itself is unverified.

**Suggested fix:**
In both the Python and JS 'no truthy id' tests, construct the group so the id-less finding is the higher-priority (core-dimension) member and explicitly assert has_id['consolidation_primary'] is True (or the JS equivalent withId.consolidation_primary === true) — that is the only way to prove the implementation walks past the id-less top-ranked member rather than leaving no member stamped as primary.

scripts/post_review.py
```diff
@@ -1199,15 +1320,21 @@ def post_gitlab(data, valid_lines, new_files=None, old_paths=None):
     already_present = 0
     invalid = 0
     failed = 0
-    for filepath, f in remaining:
+    for filepath, group in remaining:
```
leehopper: **💡 [LOW] GitLab group_size counter accounting untested for posted/already_present/failed paths**

post_gitlab's inline-discussion loop was changed so posted, already_present, invalid, and failed all increment by group_size (1 + len(corroborators)) instead of 1, so a consolidation group's counters reflect every member, not just the primary. Only the 'invalid' (malformed position) branch has a test asserting the summed count for a group of 3; the posted, already_present, and failed branches have no test exercising group_size > 1, so a regression silently reverting any of those three sites to += 1 would go undetected.

**Suggested fix:**
Add GitLab dry-run/live tests with a primary+corroborator group that (a) succeeds (assert '2 inline discussion(s) posted/captured'), (b) is already delivered per delivered_keys (assert already_present reflects 2, not 1), and (c) is rejected by the GitLab API mock (assert failed reflects 2). Mirror the structure of the existing malformed-position group test.

scripts/post_review.py
```diff
@@ -1281,9 +1410,11 @@ def post_gitlab(data, valid_lines, new_files=None, old_paths=None):
                 f"Skipping finding '{f.get('title', '?')}' at {filepath}:{line} "
                 f"— GitLab rejected the inline discussion.\n{error}"
             )
-            failed += 1
+            # The whole group rides one discussion, so all of its findings —
+            # primary and corroborators alike — count toward the total.
+            failed += group_size
```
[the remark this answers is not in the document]
leehopper: Fixed. Both late-loss paths (malformed position on the primary, and a rejected group POST) now fall back to posting each corroborator as its OWN discussion — its own file/line position, body from the single-finding `render_comment_body`, through the same validate_position → POST → dedup(finding_key) → counter path any non-grouped finding takes. The primary still counts 1 toward `invalid`/`failed`; each corroborator is then counted individually as posted / invalid / failed / already_present on its own merits. The single-discussion delivery is extracted into one local helper so the group body and the fallback bodies cannot drift.

Tests in `tests/test_post_review.py`: `test_malformed_position_loss_counts_the_whole_group_not_just_the_primary` (updated to the new semantics), plus new `test_group_fallback_partial_position_failure`, `test_group_fallback_partial_post_failure` (asserts 1 rejected group attempt + N individual corroborator POSTs, and that the fallback bodies are single-finding renders), and `test_group_fallback_total_failure` (invalid + failed sums to group_size). Mutation-verified: reverting the fan-out to the old "add group_size and skip" turns all four red.

scripts/filter_findings.py
leehopper: Fixed. `_line_bucket_proximity` is deleted. `consolidate_cross_agent` now iterates `group_by_proximity(...).items()` and builds `consolidation_key` from the group's own `(file, bucket)` key, so `group_by_proximity` is the single definition of the bucketing formula. The JS twin already shares one `lineBucket` helper, so it needed no change and `consolidation_key` stays byte-identical across runtimes (parity fixtures unchanged, `tests/test_parity_fixtures.py` green).

leehopper: Fixed — same change as the sibling comment: there is no second formula any more. `consolidate_cross_agent` takes the `(file, bucket)` key straight from `group_by_proximity`'s returned dict and spells `consolidation_key` from it; `_line_bucket_proximity` is gone. Covered by the existing `TestConsolidateCrossAgent` / `group_by_proximity` tests in `tests/test_filter_findings.py` and by the frozen parity fixtures, which show zero diff.

scripts/filter_findings.py
leehopper: Fixed. The fixture in `test_no_truthy_id_passes_through_unstamped` (`tests/test_filter_findings.py`) and its JS twin (`workflows/test/consolidation.test.js`, "findings without a truthy id pass through unstamped") now give the id-LESS finding the higher priority (core dimension `bug` AND confidence 95, vs `test_coverage` / 50), so the primary walk must step past it. Both tests now assert `consolidation_primary` is true on the id-carrying finding and absent on the id-less one. Mutation-verified in both runtimes: replacing the walk with `ranked[0]` / `ranked[0]` turns each test red.

scripts/post_review.py
leehopper: Fixed. New tests in `tests/test_post_review.py`: `test_group_success_counts_every_member_as_posted` (one discussion, "2 inline discussion(s) posted.") and `test_group_already_delivered_counts_every_member` (marker already delivered → `already_present` == group_size == 2, zero POSTs). The API-rejected path is covered under the new fallback semantics from the medium-severity finding above: `test_group_fallback_partial_post_failure` and `test_group_fallback_total_failure` — a rejected group discussion charges the primary 1 `failed` and then accounts for each corroborator by its own individual outcome, rather than charging the whole group.

scripts/post_review.py
```diff
@@ -1281,9 +1400,62 @@ def post_gitlab(data, valid_lines, new_files=None, old_paths=None):
...
+        outcome = deliver(f, filepath, f["line"], render_group_body(f, corroborators))
+        if outcome in ("invalid", "failed"):
+            # The group's single discussion is lost, but its corroborators were
+            # never given a chance of their own — post each as its own
+            # discussion rather than dropping validated findings. The primary
+            # still counts 1 toward its own loss; each corroborator is counted
+            # on its own merits.
+            counters[outcome] += 1
+            for c in corroborators:
+                counters[deliver_corroborator(c)] += 1
```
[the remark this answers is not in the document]
leehopper: Fixed in 5f8cb13. Delivery keys are now member-level and independent of the shape that posts them: each key comes from the member's own anchor plus its single-finding render (`member_key` in `post_gitlab`), the group discussion body carries one `code-gauntlet-finding-key` marker per member, and `find_finding_markers`/`finding_keys_for_sha` collect every marker in a body rather than only the last. Before posting a group, all member keys are checked: all present -> whole group counts already_present with no POST; some present (a prior run's fallback landed part of it) -> the group is not posted and only the missing members are delivered individually; none present -> unchanged.

New tests: `test_group_rerun_after_full_delivery_posts_nothing`, `test_group_partial_prior_delivery_posts_only_missing_members`, `test_group_body_carries_a_marker_for_every_member` (tests/test_post_review.py), `test_every_marker_in_one_body_is_returned` (tests/test_review_marker.py), `test_finding_keys_for_sha_collects_every_key_in_one_body` (tests/test_detect_prior_review.py); `test_group_already_delivered_counts_every_member` updated to member keys.

scripts/post_review.py
```diff
@@ -1281,9 +1417,111 @@ def post_gitlab(data, valid_lines, new_files=None, old_paths=None):
...
+                counters[
+                    "already_present"
+                    if anchor and member_key(c, *anchor) in delivered_keys
+                    else deliver_corroborator(c)
+                ] += 1
             continue
-        posted += 1
+        outcome = deliver(
+            f, filepath, f["line"], render_group_body(f, corroborators), member_keys
+        )
```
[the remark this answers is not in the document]
leehopper: Fixed in #208: unanchored corroborators (no line, or a line outside the diff) now get a delivery key derived from the same material their group-body marker uses, and — when partially delivered by an earlier fallback run — are posted as a position-less MR note instead of being silently absorbed into an `already_present` count. See PR description for the full mechanism and mutation-verification evidence.

PR #217, opened 2026-08-19
feat(workflows): per-dimension required mechanism in the finding schema

### LINKED ISSUE #66: No per-dimension required mechanism in the finding schema
## Problem / Goal

`findingItemSchema` (`workflows/src/stages.js`) always emits `required: FINDING_REQUIRED` — one flat list shared by every dimension's discovery dispatch **and** by the verify echo, which carries all agents' findings mixed together post-merge. There is no way to mark a field required for one dimension only.

So four fields the agent contracts call REQUIRED are enforced by prose alone:

| Field | Dimension | Contract text |
|---|---|---|
| `claude_md_rule` | convention | "MUST be non-null … Findings without a cited rule will be rejected" |
| `spec_text` | intent | "MUST be non-null … Findings without a cited spec will be rejected" |
| `criticality` | test_coverage | emitted unconditionally, no omit branch |
| `failure_scenario` | test_coverage | emitted unconditionally, no omit branch |

#47 made this explicit in `registry.js`'s `FINDING_REQUIRED` comment and in CLAUDE.md rather than faking it — appending a single-dimension field to the shared list would reject every finding from the other eight dimensions, including at the verify echo.

## Requirements

1. A per-dimension `requiredExtra` on the `DIMENSIONS` row, threaded into `findingItemSchema`'s `required` for the **discovery** dispatch only.
2. The verify echo must keep the flat list: it transcribes findings from every agent at once, so a per-dimension requirement there is unsatisfiable by construction.
3. Weigh against retry cost before shipping: a required field the model cannot supply is a retry storm — the same failure class the omit-not-null rule exists to prevent. A field that is required *and* genuinely conditional should stay prompt-enforced.

## Evidence

`workflows/src/stages.js` — `findingItemSchema` returns `{ type: 'object', properties: props, required: FINDING_REQUIRED }` with no dimension input to `required`.
`workflows/src/registry.js` — the `FINDING_REQUIRED` comment records this constraint and the instruction not to work around it.

leehopper:
Closes #66.

## What

`findingItemSchema` always emitted the flat `FINDING_REQUIRED`, so every field an agent contract called REQUIRED was enforced by prose alone. This adds the per-dimension mechanism the `FINDING_REQUIRED` comment tracked:

- Every `DIMENSIONS` row carries `requiredExtra` (registry.js), the subset of that row's own `schemaExtra` its contract emits unconditionally.
- `agentSpecs(dims = DIMENSIONS)` computes the per-agent effective list as the **intersection** across the agent's rows and `findingItemSchema` appends it to `FINDING_REQUIRED` — for the discovery dispatch only, the one boundary that carries finding items by value (requirement 2 is satisfied by construction: verify echoes per-id deltas, never findings).
- Agents with an empty effective list (`bug-detector`, `conventions-and-intent`, `type-design-analyzer`) dispatch the byte-identical flat list they always did, pinned by test.

## Promotion judgment (requirement 3)

A full quote-backed audit of the seven contracts classified every extra plus `claude_md_rule`:

| Field | Dimension | Verdict | Why |
|---|---|---|---|
| `attack_vector` | security | **promoted** | no OMIT branch; the contract's own exclusion list already bars findings without a concrete vector |
| `affected_consumers` | cross_file_impact | **promoted** | no OMIT branch; `required` forces key presence, not a non-empty list |
| `criticality`, `failure_scenario` | test_coverage | **promoted** | rating fields assigned to every finding, no omit branch (the issue's two named unconditional fields) |
| `behavior_preserved` | simplification | **promoted** | contract supplies a sanctioned fallback (`'uncertain'`) instead of ever permitting omission |
| `claude_md_rule`, `spec_text` | convention / intent | **stays contract-enforced** | required for ONE dimension of the conventions-and-intent dispatch, which mixes three dimensions in one schema — unsatisfiable per-dimension there |
| `hidden_errors`, `invalid_state_example` | bug / type_design | **stays contract-enforced** | explicit OMIT branches; schema-requiring them forces fabrication (the retry-storm class the omit-not-null rule exists to prevent) |

Conditional schema enforcement was measured, not assumed (2026-08-18, CLI 2.1.235, first-party API): a top-level `oneOf`/`allOf`/`anyOf` in input_schema is rejected outright (API 400); `if`/`then` nested inside `items` was accepted AND enforced by the retry loop (first attempt omitting the conditional field was rejected with `must match "then" schema` and retried). It stays out of this PR because it is unmeasured on Bedrock/Vertex — the same provider-400 class as #179 — and one such 400 would degrade the agent's whole dimension set. Follow-up filed as an option.

## Drift guards (both runtimes)

- `requiredExtra` must be a subset of the row's **own** `schemaExtra` (a field every dimension emits belongs in `FINDING_REQUIRED`, not per-dimension promotion) and disjoint from `FINDING_REQUIRED`.
- Sibling-parity: rows sharing an agentType must declare identical `requiredExtra` sets — the intersection is the fail-safe algebra, the parity test the authority, so declared-but-unenforced fails the build.
- OMIT-drift guard: a promoted field whose owning contract (re)gains an OMIT branch fails the build; the value parser handles every contract value shape and **raises** on an unrecognized one instead of silently skipping.
- Bidirectional contract-sentence lockstep on the canonical phrase "required by the dispatch schema": every promoted field's contract must state it; no unpromoted field may claim it.
- report-format.md's per-dimension table gained a Required column, locked to the registry.

## Review and verification

Two adversarial rounds (design/tests/docs lenses). The review round mutation-proved two real holes that were then fixed: the OMIT guard was vacuous for the array-valued `affected_consumers`, and replacing the `agentSpecs` intersection call site with first-row-verbatim left the whole suite green (closed by parameterizing `agentSpecs(dims)` plus an end-to-end synthetic-rows test).

Ten mechanism mutations verified red across the two rounds (threading dropped, union-instead-of-intersection, row emptied, Python bridge stops serializing, Required cell flipped, array-shape OMIT injection, call-site first-row-verbatim, contract sentence deleted, bogus sentence added, canonical-field promotion) — each restored to green.

Gates: tests/ 1420 passed (coverage 93.18 ≥ 92.1), bench/tests 537 passed (88.48 ≥ 87), JS 836/836 (98.92/86.65/98.42 ≥ 98/85.7/97.4; functions headroom is 1.0pp at the one-decimal rounding the last ratchet used, so floors unchanged), bundle byte-fresh, pre-commit all hooks green.

## Measurement

Suites-only per the wave rule. Recall-relevant delta for the Wave 5 wave-end paired mini: five fields are now schema-required at their dispatch, so a finding omitting one burns platform schema retries (cap 5) and persistent non-compliance would degrade that agent's dimensions — pre-#47 evidence says models emit these fields spontaneously, and each promoted field is one the contract already made mandatory, so the expected live delta is enforcement of already-instructed behavior, not new rejection.

https://claude.ai/code/session_01CeXNjP9RtZpvE28LCXzmdR

---

> [!NOTE]
> **Medium Risk**
> Changes StructuredOutput validation for four discovery agents—omitting a promoted field can exhaust retries and degrade those dimensions—but behavior matches existing contract requirements and multi-dimension safety is guarded by intersection plus sibling-parity tests.
> 
> **Overview**
> Discovery finding schemas used a single flat `FINDING_REQUIRED`, so dimension-specific fields (e.g. `attack_vector`, `criticality`) were only enforced in agent prose. This PR adds **`requiredExtra`** on each `DIMENSIONS` row and appends the agent’s effective list to `required` in `findingItemSchema` — **`FINDING_REQUIRED` plus `requiredExtra`**, not a global bump.
> 
> **Promoted** (unconditional in contract): `attack_vector`, `affected_consumers`, `criticality`, `failure_scenario`, `behavior_preserved`. **Not promoted**: fields with OMIT branches or mixed-dimension agents (`claude_md_rule`, `spec_text`, etc.). Multi-dimension agents use **`intersectRequiredExtra`** so a field is required only if every sibling row agrees; `agentSpecs(dims)` is parameterized for tests.
> 
> Agent contracts gain canonical “required by the dispatch schema” sentences; **report-format.md** adds a per-dimension **Required** column. **Python/JS tests** guard registry↔contract drift (OMIT vs required, dispatch sentence lockstep, ReDoS-safe OMIT parsing).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eda2aa4269356e82fe1cf550baba1fed1f1bc745. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_dimensions_registry.py
+++ b/tests/test_dimensions_registry.py
@@ -205,7 +205,12 @@ def field_carries_omit_instruction(raw_block, field, source=None):
     if quoted:
         return "omit" in quoted.group(1).lower()
 
-    array_of_strings = re.match(r'\[\s*(?:"(?:[^"\\]|\\.)*"\s*,?\s*)*\]', rest)
+    # First-string-then-comma-separated-rest, not a single starred group with an optional
+    # comma: `(?:"…"\s*,?\s*)*` lets the engine partition inter-string whitespace ambiguously
+    # and backtrack exponentially on an unclosed array (CodeQL py/redos on PR #217).
+    array_of_strings = re.match(
+        r'\[\s*(?:"(?:[^"\\]|\\.)*"(?:\s*,\s*"(?:[^"\\]|\\.)*")*\s*)?\]', rest
+    )
     if array_of_strings:
         return "omit" in array_of_strings.group(0).lower()
 
--- a/tests/test_dimensions_registry.py
+++ b/tests/test_dimensions_registry.py
@@ -119,6 +119,7 @@ def registry():
             "  dimensions: m.DIMENSIONS.map(d => ({"
             "    dimension: d.dimension, agentType: d.agentType,"
             "    extras: Object.fromEntries(Object.entries(d.schemaExtra || {}).map(([k, v]) => [k, t(v)])),"
+            "    requiredExtra: d.requiredExtra || [],"
             "  })),"
             "})))"
         )
@@ -162,6 +163,89 @@ def all_extras():
         for field in row["extras"]:
             owner[field] = row["dimension"]
     return owner
+
+
+def all_required_extras():
+    """The set of per-dimension fields that are dispatch-required somewhere (issue #66).
+
+    Every field with a non-empty requiredExtra currently lives on a single-dimension agent
+    (security, cross_file_impact, test_coverage, simplification), so the row-level set and the
+    post-intersection agentSpecs()-effective set agree; this helper reads the row-level
+    declaration directly, which is what report-format.md's per-field table documents.
+    """
+    required = set()
+    for row in registry()["dimensions"]:
+        required |= set(row["requiredExtra"])
+    return required
+
+
+def field_carries_omit_instruction(raw_block, field, source=None):
+    """True if `field`'s own value inside one raw ```json contract block mentions OMIT.
+
+    Scoped to the field's OWN value (not the whole block) so an OMIT instruction on a
+    neighbouring field in the same block — e.g. claude_md_rule sitting beside attack_vector —
+    can never false-positive this guard. Recognizes every value shape a contract block
+    actually uses: a quoted string placeholder (`"field": "<...>"`), an array of quoted
+    strings (`"field": ["<...>"]`, the affected_consumers shape — issue #66 promoted an
+    ARRAY field and the original version of this guard only handled a scalar, silently
+    never inspecting it), an unquoted angle-bracket placeholder (`"field": <1-10>`), and a
+    bare JSON literal (`"field": 9`, the criticality worked example).
+
+    A field present in the block whose value matches NONE of those shapes raises, naming
+    `source` (the file) and `field`, rather than returning False — this file's own
+    convention (see the module docstring): a silent skip is the one failure mode that would
+    let this guard quietly stop guarding.
+    """
+    key_match = re.compile(r'"' + re.escape(field) + r'"\s*:\s*').search(raw_block)
+    if not key_match:
+        return False
+    rest = raw_block[key_match.end() :]
+
+    quoted = re.match(r'"((?:[^"\\]|\\.)*)"', rest)
+    if quoted:
+        return "omit" in quoted.group(1).lower()
+
+    array_of_strings = re.match(r'\[\s*(?:"(?:[^"\\]|\\.)*"\s*,?\s*)*\]', rest)
+    if array_of_strings:
+        return "omit" in array_of_strings.group(0).lower()
+
+    angle_bracket = re.match(r"<[^>]*>", rest)
+    if angle_bracket:
+        return "omit" in angle_bracket.group(0).lower()
+
+    bare_literal = re.match(r"(?:-?\d+(?:\.\d+)?|true|false|null)\s*(?=[,}])", rest)
+    if bare_literal:
+        return False  # a bare number/bool/null cannot carry OMIT prose by construction
+
+    raise AssertionError(
+        f"field_carries_omit_instruction: {field!r} in "
+        f"{source or 'a contract block'} has a value shape this parser does not recognize "
+        "(not a quoted string, array-of-strings, angle-bracket placeholder, or bare "
+        f"literal) — extend the parser rather than silently return False: {rest[:80]!r}"
+    )
+
+
+# The exact canonical phrase every requiredExtra field's owning contract must use (F5, issue
+# #66) — pinned literally so a paraphrase silently stops matching this guard the same way a
+# grep for "suggestion" would (see the module docstring's warning against prose-frequency
+# checks). Matched per LINE, not per file, so the backticked field name(s) attributed to the
+# claim are exactly the ones sharing that sentence — a phrase on one line making no claim
+# about a field named three paragraphs away must not count as a claim about it.
+_DISPATCH_REQUIRED_PHRASE = "required by the dispatch schema"
+
+
+def dispatch_required_claims(name):
+    """Field names agents/<name>.md's prose claims are 'required by the dispatch schema'.
+
+    Scans line by line: a line containing the canonical phrase contributes every backticked
+    field name ON THAT LINE to the claimed set.
+    """
+    text = (REPO / "agents" / f"{name}.md").read_text()
+    claimed = set()
+    for line in text.splitlines():
+        if _DISPATCH_REQUIRED_PHRASE in line:
+            claimed |= set(_BACKTICKED_FIELD.findall(line))
+    return claimed
 
 
 def raw_contract_blocks(name):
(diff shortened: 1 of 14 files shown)


tests/test_dimensions_registry.py
```diff
@@ -547,6 +672,125 @@ def test_criticality_is_declared_as_a_number(self):
...
+        self.assertFalse(
+            field_carries_omit_instruction(
+                '{"affected_consumers": ["<file paths>"]}', "affected_consumers"
+            )
+        )
+        # Unquoted angle-bracket placeholder (criticality's <1-10> template form).
+        self.assertFalse(
+            field_carries_omit_instruction('{"criticality": <1-10>}', "criticality")
+        )
+        self.assertTrue(
```
leehopper, line 737: **💡 [LOW] Test comment says retries are infinite, contradicting this PR's own 5-retry-cap documentation**

The new test_required_extra_fields_carry_no_omit_instruction comment claims a contradictory OMIT branch "produces a schema violation the platform retries forever," but this same PR's registry.js/stages.js comments explicitly document a capped retry count of 5 followed by a terminal agent failure, not an infinite retry loop.

**Suggested fix:**
Update the test comment to say retries are capped (5) and then fail terminally, matching the accurate description already added to registry.js/stages.js in this same PR, instead of "retries forever."

tests/test_dimensions_registry.py
```diff
@@ -164,6 +165,94 @@ def all_extras():
...
+    quoted = re.match(r'"((?:[^"\\]|\\.)*)"', rest)
+    if quoted:
+        return "omit" in quoted.group(1).lower()
+
+    # First-string-then-comma-separated-rest, not a single starred group with an optional
+    # comma: `(?:"…"\s*,?\s*)*` lets the engine partition inter-string whitespace ambiguously
+    # and backtrack exponentially on an unclosed array (CodeQL py/redos on PR #217).
+    array_of_strings = re.match(
+        r'\[\s*(?:"(?:[^"\\]|\\.)*"(?:\s*,\s*"(?:[^"\\]|\\.)*")*\s*)?\]', rest
+    )
```
leehopper, line 214: **💡 [LOW] No regression test proves the rewritten array-of-strings OMIT regex resists catastrophic backtracking**

The array_of_strings pattern inside field_carries_omit_instruction was rewritten specifically to fix a CodeQL py/redos finding (PR #217) on an unclosed-array input, per the inline comment, but the only test exercising this code path (test_field_carries_omit_instruction_parses_every_real_value_shape) checks return-value correctness on well-formed and malformed inputs, never timing on a pathological unclosed-array string. A revert to the vulnerable single-starred-group pattern the comment describes would still pass every existing assertion because the two patterns match identically on all well-formed test fixtures; only a pathological (long, unclosed) input distinguishes them, and nothing in the suite constructs one.

**Suggested fix:**
Add a timing-based regression test, mirroring entry_guard.test.js's ReDoS guard: construct a pathological unclosed-array string (e.g. '[' + '"a",'*5000) and assert field_carries_omit_instruction returns within a small time bound (e.g. under 1 second), so a future revert to the vulnerable regex form fails loudly instead of silently reintroducing exponential backtracking.

**Cited rule:**
> AGENTS.md: 'A regression test must fail against the bug it names. Verify that by mutating the implementation and watching it go red — not by reading the test. Mutate the whole mechanism; a partial mutation falls through to a neighbouring fallback and passes misleadingly.'

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: 9e463de38dd92de487e8588828020279cf7167e7

tests/test_dimensions_registry.py
leehopper: Fixed in eda2aa4: the comment now states the capped retries (5) and terminal degradation, matching registry.js/stages.js.

tests/test_dimensions_registry.py
leehopper: Fixed in eda2aa4: added test_field_carries_omit_instruction_array_parse_is_linear_time (unclosed whitespace-separated array, 1s bound). Mutation-verified: the reverted vulnerable pattern hangs past a 20s timeout; the shipped form returns in under a millisecond. Note the trigger had to be comma-less — comma-separated input parses one way and cannot distinguish the two regex forms.

tests/test_dimensions_registry.py
```diff
@@ -164,6 +165,89 @@
...
+    key_match = re.compile(r'"' + re.escape(field) + r'"\s*:\s*').search(raw_block)
+    if not key_match:
+        return False
+    rest = raw_block[key_match.end() :]
+
+    quoted = re.match(r'"((?:[^"\\]|\\.)*)"', rest)
+    if quoted:
+        return "omit" in quoted.group(1).lower()
+
+    array_of_strings = re.match(r'\[\s*(?:"(?:[^"\\]|\\.)*"\s*,?\s*)*\]', rest)
```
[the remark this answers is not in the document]
leehopper: Fixed in 9e463de (first-string-then-comma-separated-rest form); the CodeQL alert is closed and the check passes.

Issue #220, opened 2026-08-19
Report-side apply-check: REVIEW.md renders suggested_fix_code blocks no check has vetted


leehopper:
## Problem

After #63, delivery runs a deterministic apply-check before any one-click `suggestion` fence ships, but the REVIEW.md report template renders `suggested_fix_code` as a plain labeled block (`Proposed replacement for {file}:{line_start}-{line_end} (not apply-checked):`) with no verification at all. In a local-diff review (no PR target) the report is the only rendered surface, so a stale, no-op, or mis-ranged patch reaches the reader unflagged — labeled as unchecked, but present.

## Fix shape

The gate's checks that need only the diff (range-in-diff, no-op against the new-side text, indentation charset, size, CR, end_line presence) are pure functions of data the pipeline already holds at report time. A report-side variant could annotate or drop failing blocks the same way delivery does, and the label could then claim "apply-checked" honestly. Needs a decision on where it runs (report-writer contract vs a pipeline stage) since the report renderer is a model, not a script.

Deliberately excluded from #63 (design decision D9): the report block is not a one-click affordance, so the risk is lower than delivery's was — the #63 PR demoted the template from a `suggestion` fence to a labeled plain block as the interim fix.

Surfaced by the #63 round-1 adversarial review.

Issue #219, opened 2026-08-19
GitLab multi-line suggestions: emit `suggestion:-m+n` so multi-line patches can ship instead of downgrading


leehopper:
## Problem

The #63 apply-check downgrades every multi-line `suggested_fix_code` on GitLab to prose (`anchor_mismatch`), because `post_review.py` only ever emits a plain ```` ```suggestion ```` fence and GitLab applies that to the single anchored line. GitLab has native syntax for wider ranges: ```` ```suggestion:-m+n ```` replaces m lines above through n lines below the anchor.

## Fix shape

Teach the GitLab delivery path to emit the offset syntax when the finding's `[line, end_line]` spans more than one line, and extend the apply-check's `anchor_mismatch` computation so the fence's effective apply range (anchor + offsets) is what gets compared against the stated range. The offset math must be tied to the actual `position.new_line` the discussion is anchored to, and the render stays platform-blind — the offsets would need to be decided where the platform is known, which is a deliberate design decision, not a one-liner.

Until then the behavior is safe but conservative: multi-line patches on GitLab deliver as prose.

Surfaced by the #63 implementation (apply-check design decision D9).

Issue #212, opened 2026-08-18
suggested_fix_code reaches a one-click-apply suggestion fence gated by prose only


leehopper:
## Problem

`scripts/post_review.py` (~805-812) renders `suggested_fix_code` into a GitHub ```suggestion
fence — one-click-applicable by a reviewer — with structural sanitize deliberately OFF (redaction
only), and neither filter twin scans the field. Its safety guarantee is prose: "caller-supplied
only; no review-pipeline agent emits it" (delivery-guide.md:61, report-format.md:316,
phase8-delivery.md:126). But `scripts/merge_findings.py` validates only `REQUIRED_FIELDS`
presence with no field allowlist, so an agent that did emit the key would have it passed straight
through to the wire.

## Fix shape

Build the mechanism, not the instruction: drop non-registry delivery-side keys from
agent-originated findings at the merge boundary, making the "no agent emits it" comment true by
construction. This is the ungated twin of the exact channel #62 closed for `suggestion` (#210).

leehopper:
PR #221 (#63) changes this issue's premise. `suggested_fix_code` is now registry-declared (`FINDING_PROP_TYPES`), instructed in all 7 discovery contracts, deterministically gated at delivery (12-reason apply-check, fail-closed), non-string/oversize/propagation-stripped in both filter twins, and dropped at the discover intake whenever its stated range is lost. The "Delivery-side fields" table this issue's fix shape keyed on is gone — there is no undeclared delivery-side field left for a merge-boundary allowlist to drop.

What remains of this issue is only the general mechanism question: whether `merge_findings.py` should drop unknown keys as a guard for FUTURE delivery-side fields. Recommend closing as superseded by #221 unless that generic guard is wanted on its own.

leehopper:
Closing as superseded by #63 / PR #221 (owner-directed close-out sweep, 2026-08-19). The channel this issue named is gone: `suggested_fix_code` is now declared in `FINDING_PROP_TYPES`, instructed in all 7 contracts, and every ​```suggestion fence passes a deterministic 12-reason fail-closed apply-check at delivery — the prose-only guarantee this issue objected to no longer exists, and the delivery-side field class is empty. The generic fix shape proposed here (drop non-registry keys at the merge boundary) was deliberately not built: with no undeclared delivery-side field left, it would guard nothing today. If a future field reintroduces the class, reopen or file fresh against that field.

PR #205, opened 2026-08-17
fix(workflows): bound implausible discovery-agent line spans at intake

leehopper:
## Summary

- A 2026-08-17 live run had a discovery agent emit `line_start=68, line_end=940` — an ~872-line span crossing hunks. Delivery-side, this made `post_review.py` reject the whole GitHub review POST with a 422 (fixed separately in PR #200); this PR closes the matching emission-side gap so an implausible span never leaves discovery in the first place.
- `discover()` (`workflows/src/stages.js`) now normalizes each finding's `line_end` right after StructuredOutput results are collected, before anything reaches `mergeStage`. `line_end` is **dropped, never clamped**, when it is not an integer, is before `line_start`, or spans more than `limits.maxLineSpan` (new, default 100) lines past `line_start`. A clamped value would be a fabricated line number nobody reported; the finding survives, anchored at its still-valid `line_start`, exactly like a finding that never had an end line.
- `maxLineSpan` joins `LIMIT_DEFAULTS` (`workflows/src/args.js`) as a fifth caller-tunable limit, consumed via the existing `effectiveX`-accessor pattern (`effectiveMaxLineSpan`), and gets the same positive-safe-integer shape check the other bucket/batch/slice sizes already have. A codebase with legitimately large multi-line findings can raise it.
- No silent drops: `discover()` pushes a gap line disclosing the dropped count through the stage's existing `gaps` disclosure channel (the same channel the report/mergeStage already consume) — no new stats object was needed.
- Rebuilt `workflows/pipeline.js` from `workflows/src/*.js` and committed the bundle per `workflows/AGENTS.md`.
- Did not touch the Python/JS parity twins, their golden fixtures, or `agents/*.md` contracts — the drop happens before merge, so their inputs are unchanged. `post_gitlab` never reads `end_line` at all (confirmed via `grep`), so this PR's scope stays entirely on the discovery/GitHub side, matching #204.

## Verification

- `node --test workflows/test/*.test.js` → 803 passed
- JS coverage gate (`--test-coverage-lines=98 --test-coverage-branches=83.3 --test-coverage-functions=97.2`) → measured 98.89% / 86.47% / 98.38%, presence check exit 0
- `node workflows/build.js && git diff --exit-code workflows/pipeline.js` → clean (bundle committed matches a fresh build)
- `uv run python -m pytest tests/ -q` → 1324 passed, 513 subtests passed (boundary-parity suites stayed green untouched)
- `pre-commit run --all-files` → all hooks passed

### Mutation table

[...]

PR #221, opened 2026-08-19
feat(delivery): suggested_fix_code end-to-end, gated on a deterministic apply-check

### LINKED ISSUE #63: Implement suggested_fix_code end-to-end, gated on a deterministic apply-check
## Problem / Goal

`scripts/post_review.py:386-392` renders `suggested_fix_code` as a committable GitHub/GitLab `suggestion` block — a **one-click apply** button — and has dedicated test coverage. But no agent emits the field and no schema declares it, so the review pipeline never populates it. #47 corrected the documentation (it now lives under "Delivery-side fields — not produced by the review pipeline") rather than implementing it.

## Why it was deferred rather than implemented

The missing half is not the schema entry — that is one line in `registry.js` plus a clause in 7 contracts. The missing half is a **deterministic guard**. A `suggestion` fence writes the model's guess straight into the author's branch when clicked. Nothing in the pipeline checks that the replacement applies at the stated `file:line_start-line_end`, that the range still exists at the posted SHA, that indentation matches the replaced span, or that the block is non-degenerate. A wrong one-click patch is worse than no patch, and the smoke tier would not measure patch correctness.

## Requirements

1. A deterministic pre-render check in `post_review.py`, run before any `suggestion` fence is emitted:
   - the file exists at the posted SHA and `line_start..line_end` is inside it,
   - the range is inside the diff (the poster already resolves the diff for line validation),
   - the replacement is non-degenerate (#47 already made whitespace-only render nothing),
   - ideally: leading indentation is consistent with the replaced span.
   A finding that fails the check falls back to the prose `suggestion` and records why.
2. Only then: declare `suggested_fix_code` in `registry.js` and instruct it in the discovery contracts — with an OMIT-not-null clause, per the schema's nullability rule.
3. Move it out of report-format.md's "Delivery-side fields" table into the canonical table (`tests/test_dimensions_registry.py::TestReportFormatFieldTables` enforces that the two agree).
4. Measurement: patch-acceptance rate, not just recall/noise.

## Evidence

```
$ grep -rn "suggested_fix_code" agents/ workflows/src/
(no output)

$ sed -n '386,392p' scripts/post_review.py
    if suggested_fix:
        parts += [
            "",
            "```suggestion",
            suggested_fix,
            "```",
        ]
```

leehopper:
Closes #63.

`scripts/post_review.py` has rendered `suggested_fix_code` as a committable ```suggestion fence (one-click apply) since #47, but only for caller-supplied JSON — no agent emitted it and no schema declared it, because nothing checked that the patch actually applies at the stated range. This PR builds that check and then opens the field to the pipeline.

## The gate

A fence renders if and only if the delivered comment's one-click apply range exactly equals the finding's stated `[line, end_line]` at that render site, and every content check passes. Everything else downgrades to the prose `suggestion` on the same finding — the finding always ships; only the fence is conditional.

`_suggested_fix_gate` is a pure function (no I/O, no subprocess) with a closed 12-reason vocabulary, first failure wins: `non_string`, `empty`, `carriage_return`, `redacted`, `missing_end_line`, `invalid_range`, `no_diff_oracle`, `range_not_in_diff`, `anchor_mismatch`, `no_op_replacement`, `indentation_mismatch`, `replacement_too_large`. Each downgrade lands in stderr and the dry-run payload's `skipped` list as `suggested-fix downgraded: {file}:{line} ({reason})`.

Design points worth knowing:

- **The content oracle is the diff itself, not `git show`.** `parse_diff_lines` already reads every new-side line's text and used to discard it; it now returns a parallel `line_texts` dict. That text is by construction what the platform's anchor points at, so the no-op and indentation checks are checkout-state-independent — `git show` would silently compare against the wrong blob whenever local HEAD is not the PR head, and fail entirely in shallow clones.
- **The gate fails closed.** A missing or partial diff oracle (failed `gh pr diff`, mixed path spellings across a span) is `no_diff_oracle`, never a pass-through. The anchor validation keeps its historical fail-open (a wrong anchor costs a misplaced comment); the patch cannot (a wrong patch costs corrupted code, and the prose fallback is free). The posters' oracle parameters lost their `=None` defaults so the gate cannot be half-disabled by omission.
- **The gate runs at every render site, keyed to that site's real apply range.** GitHub group comments carry only the primary's fence at the primary's range; GitLab anchors are single-line and this script never emits `suggestion:-m+n`, so a multi-line span on GitLab downgrades; position-less notes and the #192 body section can never apply a fence and always strip; corroborators delivered on their own anchor gate against that anchor.
- **Delivery keys are fence-independent.** `member_key`/`finding_key` render a copy with the field stripped unconditionally, so keys are byte-equal across grouped vs individual delivery, across gate outcomes, and against pre-#63 runs — the #132/#208 dedup and every standing marker survive this PR.
- **Redaction is gated, not applied.** A patch that redaction would rewrite downgrades (`redacted`) instead of shipping a literal `[REDACTED]` into the author's branch; passing the gate proves the render-time redaction is a no-op, so posted bytes equal checked bytes.
- **Interior carriage returns fail closed** (`carriage_return`): CommonMark treats a lone `\r` as a line ending, so CR-bearing content is a different document to the renderer than to the gate's measurements — and CRLF-file patches are exactly the ambiguity a one-click apply must not ship.
- **Edge blank lines are content.** A dedicated normalizer strips exactly one trailing newline (the terminator) and nothing else, used by both the gate and the fence render, so stated == checked == applied. The shared `_rendered_text` would have silently truncated a patch's leading/trailing blank lines.
- **The gate runs identically under `--dry-run`** (the `validate_position` precedent), and caller-supplied JSON goes through the same path — this is a deliberate behavior change: previously unvalidated caller fences now downgrade unless the checks pass.

## Upstream mechanism

- `workflows/src/registry.js`: `suggested_fix_code: 'string'` in `FINDING_PROP_TYPES` — optional, never required (it has a legitimate OMIT branch, which per the #217 rule excludes it from `requiredExtra`).
- All 7 discovery contracts instruct it with a shared OMIT-not-null clause: byte-exact drop-in replacement for exactly `line_start..line_end`, emitting it REQUIRES `line_end`, omit unless certain.
- `workflows/src/stages.js` discover intake: dropping an implausible `line_end` (the #205 drop-not-clamp) now deletes `suggested_fix_code` in the same breath — the stated range was just destroyed — and a finding that emits the field with no `line_end` at all loses it there too, each with its own gaps message. A patch that cannot state its range never travels six stages to a guaranteed downgrade.
- Filter twins (both runtimes, #210 mechanism): non-string strip, oversize strip (same bounds as delivery), and propagation — a finding whose `suggestion` was stripped by an injection match also loses its patch. Deliberately NO pattern scan of the code content: #62 measured the content sets killing 10/10 legitimate security remediations, and code trips them harder than prose. New parity golden `injection/suggested_fix_code_scan` (FX1 non-string, FX2 line bound, FX3 propagation, FX4 benign kept, FX5 an 8000-code-point astral boundary case pinning JS code-point counting against UTF-16 `.length`). All pre-existing goldens byte-identical.
- The three homes of `_FIX_MAX_LINES = 100` / `_FIX_MAX_CHARS = 8000` (delivery gate + both filter twins) share one measurement definition (normalize, then lines = `split("\n")`, chars = code points) and a lockstep test that regex-parses all three files and asserts equality.

## Docs

`suggested_fix_code` moved from report-format.md's "Delivery-side fields" table (now deleted — it has no members left) into the canonical table; the "no agent emits it / caller-supplied only" claims across delivery-guide.md, phase8-delivery.md, and the module docstring now describe the gated path. The Full Report Template renders the field as a plain labeled block (`Proposed replacement for {file}:{line_start}-{line_end} (not apply-checked):`) instead of a ```suggestion fence — a fence pasted from REVIEW.md into a PR comment would be a one-click apply that never passed the gate; the Inline PR Comment Format keeps the fence because that path is gated.

## Measurement

Suites-only per the Wave 5 rule. For the wave-end Smoke B → paired mini:

- Recall-relevant deltas: all 7 discovery contracts gain one optional field (prompt surface at discovery); fence text now enters bench's scored candidate text (the benchmark scores captured dry-run comment bodies, and its fixtures already carry the field); no elimination-path change anywhere — the gate and the filter strips only ever swap fence↔prose within a finding that still delivers.
- Patch-acceptance rate (requirement 4) is now a deterministic $0 per-run readout: `N suggested fix(es) passed the apply-check. / M suggested fix(es) downgraded to prose.` printed whenever any finding carried the field, with per-finding reasons in stderr and the dry-run payload. The readout reports gate verdicts at render sites, deliberately not delivery outcomes (those remain the per-platform count lines' business).
- Proposed Smoke B watch item: passed + downgraded == gate evaluations, and downgrade reasons stay within the closed vocabulary. One accounting caveat: a finding that carries the field but ships as a corroborator inside a group body renders no fence and appears in neither counter (`_render_corroboration` emits no fence by construction).
- `bench` note: downgrade warnings ride the dry-run payload's existing `skipped` list, so the adapter's `n_skipped` now counts delivery-side skip warnings in addition to unanchored findings. Candidate extraction is unaffected.
- Coverage floors: Python scripts gate raised 92.1 → 92.4 (this PR measures 93.37–93.41 locally; policy threshold). JS branches/functions sit at ~1.0pp local headroom — borderline, left for the CI measurement to decide a follow-up ratchet.

## Review record

Design red-teamed by two independent adversarial reviews before implementation (both returned REVISE; the adopted corrections include the diff-as-oracle design, the explicit-`line_end` requirement, and fence-independent delivery keys). Two post-implementation adversarial rounds with fresh reviewers: round 1 produced 10 adjudicated findings (fail-closed oracle, edge-blank-line normalizer, truthful counter labels, the stages missing-`line_end` strip, bound-definition alignment, JS code-point parity, report-template fence demotion, the constants lockstep test); round 2 produced 3 (interior CR, partial-oracle fail-open, silent patch discard in the implausible-span gap message); a final verification round confirmed all fixes with a clean spot-review. Every mechanism was mutation-verified one mutation at a time with sha256-checked restores across five rounds (work-package level: 23 + 3 + 11; round-1 fix cycle: 4 + 4, re-verified 5; round 2: 3, re-verified 3), each red in the runtime it targets.

## Not in this PR

- GitLab `suggestion:-m+n` multi-line syntax — multi-line spans on GitLab downgrade to prose instead (#219).
- A report-side apply-check for REVIEW.md's patch blocks (#220); the template's block is labeled not-apply-checked.
- Content scanning of `suggested_fix_code` (documented decision; #62 measurement).
- #212's general merge-boundary allowlist: this PR changes its premise — the field is now registry-declared and the delivery-side table is gone (comment left on the issue).

## Gates (local, final tree)

- `pytest tests/ -q` → 1513 passed, 634 subtests; `pytest bench/tests/ -q` → 537 passed, 22 subtests.
- Coverage: scripts+.github 93.41% vs floor 92.4; bench 88.48% vs 87; JS 98.94 / 86.79 / 98.43 vs 98 / 85.7 / 97.4 plus the presence check.
- `node --test workflows/test/*.test.js` → 855 pass; bundle rebuild byte-identical; `sync_agent_rules.py` twins current; `pre-commit run --all-files` clean.

https://claude.ai/code/session_017Yh6Xs3vNyJdy9vWUUoRVs

---

> [!NOTE]
> **Medium Risk**
> Changes one-click PR apply behavior and delivery JSON shape; mistakes in the gate or oracle could strip valid patches or, if bypassed, commit bad replacements—mitigated by fail-closed oracle handling, extensive tests, and downgrade-to-prose fallback.
> 
> **Overview**
> **End-to-end `suggested_fix_code`:** discovery agents and the registry can emit byte-exact replacement patches; delivery only renders them as committable ```suggestion fences after a pure apply-check, otherwise the finding keeps shipping with prose `suggestion` only.
> 
> **`post_review.py`** adds `_suggested_fix_gate` / `_gated_finding` (closed reason vocabulary, fail-closed when diff oracle is missing), extends `parse_diff_lines` with parallel `line_texts` for no-op/indent checks, gates at each GitHub/GitLab render site against that site's real apply range, strips fences from delivery-key material unconditionally, and prints patch pass/downgrade counts. **`filter_findings.py`** mirrors upstream stripping (non-string, oversize bounds shared with JS, strip patch when `suggestion` was injection-stripped) plus `suggested_fix_codes_removed` stats.
> 
> **Contracts & docs:** all seven agent JSON shapes and AGENTS docs list `suggested_fix_code`; report-format moves the field into canonical fields and shows plain (not apply-checked) blocks in the full report template while inline delivery docs describe the gate.
> 
> **Tests & CI:** broad `test_filter_findings` / parity golden coverage, bench adapter fixture expanded for gated rendering (`missing_end_line` downgrade), registry table count updated, Python coverage floor **92.1 → 92.4**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6710537ba704b5c7251713cbb73313770a55d20. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/post_review.py
```diff
@@ -1020,14 +1357,22 @@ def resolve_marker_sha(data):
 # ---------------------------------------------------------------------------
 
 
-def post_github(data, valid_lines):
+def post_github(data, valid_lines, line_texts):
```
leehopper, line 1375: **💡 [LOW] bench fixture-fidelity test bypasses the new suggested_fix_code apply-check gate**

post_github/post_gitlab now wrap every rendered finding in _gated_finding before calling render_comment_body, downgrading a fence that fails the deterministic apply-check (e.g. missing end_line). bench/tests/test_adapter.py's reference-payload builders hand-mirror the old, pre-#63 construction and call render_comment_body(f) directly, never through the gate, so the committed golden fixture still embeds a raw suggestion fence for a finding that lacks end_line, which real post_review.py --dry-run would now downgrade to no fence at all.

**Suggested fix:**
Update _github_comment/_gitlab_discussion in bench/tests/test_adapter.py to route through post_review._gated_finding (or call the real post_github/post_gitlab entry points) before rendering, so TestFixtureFidelity actually proves the fixture matches post_review.py's true --dry-run output for findings carrying suggested_fix_code.

scripts/post_review.py
```diff
@@ -1020,14 +1357,22 @@ def resolve_marker_sha(data):
...
     owner = data["owner"]
     repo = data["repo"]
     pr_number = data["pr_number"]
     findings = data.get("findings", [])
 
     check_tool("gh")
 
+    def degraded(filepath, line, finding):
+        """One entry for the body section, which has no anchor to apply against."""
+        return filepath, line, _gated_finding(finding, None, valid_lines, line_texts)
```
leehopper: **💡 [LOW] Byte-identical `degraded` closure duplicated in post_github and post_gitlab**

post_github (lines 1372-1374) and post_gitlab (lines 1595-1597) each define a local `degraded(filepath, line, finding)` closure with an identical body and docstring, called 4 times per function. Extracting one module-level helper that takes valid_lines/line_texts explicitly removes the duplication and keeps both posters' gating logic in lockstep by construction instead of by copy-paste discipline.

**Suggested fix:**
Before: two local closures `def degraded(filepath, line, finding): return filepath, line, _gated_finding(finding, None, valid_lines, line_texts)` defined separately inside post_github and post_gitlab. After: one module-level `def _degraded_entry(filepath, line, finding, valid_lines, line_texts): return filepath, line, _gated_finding(finding, None, valid_lines, line_texts)`, called as `_degraded_entry(fp, ln, f, valid_lines, line_texts)` from both posters.

workflows/src/stages.js
```diff
@@ -541,17 +541,48 @@ export async function discover(ctx, input) {
...
+      // too (only when it was actually present) — the sibling missing-line_end branch counts
+      // its drop and says so in its own gap; a silent discard here left the same class of
+      // loss unreported and unmeasured.
+      if ('suggested_fix_code' in f) {
+        delete f.suggested_fix_code;
+        droppedFixForImplausibleSpan += 1;
+      }
       droppedLineSpans += 1;
     }
   }
```
leehopper, line 580: **💡 [LOW] Repeated suggested_fix_code drop-and-count block in the discover line_end loop**

The no-line_end branch (lines 555-558) and the implausible-span branch (lines 571-574) each repeat the identical three lines `if ('suggested_fix_code' in f) { delete f.suggested_fix_code; droppedX += 1; }`, differing only in the counter name. A tiny helper returning whether the field was present would let both branches read as a single line and keep the check itself in one place.

**Suggested fix:**
Before: two inline `if ('suggested_fix_code' in f) { delete f.suggested_fix_code; droppedX += 1; }` blocks. After: a helper `function dropSuggestedFixCode(f) { if (!('suggested_fix_code' in f)) return false; delete f.suggested_fix_code; return true; }` and each call site becomes `if (dropSuggestedFixCode(f)) droppedNoLineEnd += 1;` / `if (dropSuggestedFixCode(f)) droppedFixForImplausibleSpan += 1;`.

leehopper (commented):
Code Gauntlet automated review: 3 low-severity findings (1 bug, 2 simplification suggestions). See inline comments.

---
Generated by code-gauntlet | Reviewed up to: 76d7eefe1a0172fccd843a9c7c32ec477d278ae8

scripts/post_review.py
```diff
@@ -765,6 +814,286 @@ def _suggestion_fence(payload):
...
+_FIX_NO_OP = "no_op_replacement"
+_FIX_INDENTATION = "indentation_mismatch"
+_FIX_TOO_LARGE = "replacement_too_large"
+_FIX_CARRIAGE_RETURN = "carriage_return"
+
+# The vocabulary is CLOSED: every downgrade names exactly one of these, in the
+# stable warning `suggested-fix downgraded: {file}:{line} ({reason})`. Adding a
+# reason is a deliberate act — a free-text reason would make the record
+# unreadable in aggregate.
+_FIX_REASONS = frozenset(
```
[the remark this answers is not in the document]
leehopper: Addressed in a671053: `_FIX_REASONS` is now load-bearing — `_gated_finding` validates every recorded downgrade reason against it and raises `ValueError` on an unknown reason, so a typo'd reason in a future gate edit fails at the first downgrade instead of recording garbage. Covered by `TestGatedFindingRejectsUnknownReason` (mutation-verified red with the check removed).

scripts/post_review.py
leehopper: Fixed in a671053. Both reference builders now route through `post_review._gated_finding` with the same apply-range formula the posters use, against explicit `valid_lines`/`line_texts` oracles. The SQL-injection finding gained `end_line: 15` so its fence is legitimately kept, and a fourth GitHub fixture finding carries `suggested_fix_code` with no `end_line` — the committed fixture now embeds its `missing_end_line` downgrade (no fence, prose fallback, skip warning, counts updated; fixture renamed `github_4_comments_2_skipped.json`). Mutation-verified: reverting the builders to `render_comment_body(f)` turns the byte-compare and a direct assertion red.

scripts/post_review.py
leehopper: Done in a671053: module-level `_degraded_entry(filepath, line, finding, valid_lines, line_texts)`; both posters' local closures removed, all 8 call sites updated, docstring consolidated.

workflows/src/stages.js
leehopper: Done in a671053: `dropSuggestedFixCode(f)` extracted; both branches are now `if (dropSuggestedFixCode(f)) <counter> += 1;` with the why-comments kept. Bundle rebuilt.

PR #222, opened 2026-08-20
feat(delivery): emit GitLab suggestion:-m+n so multi-line patches ship instead of downgrading

### LINKED ISSUE #219: GitLab multi-line suggestions: emit `suggestion:-m+n` so multi-line patches can ship instead of downgrading

leehopper:
Closes #219.

## What changes

Multi-line `suggested_fix_code` on GitLab previously downgraded to prose by design (#63 D9: the poster never emitted the offset syntax, so `anchor_mismatch` fired on any `end_line != line`). This PR teaches the GitLab delivery path to emit ```` ```suggestion:-m+n ```` and extends the apply-check comparison to the range the offsets actually realize.

- `_gitlab_fence_offsets(anchor, line, end_line)` is the single pure producer of offsets; it distinguishes anchor-outside-range from cap-exceeded. `_gitlab_anchored` (module-level, so the bench mirror calls the real decision) turns that into the `apply_range` the existing gate compares against the stated `[line, end_line]`. A kept fence's offsets therefore provably realize the stated range; the gate's nine pre-existing checks, order, and reason strings are byte-identical for existing callers.
- Offsets travel out of band as keyword-only parameters (`_suggestion_fence(payload, *, offsets=None)`, `render_comment_body(finding, *, fence_offsets=None)`, `render_group_body(..., *, fence_offsets=None)`). Nothing reads them from the finding dict: caller JSON is untrusted, and an in-band key would let a hand-assembled payload widen a range the gate approved as single-line. A forged in-band key is inert (tested).
- `deliver()` now takes a body factory invoked with the same `line` it writes into `position["new_line"]`, so a call site cannot bake offsets for one anchor and post the position at another. A wire-level test asserts every captured offsets header is centred on the position it ships with.
- Spans past GitLab's offset cap downgrade with the new `span_exceeds_platform_cap` reason (vocabulary 12 → 13; both tripwires updated). GitLab silently CLAMPS offsets above 100 server-side (`Suggestible::MAX_LINES_CONTEXT`), so an emitted header past the cap would one-click-apply a range the header does not state — the emitter never renders one. This branch only binds end-to-end when a caller raises `maxLineSpan` above its default 100.
- Position payload, `validate_position`, GitHub delivery bytes, single-line GitLab fence bytes, delivery keys/markers (#132/#208), and the `warn_skip` format are unchanged.
- `bench/tests/test_adapter.py::_gitlab_discussion` calls `post_review._gitlab_anchored` instead of hard-coding `(line, line)`, so the fidelity mirror cannot silently become fiction.
- Skill docs: `delivery-guide.md` records the GitLab fence header and the overlap decision; `report-format.md` wording now covers the offset fence.

## Platform facts the design rests on (verified against GitLab source/docs)

- Offsets are parsed relative to the note's `position.new_line` (`lib/gitlab/diff/suggestions_parser.rb`), identically for API-created discussions (GitLab's own Suggestions API doc demonstrates it).
- The POST is not opaque markdown: suggestions are materialized synchronously at note creation from the whole file blob at head_sha. Safety comes from the range invariant — every line of a kept range is a new-side diff line at the head_sha the position pins — not from opacity.
- 4+ backticks with an offsets header is documented working; `_suggestion_fence`'s longest-run+1, min-3 rule matches GitLab's own suggestion UI (gitlab-org MR !172981).

## Decision of record (rejected alternative)

Overlapping multi-line ranges in one file make GitLab refuse the whole batch apply, and serially applying one outdates the other. Accepted as parity with the already-shipped GitHub multi-line path (same exposure, no predicate there either); recorded in delivery-guide.md. Follow-up filed as an adjacent issue rather than growing this PR.

## Verification

- Pre-implementation red-team: 2 opus adversaries (GitLab platform semantics with source-level verification; codebase mechanism) → REVISE; all findings adjudicated into the design, including the out-of-band channel (was an in-band forgeable key), the 13th reason, the body-factory coupling, and the bench-mirror requirement.
- Two post-implementation adversarial review rounds (fresh-eyes + independent claim re-verification that re-ran gates and re-drove mutation-ledger entries). No correctness findings survived round 2; the one round-2 finding (coverage provenance number) is fixed in this diff.
- Mutation ledger: 11 mechanism mutations driven red one at a time (m/n swap, off-by-one below-offset, helper always-None, cap comparisons deleted, factory ignoring its anchor, cap reason silenced to anchor_mismatch, vocabulary reverted to 12, bench mirror re-hard-coded, group-body offsets unforwarded, in-band read reintroduced, corroborator fallback anchored at the primary), each restore byte-verified with __pycache__ purged.
- Gates (Python 3.12.13, serial, uncontended): `tests/` 1541 passed + 640 subtests; `bench/tests/` 539 passed; scripts coverage 93.42 ≥ 92.5; bench coverage 88.48 ≥ 87.5; `node --test workflows/test/*.test.js` green; `git diff --exit-code workflows/pipeline.js` clean; `pre-commit run --all-files` all hooks pass.
- Coverage floors ratcheted per the AGENTS.md policy (headroom exceeded 1.0 pp on both): scripts 92.4 → 92.5, bench 87 → 87.5, provenance note updated to this PR's measurements, `AGENTS_SET_BUDGET_BYTES` re-pinned (20_579 → 20_613), byte-parity with ci.yml test-enforced.

## Recall/noise note for the wave-end paired mini

Multi-line GitLab fences now enter delivered comment bodies where prose previously stood; delivery keys are unchanged by construction (fence-independent), so reruns over pre-#219 reviews stay `already_present`.

https://claude.ai/code/session_017Yh6Xs3vNyJdy9vWUUoRVs

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 3555604447ff02479b1443997fac529d870b96f5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/post_review.py
```diff
@@ -1059,6 +1098,70 @@ def _gated_finding(finding, apply_range, valid_lines, line_texts):
...
+    consume it: a non-integer or absent bound answers ``None`` and leaves the
+    gate's own ``missing_end_line`` / ``invalid_range`` rules to name what is
+    wrong. *cap_exceeded* is True only when the range was otherwise realizable
+    and the platform cap alone forbade the header — which is what separates
+    ``span_exceeds_platform_cap`` from ``anchor_mismatch``.
+    """
+    if any(
+        isinstance(v, bool) or not isinstance(v, int) for v in (anchor, line, end_line)
+    ):
+        return None, False
```
leehopper: **💡 [LOW] Third copy of the bool/int type-guard idiom**

The one-liner idiom isinstance(v, bool) or not isinstance(v, int) — used to keep True/False and floats from silently passing an int check — now appears a third time in this file (validate_position line 598, _suggested_fix_gate lines 991-994, and this new _gitlab_fence_offsets). A tiny shared predicate such as _is_plain_int(v) would let the three call sites read as one guarded concept instead of three near-identical inline expressions that could drift if one is edited and the others are not.

**Suggested fix:**
Extract a small helper, e.g. `def _is_plain_int(v): return isinstance(v, int) and not isinstance(v, bool)`, and rewrite each site to `if not all(_is_plain_int(v) for v in (...))`. Before: `isinstance(v, bool) or not isinstance(v, int)` repeated at 3 sites. After: one `_is_plain_int` helper called from all 3.

leehopper (commented):
Code Gauntlet review: 627 changed lines across 8 files, 0 correctness/security/cross-file/test/convention issues found. 1 low-severity simplification suggestion below.

---
Generated by code-gauntlet | Reviewed up to: d924cce86c604a366fdc2c14eb5a1e5075672ab3

scripts/post_review.py
leehopper: Applied in df36d51: extracted `_is_plain_int` and pointed all three sites at it (validate_position, _suggested_fix_gate's invalid_range check, _gitlab_fence_offsets); the bool-hashes-equal rationale now lives once on the predicate.

Issue #163, opened 2026-08-10
Migrate post_review.py parse_diff_lines onto scripts/diff_lines.py (follow-up to #134)


leehopper:
## Summary

Follow-up to #134: once `scripts/diff_lines.py` exists (PR `fix/134-verify-diff-lines`), migrate `scripts/post_review.py::parse_diff_lines` onto the shared primitive so the fixed walking mechanism (split on `"\n"` never `str.splitlines()`, per-hunk budget tracking, between-hunks header matching) has exactly one implementation.

## Scope

- `post_review.py::parse_diff_lines` drives the shared primitive; its platform-conditional header regexes, `new_files`/`old_paths` bookkeeping, and `(dict, set, dict)` return stay entirely local.
- Remove the interim `docs/duplication-register.md` row covering the `post_review.py ↔ verify_findings.py` `parse_diff_lines` pair (added by the #134 PR to keep the interim state classified).

## Ordering and proof obligations

Lands **after** both:
1. the #133 fixture PR (`test/133-glab-diff-fixtures`) — the pinned glab contract the migration's behavior-preservation is demonstrated against, and
2. the #134 PR — the primitive itself.

Proof: the full `TestParseDiffLinesPostReview` suite (~30 tests accumulated over three dedicated bug-fix PRs), the GitLab position-contract suites, and the #133 fixtures all pass **unchanged**. This is deliberately its own PR so "the pre-existing suite passes unchanged" is a claim a reviewer can actually verify — the reason the batch did not fold this into the #134 PR.

Deliberately sequenced last in the #131/#132/#133/#134/#141 batch: it rebases `post_review.py` once, against a settled file.

leehopper:
Pickup note from the #63 work (2026-08-19): `parse_diff_lines` in `post_review.py` now also returns a fourth value — `line_texts: dict[(path, new_line) -> str]`, the diff's new-side line text, which the suggested_fix_code apply-check uses as its content oracle. A migration onto `scripts/diff_lines.py` needs to carry that retention (and its `\ No newline` handling) along with the #134-era fixes, and `verify_findings.py`'s twin copy still lacks it.

PR #225, opened 2026-08-20
feat(report): strip suggested_fix_code from every report-path model input

### LINKED ISSUE #220: Report-side apply-check: REVIEW.md renders suggested_fix_code blocks no check has vetted

leehopper:
## What

The report path can no longer render `suggested_fix_code` at all: `reportStage` strips the field
from the report-writer's input (new `stripFixCode`, copy-based, built on `dropSuggestedFixCode`),
and `references/report-format.md` drops the "(not apply-checked)" render instruction. Delivery is
now the field's only rendered surface, gated by `post_review.py`'s existing apply-check against
its live oracle — delivered comment bytes and bench candidate bytes are byte-identical to main.

Closes #220.

## Why this shape

#220 asked for a report-side apply-check and flagged the open decision: where can a check run
when the report renderer is a model? The design went through two red-team rounds (four
independent opus reviews) before implementation; the full memos are summarized here because they
eliminated every other placement:

- **Verify-stage compute** (v1): dead. `VERIFY_SLICE_FIELDS` deliberately omits
  `suggested_fix_code`; adding it would push unbounded escape-dense patches through the
  content-proofed artifact-writer transcription channel, where one mangled backslash costs a
  whole slice its verification (`deterministic: true`, no retry). Also, a
  `from scripts.post_review import …` in `verify_findings.py` raises `ModuleNotFoundError` in
  the shipped invocation (`sys.path[0]` is `scripts/`, repo root absent) while staying green
  under pytest.
- **Strip at the verify boundary** (v1): not fail-closed — `runPhase` replays checkpointed
  phases (`buildResumeCheckpoints` carries every phase's output; `PERSISTED_RESUME_PHASES`
  replays `challenge` straight into `reportInput`), bypassing any verify-time strip. And an
  in-place strip would also strip delivery's input: verify's pinned Phase-2 diff and delivery's
  live diff diverge (incremental mode's bounded diff, moved PR heads), so it would delete fences
  delivery's own stronger check would have kept — a delivery regression and a moved
  bench-scored surface.
- **Phase 8 deterministic gate + code-rendered section** (v2): the render half survives review
  but is blocked on a correct file-based oracle. `post_review.py`'s diff parser is CLI-shaped
  and cannot parse the pinned `args.diffPath` (three producers; git-diff TAB termination and
  C-quoting; no platform value exists at the call site), and a findings.json rewrite collides
  with the persist plan's `expect[]` checksums, `assemble_artifacts` re-derivation, and
  `materialize_artifacts` self-healing. That capability is split into #226 (blocked on
  #163, the walker line-text migration) with all review-established requirements recorded there.

What remains is the structural core both red-team rounds converged on: the model never receives
the field, so no template instruction, writer improvisation, or replay can render unvetted patch
bytes, and the dishonest-label state is unreachable. The strip is copy-based at `reportStage`
entry — upstream of `dimensionsSummaryTable`, `consolidateForReport`, the oversized-payload
sizing, chunking, `minimalReport`, and `reportPrompt` (both buckets, every segment) — while
`selectDelivery`/`writeArtifacts` keep reading the caller's unstripped objects.

## Accepted residuals (named, not hidden)

- A `report`-phase checkpoint replayed from a pre-change run carries pre-change bytes with the
  old honest "(not apply-checked)" label — no false claim is possible.
- The Phase 8 orchestrator is a model with disk access reading findings.json; the template no
  longer instructs any patch render, which is the same trust posture as every other Phase 8
  rule.
- Unverified-bucket patches stop rendering too (they were never vetted; prose `suggestion`
  still renders).
- Pre-existing, unchanged: `suggested_fix_code_removed_by`/`_removal_reason` filter stamps can
  reach the report-writer prompt as metadata; carries no patch bytes. Recorded in #226.

## Verification

- Full gates green locally (and expected in CI): pytest `tests/` 1542 passed (scripts+.github
  coverage 93.45% ≥ 92.5), `bench/tests/` 539 passed (88.48% ≥ 87.5), `node --test` 864 passed
  (98.94 / 86.70 / 98.45 ≥ 98 / 85.7 / 97.4 + presence check), bundle byte-identical to a fresh
  build, `pre-commit run --all-files` clean.
- Mutation ledger (every mechanism, byte-verified restores): strip call removed → 2 JS tests
  red; `stripFixCode` neutered to identity → 4 JS tests red; partial strip (`unverified` left
  raw) → 1 JS test red; copy discipline neutered → 2 JS tests red (delivery invariance is
  genuinely pinned); template block restored → the new
  `TestFullReportTemplateRenderInstructions` pin red; old `!f ||` guard restored → the
  primitive-passthrough test red (TypeError).
- Two adversarial implementation-review rounds (four opus reviews); round 1's one minor (the
  guard could throw on primitive members in a replayed checkpoint, escaping the non-fatal
  report contract) and six nits fixed in the amended commit; round 2 verified the fixes.

## Notes for the wave record

- Suites-only per the Wave 5 rule: recall/noise surfaces unmoved — delivery payload and bench
  candidate bytes are byte-identical (bench scores `post-review-payload.json`; the report is
  not a scored surface).
- JS functions coverage headroom measured locally at 1.05 pp (98.45 vs 97.4 floor), pre-existing
  on main (98.44); per policy a ratchet belongs in a housekeeping PR keyed to this PR's CI
  measurement, not here.
- Adjacent issue filed: #226 (render apply-checked patches via a read-only Phase 8 gate;
  blocked on #163).

https://claude.ai/code/session_018ReTwprjW1UMgw5vyyqoTS

### DIFF
--- a/skills/code-gauntlet/references/report-format.md
+++ b/skills/code-gauntlet/references/report-format.md
@@ -54,7 +54,7 @@ The pipeline's declaration lives in `workflows/src/registry.js`. A field this ta
 | `suggestion` | string | no | Prose fix advice; rendered by `post_review.py` as a "Suggested fix:" block. |
 | `claude_md_rule` | string | no | The cited project rule, quoted with its source file; required by contract for convention findings; OMITTED (never null) when no rule applies. |
 | `cross_file_refs` | array | no | Other files involved in the finding |
-| `suggested_fix_code` | string | no | Exact replacement source for `file:line_start-line_end`, emitted by discovery agents only when the fix is a byte-exact, drop-in replacement for exactly those lines. Delivery (`scripts/post_review.py`) runs a deterministic apply-check before ever rendering it as a committable GitHub/GitLab `suggestion` block (one-click apply) — non-string, stale, wrong-range, wrong-anchor, or oversized fails the check and the finding downgrades to the prose `suggestion` instead. On GitLab, the render-site apply range is the discussion anchor plus offsets, not the stated range directly — a span the anchor and offsets can't realize within GitLab's cap downgrades the same way. |
+| `suggested_fix_code` | string | no | Exact replacement source for `file:line_start-line_end`, emitted by discovery agents only when the fix is a byte-exact, drop-in replacement for exactly those lines. The report path never renders this field — the pipeline strips it from the report-writer's input before dispatch (`stripFixCode` in `workflows/src/stages.js`). The only rendered surface is delivery: `scripts/post_review.py` runs a deterministic apply-check before ever rendering it as a committable GitHub/GitLab `suggestion` block (one-click apply) — non-string, stale, wrong-range, wrong-anchor, or oversized fails the check and the finding downgrades to the prose `suggestion` instead. On GitLab, the render-site apply range is the discussion anchor plus offsets, not the stated range directly — a span the anchor and offsets can't realize within GitLab's cap downgrades the same way. |
 
 ### Per-dimension fields
 
@@ -135,18 +135,6 @@ Example: "This PR adds JWT-based authentication to the API layer. The token vali
 {If finding.claude_md_rule or finding.spec_text is present — the rule the finding is measured
 against. `agents/report-writer.md` instructs the same:}
 **Cited rule:** {finding.claude_md_rule or finding.spec_text}
-
-{If finding.suggested_fix_code is present — informational only, this template does not run
-delivery's apply-check, so render it as a PLAIN fenced code block, never a ```suggestion
-fence: a suggestion fence pasted into a PR comment is a one-click apply button, and nothing
-on the report path has gated this content the way `post_review.py` gates the Inline PR
-Comment Format below.}
-Proposed replacement for {finding.file}:{finding.line_start}-{finding.line_end} (not apply-checked):
-```
-
-{finding.suggested_fix_code}
-
-```
 
 ---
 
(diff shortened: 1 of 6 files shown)


leehopper (commented):
code-gauntlet review: 0 high-confidence findings across all 7 dimensions (correctness, security, cross-file impact, test coverage, conventions, type design, simplification). No issues identified.

---
Generated by code-gauntlet | Reviewed up to: 828814a21c553709be9c1cfb2ec7956fc9721255

Issue #226, opened 2026-08-20
Render apply-checked `suggested_fix_code` blocks in the report (read-only Phase 8 gate)


leehopper:
## Context

#220 removed the report path's unvetted `suggested_fix_code` render structurally: the pipeline
strips the field from the report-writer's input (`stripFixCode`, `workflows/src/stages.js`), and
the template no longer instructs the "(not apply-checked)" block. That resolves the honesty
problem but also removes the display of patches that WOULD pass the apply-check — in local-diff
mode a vetted patch is now visible only as the prose `suggestion`.

This issue is the capability half: run the diff-only subset of delivery's apply-check
deterministically at Phase 8 and render the KEPT patches into the report as code-owned bytes
(the `dimensionsSummaryTable` / footer precedent: code owns the content, the doc designates the
place).

## Requirements established by the #220 design red-team (2 rounds, 4 opus reviews — see PR #225)

These are the constraints any implementation must satisfy; each was verified against the code:

1. **Oracle**: `post_review.py`'s `parse_diff_lines` regexes are CLI-shaped (`gh pr diff` /
   `glab mr diff`) and cannot parse the pinned `args.diffPath` file, which has three producers
   (`gh pr diff` a/-prefixed, `glab mr diff` verbatim, `git diff` a/-prefixed with TAB
   termination and C-quoting — including the incremental path, which is git-diff-shaped even on
   GitLab). No correct `platform` value exists at the call site. The correct oracle is the
   both-spellings approach `verify_findings.py::_path_spellings` uses, over
   `diff_lines.py::walk_diff` — which requires #163 first (the walker's `DiffEvent` carries no
   line text today). `line_texts` must be populated under both spellings (the gate's
   `_span_texts` needs one spelling to hold across a whole span). Header paths need
   `diff_lines._decode_header_path` semantics (TAB, C-quotes).
2. **Never mutate findings.json**: its bytes are covered by the persist plan's `expect[]`
   checksums, `assemble_artifacts._assemble` re-verifies them on re-runs (and would re-derive
   post-review.json from a gated file — changing delivered bytes), and a
   `materialize_artifacts.py` re-run self-heals from `persistReturn`, silently discarding any
   stamp. Emit a separate derived artifact (or fill a designated report section) instead.
3. **Never let the mode fall into delivery's dry-run path**: `main()` sets `DRY_RUN` from
[...]

PR #227, opened 2026-08-20
fix(delivery): migrate post_review parse_diff_lines onto diff_lines.walk_diff (#163)

### LINKED ISSUE #163: Migrate post_review.py parse_diff_lines onto scripts/diff_lines.py (follow-up to #134)

leehopper:
Closes #163. Follow-up to #134; prerequisite for #226 (the walker now carries line text).

## What changes

- `scripts/diff_lines.py`: `DiffEvent` gains `text` — the body line with its marker column removed (`raw[1:]` for `+`/`-` lines; a context line loses its leading space only when it has one, so a zero-prefixed bare line is passed through whole). Removed lines carry text too: the walk stays lossless.
- `scripts/post_review.py::parse_diff_lines`: the hand-rolled split/regex/budget loop is gone; it reads `walk_diff` events and keeps only its platform header semantics (GitHub `a/`/`b/` strip, GitLab verbatim, `/dev/null`, the `@@ -0,0` added-file signal, `new_files`/`old_paths`). Signature and the four-value return are unchanged.
- `docs/duplication-register.md`: the `parse_diff_lines` row is removed; the count sentence is corrected (it was already stale: 40 rows at HEAD, not 39) and a new mechanism test in `tests/test_docs_registry.py` keeps it honest. One row added for the remaining `is_line_valid` ↔ `is_line_in_diff` shard.

## Proof obligation (from the issue)

The pre-existing `tests/test_post_review.py` suite passes **unchanged** (the diff to that file is pure addition: 0 deleted lines). `tests/test_diff_lines.py`'s whole-stream assertions gain `text=`; the glab fixture contract and GitLab position suites pass as before.

A reviewer ran a differential harness (HEAD vs branch, same process, 53 hand-built input classes + the four glab fixtures + 240k random diffs with and without a trailing newline). Exactly two divergence classes exist, both fixes, both named in the docstring:

1. Header paths arrive **decoded** (git's TAB terminator after a path containing a space, and C-quoting for non-ASCII/control characters). A finding on `dir with space/x.py` or `café.py` previously matched nothing; it now anchors. The decode also runs on `glab mr diff`'s verbatim paths, where it mis-reads only a name that itself carries a literal TAB or is quote-wrapped — documented as an accepted limitation in `_decode_header_path` and pinned by a test so the choice is visible.
2. A newline-terminated diff whose last hunk is cut short no longer mints a phantom final line (the old loop walked the trailing empty string as a context line, which would also have fed the apply-check an empty, false content oracle).

## Validation

- `python -m pytest tests/ -q`: 1559 passed. `pre-commit run --all-files`: clean. Scripts coverage 93.41% (floor 92.5).
- Two adversarial review rounds (3 opus lenses in isolated tree copies, then a fix pass). Mutation ledger, each red on the named test and restored byte-exact: unconditional `b/` strip on GitLab; unconditional `a/` strip; `line_texts` keyed on the stripped spelling; `current_file` guard dropped; context-line slice made unconditional; removed-line text dropped; `text` forced to `None`; `@@ -0,0` signal dropped; register count sentence falsified.

## Measurement note

Suites-only per the wave rule. Recall-relevant delta for the wave-end paired mini: findings on paths with spaces or non-ASCII names now anchor inline on both platforms where they previously degraded to the summary.

https://claude.ai/code/session_01NGDENPwjcBSjNTp3GgjFe3

---

> [!NOTE]
> **Medium Risk**
> Changes how GitHub/GitLab diffs are parsed for inline comment positions and suggested-fix text. Wrong parsing would mis-anchor or drop review comments, but the public return shape is unchanged and existing poster tests are preserved.
> 
> **Overview**
> Finishes consolidating the poster's unified-diff walk onto `diff_lines.walk_diff`. `parse_diff_lines` now consumes walker events and keeps only platform header semantics (GitHub `a/`/`b/` strip, GitLab verbatim paths, `/dev/null`, added-file `@@ -0,0`).
> 
> `DiffEvent` now carries `text` (marker column stripped, including removed lines) so the poster can fill `line_texts` without a second parse. Two real fixes come with the switch: git TAB/C-quoted header paths decode before matching, so findings on spaced or non-ASCII names can anchor; a trailing empty split no longer mints a phantom last line.
> 
> The duplication register drops the accidental `parse_diff_lines` pair, records the remaining `is_line_valid`/`is_line_in_diff` helpers, and a new docs-registry test keeps the row-count sentence in sync. Existing `test_post_review` cases are unchanged; new tests pin header decoding and line-text rules.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1ab4f17e1c13a2184a2745935c0595bf8a95036. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_docs_registry.py
+++ b/tests/test_docs_registry.py
@@ -26,7 +26,7 @@ REPO = Path(__file__).resolve().parents[1]
 # many as fifty; extend the range here if it ever does.
 _ONES = ["", "one", "two", "three", "four", "five", "six", "seven", "eight", "nine"]
 _TENS = {30: "thirty", 40: "forty", 50: "fifty"}
-_NUMBER_WORDS = {50: "fifty"}
+_NUMBER_WORDS = {50: _TENS[50]}
 for _tens in (30, 40):
     _NUMBER_WORDS[_tens] = _TENS[_tens]
     for _i in range(1, 10):
--- a/tests/test_docs_registry.py
+++ b/tests/test_docs_registry.py
@@ -14,11 +14,24 @@ Scope: git-tracked files only (``git ls-files``), so gitignored local artifacts
 never fail the suite.
 """
 
+import re
 import subprocess
 import unittest
 from pathlib import Path
 
 REPO = Path(__file__).resolve().parents[1]
+
+# Spelled-out number words for the duplication register's row-count sentence
+# ("Forty rows: ..."). The table has never held fewer than thirty rows or as
+# many as fifty; extend the range here if it ever does.
+_ONES = ["", "one", "two", "three", "four", "five", "six", "seven", "eight", "nine"]
+_TENS = {30: "thirty", 40: "forty", 50: "fifty"}
+_NUMBER_WORDS = {50: "fifty"}
+for _tens in (30, 40):
+    _NUMBER_WORDS[_tens] = _TENS[_tens]
+    for _i in range(1, 10):
+        _NUMBER_WORDS[_tens + _i] = f"{_TENS[_tens]}-{_ONES[_i]}"
+_WORD_TO_NUMBER = {word: number for number, word in _NUMBER_WORDS.items()}
 
 # Root-level markdown: community-health and contract files only. CHANGELOG.md is
 # semantic-release-generated. Review outputs (deep-review-*.md and kin) are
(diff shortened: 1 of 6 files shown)


tests/test_docs_registry.py
```diff
@@ -14,12 +14,25 @@
...
 from pathlib import Path
 
 REPO = Path(__file__).resolve().parents[1]
 
+# Spelled-out number words for the duplication register's row-count sentence
+# ("Forty rows: ..."). The table has never held fewer than thirty rows or as
+# many as fifty; extend the range here if it ever does.
+_ONES = ["", "one", "two", "three", "four", "five", "six", "seven", "eight", "nine"]
+_TENS = {30: "thirty", 40: "forty", 50: "fifty"}
+_NUMBER_WORDS = {50: "fifty"}
```
leehopper: **💡 [LOW] _TENS[50] is dead — "fifty" is hardcoded twice with no single source of truth**

_TENS defines all three tens words (30/40/50), but the loop below only iterates (30, 40), so _TENS[50] is never read; _NUMBER_WORDS's 50-entry is instead a separate hardcoded literal "fifty". The two spellings of "fifty" can silently drift — editing _TENS[50] would have zero effect on the actual output, which is confusing for a maintainer who reasonably expects _TENS to be the single source for every tens word used in this table.

**Suggested fix:**
Before: _NUMBER_WORDS = {50: "fifty"}. After: _NUMBER_WORDS = {50: _TENS[50]} — makes _TENS the single source of truth for all three tens words instead of leaving one duplicated as a bare literal.

```suggestion
_NUMBER_WORDS = {50: _TENS[50]}
```

leehopper (commented):
Code Gauntlet review: 1 low-severity finding, 0 unverified/degraded findings. See inline comment.

---
Generated by code-gauntlet | Reviewed up to: 07a7ce6d3ef97e3fb64f2243191d99bdbdef725d

tests/test_docs_registry.py
leehopper: Applied in c1ab4f1: `_NUMBER_WORDS = {50: _TENS[50]}`.

PR #228, opened 2026-08-20
feat(report): render apply-checked suggested_fix_code patches into a sibling artifact (#226)

### LINKED ISSUE #226: Render apply-checked `suggested_fix_code` blocks in the report (read-only Phase 8 gate)

leehopper:
Closes #226. Builds on #163 (PR #227, merged as 5b4829c); rebased onto main.

## What ships

A read-only Phase 8 gate, `scripts/report_patches.py`, runs the diff-only subset of delivery's deterministic apply-check over the **pinned Phase 2 diff** and renders every kept `suggested_fix_code` patch into a sibling artifact, `{output_dir}/code-gauntlet-patches-{head_sha_short}.md`. The report is never edited and delivery bytes are unchanged.

- `python3 scripts/report_patches.py --output-dir DIR --head-sha SHORT` (materialize's shape; findings, diff and output paths are derived). One JSON receipt line on stdout (`ok, path, oracle, candidates, kept, downgraded, reasons, filtered_earlier, findings, warnings, errors`); exit 0 when the artifact was written (even with 0 kept), 1 on a failure that still ships a receipt, 2 for a malformed command.
- Oracle: the producer is detected from the diff's **first line** (`diff --git "?a/` = git-shaped: `gh pr diff`, `git diff`, incremental; anything else = glab's verbatim headers) and the oracle is keyed by the new `post_review.parse_diff_text(platform, text)` — the same parser delivery runs live, factored out of `parse_diff_lines`. One keying implementation, no alias keys.
- Gate reuse: `post_review._gated_finding(..., warn_label="report-patch")` (delivery's `suggested-fix downgraded:` bytes unchanged), a per-reason tally (`_FIX_REASON_COUNTS`), a shared `reset_run_state()`, and `_fence_run` factored out of `_suggestion_fence`.
- Label honesty: `# Apply-checked patches (against {sha})`, "the pinned review diff … not the current working tree or branch", "platform render-site constraints are not applied here, so delivery may still downgrade a listed patch", high-confidence bucket only. A missing or empty diff fails every patch closed (`no_diff_oracle`) and says so.
- `stages.js`: `REPORT_EXCLUDED_FIELDS` (`suggested_fix_code` + its two removal stamps), `stripFixCode` → `stripReportExcludedFields`.
- SKILL.md Phase 8 gains the step (after Collect, before Deliver, keyed on `artifactPaths.findings` non-null, every persist channel incl. salvage, exit-code table); `phase8-delivery.md`, `headless-mode.md` (hard-rules bullet), `report-format.md` updated; `scripts/AGENTS.md` receipt-line bullet names the third script.

## Why a sibling artifact (design of record)

Two pre-implementation opus red-teams killed the in-report placement the issue sketched: the `## Review Methodology` heading exists on **no** report path (the report-writer contract never emits it; report-format.md's template has no reader), and the report is in the persist plan's `expect[]`, so an in-place edit breaks its content proof and any `materialize_artifacts.py` re-run silently reverts it. The issue's R2/R4 explicitly permit a separate artifact. The Phase 2 stale-truncate glob already covers the name.

## Requirements (issue body R1–R10)

All ten audited SATISFIED by an independent reviewer against the final tree. R1 deviates from its literal prescription (both-spellings) on purpose: both-spellings keying is sound for verify's membership **set** but not for a value **map** (`a/foo.py` + `foo.py` clobber, and `diff_path_spelling`'s own strip re-opens a dropped key to a different file); producer detection + `parse_diff_text` is strictly stronger.

## Validation

- `pytest tests/`: 1630 passed. `bench/tests/`: 540 passed. JS: 866 passed. Bundle fresh. `pre-commit` clean. Coverage: scripts 93.58 (floor 92.5), bench 88.39 (87.5), JS 98.95 / 86.73 / 98.45 (98 / 85.7 / 97.4). AGENTS.md byte ratchet re-pinned 20 605 → 20 628 for the receipt-line bullet only.
- Two adversarial review rounds (3 + 2 opus, isolated tree copies) plus two fix passes. Round 1 found a real blocker in the first keying design (wrong-file oracle under a real top-level `a/`/`b/` directory) — fixed by the producer-detection shape above and verified by an 800-diff differential harness (0 divergences vs `parse_diff_text` with the correct platform). Mutation ledger (all red, restored byte-exact): detection anchor un-anchored / bare `a/`-less / quote-less; platforms swapped; `errors="replace"` removed on either read; broad except narrowed; `ensure_ascii`/stdout guard reverted (subprocess test under `PYTHONIOENCODING=ascii`); `_one_line` skipped; neutralization applied to fences; fence run fixed at 3; `reset_run_state` removed; `_redact_secrets` removed from render; atomic write replaced; empty-diff check removed; `REPORT_EXCLUDED_FIELDS` shrunk; `stripReportExcludedFields` aliasing; G3 carrier policy gaining the artifact.
- Known residual, inherited from the shared gate and not introduced here: `diff_path_spelling`'s `^[ab]/` fallback can anchor a finding on a real top-level `b/x.py` to `x.py` when the diff touches `x.py` only — live delivery resolves identically. Tracked as an adjacent issue.

## Measurement

Suites-only. Nothing here is bench-scored: scoring reads `code-gauntlet-findings-*.json` and `post-review-payload.json`, both byte-identical. The new artifact is not a G3 degrade carrier (`_DEGRADE_CARRIER_POLICY` is a closed list; a bench test now pins that it stays out) and its scaffolding carries neither sentinel phrase. The one new bench contact is `invoke.py`'s `rglob("*.md")` identity-echo scan, which is permissive-direction only.

https://claude.ai/code/session_01NGDENPwjcBSjNTp3GgjFe3

---

> [!NOTE]
> **Medium Risk**
> New Phase 8 writer reuses the live apply-check and renders model-derived patch text into markdown (injection/redaction/path confinement). Delivery posting and persist-plan report bytes are unchanged.
> 
> **Overview**
> Adds a **read-only Phase 8 gate** (`scripts/report_patches.py`) that re-runs delivery’s diff-only apply-check against the **pinned Phase 2 diff** and writes kept `suggested_fix_code` patches to `code-gauntlet-patches-{sha}.md`. The report is never edited and PR/MR posting is unchanged.
> 
> The script detects git-shaped vs glab-verbatim diffs from the first line, then keys the oracle with a new `parse_diff_text` split out of `parse_diff_lines`. Gate helpers are shared (`_gated_finding` with `warn_label="report-patch"`, `_fence_run`, `reset_run_state`, per-reason `_FIX_REASON_COUNTS`). Failures fail closed (`no_diff_oracle` on missing/empty diff); stdout is always one JSON receipt.
> 
> `stripFixCode` becomes `stripReportExcludedFields`, also dropping the two `suggested_fix_code_removed_*` stamps from the report-writer prompt. Skill/docs add the step after materialize and before deliver. Bench pins that the patches file is **not** a G3 degrade carrier, since it can embed reviewed-repo source verbatim.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 908408e3c10e3d6fd8f9fe52782aff05ad88c286. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper (commented):
Automated code-gauntlet review: 3 findings survived verification and blind challenge (all tagged as suggestions — no blocking issues found).

---
Generated by code-gauntlet | Reviewed up to: 908408e3c10e3d6fd8f9fe52782aff05ad88c286

scripts/report_patches.py
```diff
@@ -0,0 +1,547 @@
...
+                kept=0,
+                downgraded=0,
+                reasons={},
+                filtered_earlier=0,
+                findings=0,
+                warnings=[],
+                errors=errors,
+            )
+        )
+        return 1
```
leehopper: **🟡 [MEDIUM] Duplicated pre-oracle-failure receipt block in report_patches.py's main()**

The confinement-check failure branch (lines 407-431) and the findings-load failure branch (lines 434-450) build and emit the identical `_receipt(...)` call — every keyword argument is byte-for-byte the same (ok=False, oracle="unattempted", candidates=0, kept=0, downgraded=0, reasons={}, filtered_earlier=0, findings=0, warnings=[]) — differing only in what was appended to `errors` beforehand. A small helper collapses both ~15-line blocks into two one-line calls.

**Suggested fix:**
Extract a helper, e.g. `def _pre_oracle_failure(out_path, errors): _emit_receipt(_receipt(ok=False, path=out_path, oracle="unattempted", candidates=0, kept=0, downgraded=0, reasons={}, filtered_earlier=0, findings=0, warnings=[], errors=errors)); return 1`, then replace both call sites with `if errors: return _pre_oracle_failure(out_path, errors)` and `if findings is None: return _pre_oracle_failure(out_path, errors)`. Before: two ~15-line identical `_emit_receipt(_receipt(...))` blocks. After: one 2-line helper plus two 1-line call sites.

workflows/src/stages.js
```diff
@@ -453,27 +453,42 @@ function findingSchema(spec) {
   };
 }
 
-// Shared by the two discover-side drop branches below and by stripFixCode: a dropped
-// suggested_fix_code is always the same operation (delete-and-report-whether-it-was-there),
-// only the reason for dropping it differs between call sites.
+// Shared by the two discover-side drop branches below and by
+// stripReportExcludedFields: a dropped suggested_fix_code is always the same
+// operation (delete-and-report-whether-it-was-there), only the reason for
+// dropping it differs between call sites.
```
leehopper: **💡 [LOW] dropSuggestedFixCode comment now falsely claims stripReportExcludedFields shares it**

The comment above dropSuggestedFixCode says it is 'Shared by the two discover-side drop branches below and by stripReportExcludedFields', but the renamed stripReportExcludedFields (lines 486-492) no longer calls dropSuggestedFixCode at all — it now inlines its own 'for (const key of REPORT_EXCLUDED_FIELDS) delete copy[key]' loop. The prior stripFixCode did call dropSuggestedFixCode(copy); the rename/refactor dropped that call but the comment describing the sharing relationship was only text-renamed, not corrected, so it now makes a false claim about the actual call graph.

**Suggested fix:**
Update the comment above dropSuggestedFixCode to state it is shared only by the two discover-side drop branches (remove the 'and by stripReportExcludedFields' claim), or restore a call to dropSuggestedFixCode inside stripReportExcludedFields's loop for the suggested_fix_code key if literal sharing is still intended.

---

**Corroborating finding — test-analyzer (test_coverage, confidence 57):**

**stripReportExcludedFields tests never exercise the realistic stamps-only finding shape filterFindings.js actually produces**

Every test of stripReportExcludedFields and reportStage's field-strip in workflows/test/stages_dimensions.test.js pairs suggested_fix_code together with its removal stamps (suggested_fix_code_removed_by/_removal_reason), but filterFindings.js's stripSuggestedFixCodeIfNeeded always deletes suggested_fix_code when it adds those stamps, so a real finding reaching reportStage after an earlier filter-strip carries the stamps WITHOUT suggested_fix_code — a shape no test constructs.

scripts/report_patches.py
leehopper: Applied in da3b365.

workflows/src/stages.js
leehopper: Applied in da3b365.

PR #231, opened 2026-08-24
fix(delivery): fail the suggestion fence closed on a same-diff a/-b/ path collision (#229)

### LINKED ISSUE #229: diff_path_spelling's ^[ab]/ fallback can anchor a finding on a real top-level a/ or b/ file to its stripped sibling
## Summary

`scripts/post_review.py::diff_path_spelling` resolves a finding's path against the diff oracle by trying the exact spelling, then the `^[ab]/`-stripped spelling. When the reviewed repo has a real top-level `a/` or `b/` directory, a finding on `b/x.py` whose own key is absent from the diff but whose stripped sibling `x.py` is present resolves to `x.py` — the gate then runs `_range_is_valid` / `_span_texts` (no-op, indentation, range checks) against a different file's text. A patch that is a literal no-op against the file it names can pass.

Preconditions are narrow (a real top-level `a/`/`b/` dir, the named file absent from the diff under its own key while the stripped name is present) and the behaviour is shared by both callers of `_gated_finding`:

- live delivery (`post_github` / `post_gitlab`), pre-existing since #63;
- the report-side gate `scripts/report_patches.py` (#226), which publishes the result into `code-gauntlet-patches-{sha}.md`.

Verified end-to-end during the #226 round-2 review: diff touches only `x.py`; finding on `b/x.py` with a patch equal to `x.py`'s line → `kept=1`, rendered under `` `b/x.py` ``.

## Fix shape (decision needed)

The strip fallback exists because discovery agents sometimes spell a path with git's synthetic prefix (`b/src/app.py`) while the parsed keys are unprefixed. Options:

1. Fail closed when the exact spelling is absent AND the stripped spelling collides with another header path in the same diff (needs the set of header paths at lookup time — `parse_diff_text` already has it via `old_paths`/`new_files` bookkeeping, or a new `header_paths` return).
2. Strip only when the diff is git-shaped (prefixes are synthetic there; under glab's verbatim headers a leading `a/`/`b/` is real) — the platform is known at both call sites.
3. Ratify the current behaviour as an accepted limitation and document it at `diff_path_spelling`.

Recall-sensitive (a stricter fallback can drop legitimate prefixed spellings), so it rides the Wave 5 paired mini if implemented. Surfaced by the #226 review (PR #228).

leehopper:
## Summary

Closes #229 by deciding among its three options and moving the fail-closed to where the harm is:

- **Fail closed on a same-diff collision — at the patch fence only.** When a finding's own path
  spelling matches `^[ab]/` and BOTH that spelling and its stripped form name real paths in the
  parsed diff, the diff names two distinct files one strip apart and the finding's spelling
  cannot say which file the patch targets. `_suggested_fix_gate` now downgrades the fence with
  the existing no-oracle reason (no new reason string); the prose `suggestion` still ships and
  the finding still delivers inline on its unchanged anchor. This closes both directions:
  - exact-hit collision — the fence previously validated a patch against the real prefixed
    file's text even when the patch was meant for the stripped sibling (the one-click
    corruption class the gate exists for);
  - exact-miss collision — the fence previously validated against the stripped sibling's text.
- **Reject option 2 (strip only for git-shaped diffs).** GitLab incremental runs feed discovery
  a git-shaped `git diff {last_reviewed_sha}..HEAD` (phase2-triage.md 2c branch 4) while
  delivery re-fetches verbatim `glab mr diff`, so synthetic-prefix spellings are legitimate on
  GitLab too; a platform-gated strip would drop them.
- **Ratify anchor-level behaviour (option 3), documented.** Per the gate's recorded asymmetry
  (post_review.py "a wrong anchor costs a misplaced comment a human reads and ignores; a wrong
  patch is committed by one click"), anchors keep HEAD's exact-then-strip resolution
  byte-for-byte. The off-diff-sibling case (the issue's verified repro: real `b/x.py` nowhere
  in the diff) is undecidable from the diff text and is now documented at `diff_path_spelling`
  as the accepted residual; a filesystem/git oracle was rejected because `report_patches`
  deliberately has no git access and gates against the pinned Phase-2 diff — a tree oracle in
  delivery only would make the artifact's kept set diverge from delivery's.

Mechanism changes riding along, both demanded by the design review:

- The four path helpers (`is_line_valid`, `diff_path_spelling`, `old_line_for`,
  `valid_lines_for_file`) no longer each re-derive the `^[ab]/` strip; one shared resolver owns
  it (the `is_new_file` precedent from PR #135, applied to the remaining four).
- `post_github` now posts the diff's own spelling as `comment["path"]` (as `post_gitlab`
  already did): a prefixed finding previously validated via the strip and then posted a path
  the PR does not have — the whole-review 422 class.

## Design history

The first design (parse-time alias map, fail-closed anchors, dict-subclass carrier) was killed
by a 2-opus adversarial review with verified counterexamples: fail-closed anchors drop
legitimate prefix-copied findings and their whole consolidation group; the old-side suppression
set is not computable from `walk_diff` (pure renames, binary, and mode-only changes emit no
`---`/`+++` headers); exact-hit collisions were untouched; the carrier silently desyncs bench's
payload mirror. Both reviewers independently converged on the shipped shape.

## Recall

Fence-only downgrades, confined to prefixed findings whose both spellings are in-diff paths
(requires a real top-level `a/`/`b/` file and its sibling in one diff). Rides the wave-end
paired mini per #101 measurement policy; merges on green suites.

Two further deltas inside that same collision class, both surfaced by the round-1 differential
sweep (102 payload deltas over 4 diff shapes × 6 spellings × 4 lines × 5 end_lines; zero
COMMENT_COUNT delta, zero review-body delta):

- `post_github`'s multiline check now runs against the resolved path, so a range that
  previously validated by mixing two files' line sets posts a single-line anchor on the
  resolved file instead of a multi-line anchor naming lines the PR does not have (pinned by a
  dedicated test).
- In a collision, the ambiguity downgrade wins over later, more specific reasons
  (`range_not_in_diff`, `anchor_mismatch`, the #219 span rename), so the reason tally reads
  `no_diff_oracle` for those findings. Outcome identical (fence closed either way);
  diagnostic-only.

## Validation

Two adversarial opus rounds plus an opus verification round, each in an isolated detached
worktree, on top of the 2-opus design red-team. Round 1 found no blockers; its one major (a
pre-existing mixed-spelling test now green for the wrong reason — its collision fixture is
intercepted upstream by the new check) and eight minors were fixed and re-verified
finding-by-finding in the verification round.

Every regression test is mutation-proven red, with `__pycache__` purged before each run:
gutting `_fence_path_is_ambiguous` reds exactly the 8 collision tests; reverting
`post_github`'s resolved path reds exactly its 3 pins; deleting `_resolve_spelling`'s stripped
fallback reds the 4 pre-existing strip contracts plus the recall and residual pins; removing
the `stripped == raw_file` guard reds 74 tests; `finding.get("file", "?")` → `finding["file"]`
reds the file-less-finding test. A differential sweep against main (4 diff shapes × 6
spellings × 4 lines × 5 end_lines) showed zero COMMENT_COUNT delta and zero review-body delta;
a ~1400-combination differential over the four helpers showed 0 mismatches for the resolver
consolidation.

Suites: `tests/` 1642 passed (641 subtests), `bench/tests/` 540 passed (22 subtests).
Coverage: scripts 93.62% (floor 92.5), bench 88.39% (floor 87.5), both via the exact CI
commands; local scripts headroom is 1.12 pp — whether the floor ratchets is decided by this
PR's CI measurement per the AGENTS.md policy. `workflows/` is byte-identical to main, so no
bundle rebuild is owed. `pre-commit run --all-files` green.

https://claude.ai/code/session_01B5NvycqWLZhhBfqUfrTHhh

---

> [!NOTE]
> **Medium Risk**
> Changes review-delivery path resolution and the apply-check for one-click patches, so a bug could mis-anchor comments or drop/keep suggestion fences incorrectly. Scoped to a rare two-file collision shape and covered by new GitHub/GitLab tests.
> 
> **Overview**
> Stops one-click apply from targeting the wrong file when a finding path and its `a/`/`b/`-stripped form both name real files in the same diff (issue #229). The suggestion fence now fails closed (`no_diff_oracle`); the comment still posts on the unchanged anchor and the prose `suggestion` still ships.
> 
> Path lookup is centralized in `_resolve_spelling` / `_strip_ab_prefix`. GitHub comments now post the **diff’s own spelling**, so a prefixed finding that only validates after strip no longer 422s the whole review. Multi-line GitHub anchors also validate against that resolved path, so a range that mixed two files’ line sets becomes a single-line comment on the resolved file.
> 
> Prefixed findings with no colliding sibling still keep both anchor and fence. Coverage floor for scripts is ratcheted 92.5 → 92.6.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 04d8b2f051ca27aedb01a4a7c5bca67e5e8f426d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper (commented):
Code Gauntlet review: 0 findings across 7 dimensions (bugs, security, cross-file impact, tests, conventions, type design, simplification). Full report generated but no issues surfaced.

---
Generated by code-gauntlet | Reviewed up to: 04d8b2f051ca27aedb01a4a7c5bca67e5e8f426d

Issue #234, opened 2026-08-24
bench real-poster equality guard leaves two mirror axes uncovered: skipped-section body prose and fence-bearing bodies


leehopper:
## Problem

PR #233 established `TestRealPosterMatchesPayloadMirror`: the bench payload mirror is byte-compared against a live `post_review.main() --dry-run` capture, so mirror drift on the covered shapes cannot stay green. Two mirror surfaces remain outside that guard:

1. **Skipped-section body prose.** The real `post_github` renders `build_skipped_section` prose into the review body (`review_body += skipped_section + footer`); the mirror's `build_reference_github_payload` only appends raw strings to the top-level `skipped` list and composes no section. The module docstring now names this as the documented exception, and every equality fixture is constrained to fully-anchorable findings because of it.
2. **Fence-bearing bodies.** No equality case carries `suggested_fix_code`, so `_gated_finding`'s fence rendering and `_gitlab_anchored`'s `-m+n` offsets are pinned only fixture-side (mirror-generated fixtures), never against a live capture.

## Fix shape

Extend the equality class with (a) one GitHub case containing an unanchorable finding, pinning `build_skipped_section` rendering and the skip/downgrade interleaving order (this likely requires teaching the mirror to compose the section — the real-decision-consumption pattern of #232/#222 applies), and (b) one GitLab case whose finding carries a `suggested_fix_code` spanning a range that yields real `fence_offsets`, so fence bytes are compared to the real poster's body.

Surfaced by the PR #233 adversarial review (issue #232 work).

PR #236, opened 2026-08-25
feat(delivery): overlap-aware suggestion fence demotion (#223, folds #224)

### LINKED ISSUE #223: Overlap-aware suggestion fence demotion: overlapping multi-line apply ranges break batch apply on both platforms
## Problem

Two findings in the same file whose one-click apply ranges overlap cannot both be applied. GitLab refuses the whole batch apply when any two suggestions' line ranges overlap (`Gitlab::Suggestions::SuggestionSet` returns "Suggestions are not applicable as their lines cannot overlap."), and applying one serially outdates the other (`Suggestion#outdated?` compares the whole span's content). GitHub multi-line suggestions carry the same exposure. Before PR #222, GitLab fences were all single-line at distinct anchors, so overlap was near-impossible there; multi-line fences make it routine for adjacent findings.

## Decision to make

Either ratify the current behavior (each fence individually correct; the reviewer applies the survivor after the first apply outdates the rest) or add a poster-side demotion: when two kept fences' apply ranges overlap within one file, downgrade the later one to prose with its own closed-vocabulary reason. The check is a pure function over the kept set at each poster, same shape as the existing apply-check.

Surfaced by the #219 red-team (PR #222); accepted there as parity with the already-shipped GitHub multi-line path, recorded in delivery-guide.md.

### LINKED ISSUE #224: bench GitHub mirror duplicates post_github's multiline apply-range formula instead of calling it
## Problem

`bench/tests/test_adapter.py::_github_comment` copies `post_github`'s multiline apply-range formula rather than calling the real decision. If the poster's formula changes, the mirror stays green while measuring fiction — the exact failure its own docstring warns about.

## Fix shape

Same treatment PR #222 gave the GitLab side: hoist the GitHub apply-range decision into a named helper in `scripts/post_review.py` and have both `post_github` and the bench mirror call it, with a behavior test pinning the mirror to the poster (mutation: re-hard-code the mirror, watch the pin go red).

Surfaced during the #219 implementation (PR #222).

leehopper:
Closes #223. Closes #224.

## Decision

#223 asked for a decision: ratify per-fence correctness (the #219 deferral recorded in
delivery-guide.md), or demote the later of two overlapping kept fences. **Decision: demote.**
Evidence, from primary sources checked during design:

- GitLab refuses the whole batch apply when any two suggestions' ranges overlap in one file
  (`Gitlab::Suggestions::SuggestionSet` → `FileSuggestion#line_conflict?` via ActiveSupport
  `Range#overlaps?` — closed intervals, per `diff_file.file_path`), and a batch can include
  suggestions from every discussion on the MR, so one overlapping pair poisons every suggestion
  on the MR. Serially applying one outdates the other (`Suggestion#outdated?`).
- GitHub has no clean refusal: a field report (parkerbxyz/suggest-changes#121) shows batched
  multi-line + adjacent suggestions silently corrupting content, with hard API errors in other
  constructions. The often-quoted "only one suggestion per line per batch" has no traceable
  GitHub primary source and is deliberately not relied on.

The gate's existing harm-asymmetry doctrine decides the trade: the demoted finding still ships
with its prose `suggestion`; only the one-click affordance is withheld.

## Mechanism

- 14th closed apply-check reason `overlaps_kept_fence` — the one member that names a SET-level
  decision rather than a property of the finding alone. Both vocabulary pin tests updated.
- Pure resolver `_overlap_losers` over records `(index, path_lookup, apply_range)`: first-wins
  greedy in delivery order (= the pipeline's rank order), per-path closed-interval intersection
  `max(l1,l2) <= min(e1,e2)` — GitLab's own `Range#overlaps?` exactly: same-line single-line
  fences collide, touching disjoint ranges do not. A loser occupies nothing. Greedy is
  pairwise-disjoint, not maximum-cardinality (A[1,10] B[5,6] C[8,20] keeps only A): priority
  beats count, deliberately.
- Candidate records built by module-level `_github_overlap_records` / `_gitlab_overlap_records`
  — a candidate is a group primary that is a dict, carries `suggested_fix_code`, anchors on a
  diff line, and passes the SAME pure verdict the render site applies (`_fence_verdict`, which
  owns the `path_lookup` derivation) at the SAME apply range (`_github_apply_range` /
  `_gitlab_apply_range`). "Candidate" and "would render a kept fence" are one computation.
- Silent pre-pass at both posters (no warnings, no tallies): the demoted set is a pure function
  of findings + diff — rerun-stable, dry-run == live, independent of prior-delivery state and
  live POST outcomes. Losers demote at the existing render sites through `_gated_finding`'s new
  keyword-only `demote_reason`, which is consulted only on a gate PASS and falls through the
  existing closed-vocabulary guard → tally → warn → strip tail (a typo'd demote_reason raises,
  pinned).
- GitLab's reactive fence site (`deliver_corroborator`, reachable under --dry-run via
  `validate_position`'s pure "invalid" outcome) queries the pre-pass's read-only kept-interval
  map and never claims an interval.
- #224 folded: `_github_apply_range` is extracted from `post_github`'s loop and the bench
  mirror's `_github_comment` now calls it instead of duplicating the multiline formula; the
  mirror's overlap decision likewise calls the real record builders + resolver + funnel (never a
  hand strip, so the payload `skipped` warning lines match byte-for-byte).
- Two new real-poster equality fixtures (`github_overlap_demotion.json`,
  `gitlab_overlap_demotion.json`), unstamped findings, each with a leading index-splitter
  finding so re-keyed records demote the wrong member and go red; hand-typed literals pin the
  warning line, a demoted body, and the touching-disjoint both-kept pair as the third oracle.
- delivery-guide.md's #219 ratification paragraph replaced by the demotion rule.

## Deviations from the adjudicated design, deliberate

- `_gated_finding` uses `elif` where the design memo's R3 snippet fell through the
  anchor-mismatch rename: a set-level `demote_reason` is never renamed by `mismatch_reason`
  (documented in the docstring). Behaviorally identical for every current caller.
- The GitLab index-basis test uses a fence-less line-valid splitter, not the off-diff splitter
  first proposed: an off-diff finding is filtered before `remaining` exists, so it cannot create
  the ordinal-vs-index gap the test needs (the off-diff axis is pinned by the GitLab fixture's
  leading finding instead).

## Residuals, named and out of scope

- Cross-run overlap: a fence posted by a PRIOR run is invisible to this run's resolver on both
  platforms (GitHub fetches no existing comments; GitLab delivery keys carry no ranges).
- GitHub's adjacent-range batch-apply corruption class is broader than index overlap; the
  demotion applies GitLab's source-verified closed-interval semantic, the only one confirmed
  from a primary source. Stated as a scope rule in delivery-guide.md without the incident
  narrative.
- Corroborator-vs-corroborator overlap on GitLab's reactive paths (a reactive site cannot see a
  sibling it has not been called for yet), and the mirror image: a winning primary whose own
  discussion is lost late (malformed position / rejected POST) still occupies its interval, so a
  reactive corroborator can lose its affordance to a fence that never landed — the map
  deliberately never depends on live outcomes.
- An invalid-position survivor still occupies its interval (deterministic; errs toward
  withholding an affordance).
- The demotion WARNING (not the decision) is render-site-bound: GitLab live-rerun branches that
  skip rendering also skip the warning/tally — the same delivery-state dependence
  `_print_fix_summary` already documents for every other reason.
- `scripts/report_patches.py` deliberately does not run the overlap pass: the loser set is a
  per-poster property (the candidate sets can differ across platforms above the GitLab offset
  cap) and the artifact has no poster to speak for. Its disclosure line and module docstring now
  name the set-level withholding alongside the render-site constraints, so the honesty prose
  matches the new reality.

## Process

2-sonnet + 1-web recon (platform claims verified against GitLab source; the GitHub folklore
claim rejected) → design memo → 2-opus red-team (REVISE→adopted: candidate predicate must
require the field — a fence-less finding must never claim an interval; silent pre-pass
constraint; the three pure single-source units; index-basis rule; scoped doc claims) → sonnet
TDD implementation → 2-opus adversarial round 1 (3 major: unpinned partial-delivery threading,
unpinned loser-exclusion guard, mirror re-spelled predicate → record builders hoisted; plus
index-basis pins and fixture index-splitters) → fixes → 2-opus adversarial round 2 (both
verdicts SHIP; polish round: report_patches disclosure widened, fixtures regenerated in the
directory's raw-UTF-8 capture convention, a load-bearing-only-for-one-shape pre-pass guard made
falsifiable, three docstrings corrected). ~35 one-at-a-time mutations observed red with
`__pycache__` purges and byte-verified restores across the rounds; the STAYED-GREEN mutations
found in review each gained the regression test or pin that now catches them.

## Validation

- `python -m pytest tests/ -q` — 1678 passed, 641 subtests.
- `python -m pytest bench/tests/ -q` — 553 passed, 31 subtests.
- Coverage gates green. CI measurements of record from this PR's first run: scripts 93.67
  (1.07 pp over the 92.6 floor → ratcheted to 92.7 in-PR per the AGENTS.md policy, with the
  byte-budget constant raised 20_700 → 20_761 for the provenance sentence, same shape as the
  #219/#231 raises); bench 88.29 (0.79 pp headroom, floor stays 87.5).
- Every pre-existing adapter fixture and the #234/#235 warning-interleave literal byte-identical;
  `workflows/` untouched (bundle diff clean).
- Recall/noise: nothing bench-scored changes for findings sets without overlapping kept fences
  (zero losers → byte-identical output); the overlap case swaps a fence for prose in delivered
  bodies — delta noted for the wave-end paired mini.

https://claude.ai/code/session_01F7Qa7zNRepQ5b7G2D9gAvw

---

> [!NOTE]
> **Medium Risk**
> Changes core review posting behavior for multi-fence PRs (which fence survives vs prose-only), with broad test coverage but real user-visible delivery differences when ranges overlap.
> 
> **Overview**
> **Delivery now withholds one-click `suggestion` fences when their apply ranges would collide with an earlier kept fence in the same file**, using GitLab-style closed-interval overlap and first-wins delivery order. Losers downgrade through the existing `_gated_finding` path with reason `overlaps_kept_fence` (prose `suggestion` still ships).
> 
> `post_review.py` adds a silent pre-pass on GitHub and GitLab (`_overlap_losers`, shared `_fence_verdict` / apply-range helpers), threads `demote_reason` into render sites, and on GitLab lets reactive `deliver_corroborator` paths consult a read-only map of winning intervals. **#224**: `_github_apply_range` is centralized so the real poster and bench payload mirror call the same logic.
> 
> **Tests and fixtures**: large expansion in `tests/test_post_review.py`, new GitHub/GitLab overlap dry-run JSON fixtures, and bench mirror builders wired to the real record builders. **Docs/process**: delivery guide and report schema note overlap withholding; `report_patches.py` states it does not run the overlap pass; Python CI coverage floor **92.6 → 92.7** and matching `AGENTS.md` provenance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e5dbe603ce6592d1f87f746fcc45079b8b58722. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


skills/code-gauntlet/references/delivery-guide.md
```diff
@@ -53,9 +53,21 @@ mechanism that keeps that promise now that agents populate the field too.
...
+**A fence that would overlap another kept fence in the same file demotes to prose.** GitLab
+refuses to batch-apply a set of suggestions with any two overlapping ranges in one file, and
+serially applying one outdates the other — so before rendering, the poster walks the fences in
+delivery order (the pipeline's own priority ranking, not line position) and withholds any whose
+apply range overlaps one it has already kept in that file: the earlier fence wins, and a withheld
+fence claims no range of its own. Overlap uses GitLab's closed-interval semantic: two single-line
+fences anchored at the identical line collide; two fences that merely touch (`[n, m]` next to
+`[m + 1, k]`) do not. A demoted finding still ships with its prose `suggestion` — only the
+one-click affordance is withheld — and which of a run's ranked fences are withheld is a pure
+function of the findings and the diff, so a rerun reaches the same verdict regardless of what has
```
leehopper, line 66: **💡 [LOW] New overlap-fence demotion mechanism not reflected in report-format.md's suggested_fix_code gating description**

This PR adds a new set-level gate (_FIX_OVERLAPS_KEPT_FENCE via _overlap_losers/_gated_finding's demote_reason) that strips suggested_fix_code independently of _suggested_fix_gate, and updated delivery-guide.md to describe it, but skills/code-gauntlet/references/report-format.md still enumerates only the deterministic per-finding apply-check conditions as the reasons the fence can be stripped, so a reader relying on that doc will not learn that a passing fence can also be withheld because a higher-priority fence already claimed an overlapping range in the same file.

**Suggested fix:**
Add a sentence to report-format.md's suggested_fix_code description noting that, independent of the deterministic apply-check, a fence can also be withheld when its apply range overlaps an already-kept fence in the same file (reason overlaps_kept_fence), mirroring the paragraph just added to delivery-guide.md.

scripts/post_review.py
```diff
@@ -1150,24 +1194,20 @@ def _gated_finding(
...
     if reason not in _FIX_REASONS:
-        # A typo'd reason string in a future gate edit must fail loudly at the
-        # first downgrade, not silently record garbage in the stable warning
-        # line (whose readers rely on the vocabulary being closed).
+        # A typo'd reason string in a future gate edit — or a typo'd
+        # demote_reason from a future overlap-demotion caller — must fail
+        # loudly at the first downgrade, not silently record garbage in the
+        # stable warning line (whose readers rely on the vocabulary being
+        # closed).
         raise ValueError(f"_suggested_fix_gate returned an unknown reason: {reason!r}")
```
leehopper: **💡 [LOW] ValueError message still blames _suggested_fix_gate even though the comment now says demote_reason can also trigger it**

This PR edited the comment above the raise to explicitly acknowledge that the unknown-reason error can now come from either a typo'd gate reason OR a typo'd demote_reason argument, but left the exception message text unchanged, so it always says '_suggested_fix_gate returned an unknown reason' even when the bad value was demote_reason and _suggested_fix_gate never returned it at all — a future debugger following the message would look in the wrong function.

**Suggested fix:**
Reword the message to reflect both sources, e.g. f"unknown downgrade reason: {reason!r}" or branch the message based on whether the failure came from the gate or from demote_reason.

```suggestion
        # A typo'd reason string in a future gate edit — or a typo'd
        # demote_reason from a future overlap-demotion caller — must fail
        # loudly at the first downgrade, not silently record garbage in the
        # stable warning line (whose readers rely on the vocabulary being
        # closed).
        raise ValueError(f"unknown downgrade reason: {reason!r}")
```

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: eb828d05fe642745f53c9a0e26048be8776398ca

skills/code-gauntlet/references/delivery-guide.md
leehopper: Fixed in 9e5dbe6 — report-format.md's table row and the field paragraph both name the set-level overlap withholding (`overlaps_kept_fence`) now, and the patches-artifact disclosure matches report_patches.py's widened wording.

scripts/post_review.py
leehopper: Fixed in 9e5dbe6 — the message now reads "unknown downgrade reason: {reason!r} (from the gate, a mismatch_reason rename, or a caller's demote_reason)".

PR #237, opened 2026-08-25
feat(schema): dimension-conditional dispatch requirement gated to first-party-direct runs (#218)

### LINKED ISSUE #218: Conditional per-dimension required for the multi-dimension dispatch (nested if/then measured as enforced on first-party API)
## Context

PR #217 (#66) ships `requiredExtra` with intersection semantics, so `claude_md_rule` (convention) and `spec_text` (intent) stay contract-enforced: the conventions-and-intent dispatch mixes three dimensions in one schema, and a flat `required` list cannot express a per-dimension requirement there.

## Measurement (2026-08-18, CLI 2.1.235, first-party API)

Two live probes during the #66 work:

- A top-level `allOf` in the agent schema is rejected at dispatch: `API Error: 400 tools.N.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level`.
- The same `allOf`/`if`/`then` nested inside the findings array's `items` is **accepted and enforced**: a probe agent instructed to return `{"findings": [{"dimension": "convention"}]}` had that first attempt rejected by the tool-layer validator with `must have required property 'claude_md_rule', /findings/0: must match "then" schema`, and complied on retry.

So a conditional block like

```json
"allOf": [{"if": {"properties": {"dimension": {"const": "convention"}}, "required": ["dimension"]}, "then": {"required": ["claude_md_rule"]}}]
```

inside the item schema could enforce the two remaining prose-only REQUIRED fields.

## Why #66 did not ship it

- Unmeasured on Bedrock/Vertex/Foundry. Those providers validate tool schemas independently, and an unsupported construct 400s the dispatch — which degrades every dimension the agent owns (the #179 failure class: observed live, a Bedrock run degraded all discovery dimensions in 2s on an invalid model ID).
- Enforcement semantics under sustained conditional violations (retry cost at cap 5) are unmeasured at scale; #66 requirement 3's retry-storm rule applies.
- The wave is suites-only per-PR; a schema-construct change of this class needs at least a smoke watch item (the #53 precedent) and ideally provider-matrix evidence.

## Decision surface

1. Measure: does Bedrock/Vertex accept nested `if`/`then` in tool input_schema, and does the harness enforce it there?
2. If portable: extend `requiredExtra` (or a sibling key, e.g. `requiredWhenDimension`) to emit the conditional block for multi-dimension agents, with a smoke watch item on conventions-and-intent dispatch success.
3. If not portable: either gate emission on `policy.provider === 'firstParty'` (two schema shapes to maintain — weigh against the value) or close as not-needed and keep contract enforcement.

Evidence details (probe transcripts) are in the #66 session; the decisive validator error strings are quoted above verbatim.

leehopper:
Closes #218.

## Decision

Option 3a from #218's decision surface: the nested `allOf`/`if`/`then` conditional-required construct ships, emitted only on first-party-direct runs — `policy.provider` absent/null/`firstParty` AND not `policy.gateway`. Third-party providers (bedrock/vertex/foundry) and gateway sessions keep the byte-identical flat schema; contract prose stays the enforcement floor everywhere.

Evidence of record:

- **First-party** (measured 2026-08-18, CLI 2.1.235, quoted in #218): top-level `allOf` is rejected at dispatch; the same construct nested inside `items` is accepted and enforced by the tool-layer validator (`must have required property 'claude_md_rule', /findings/0: must match "then" schema`, complied on retry).
- **Bedrock: unmeasurable in this environment** (2026-08-25). `aws sts get-caller-identity` → `NoCredentials`, zero configured profiles; the AWS MCP proxy is also unauthenticated (`failed due to expired or invalid AWS credentials`). Two independent auth paths, both dead.
- **Vertex: unmeasurable.** No gcloud SDK installed, no token path.
- **Docs/community research** (2026-08-25): Bedrock and Vertex return byte-identical Anthropic error strings for other `input_schema` violations, including Anthropic's internal `tools.N.custom.*` path shape (block/goose#2150, BerriAI/litellm#31582, vercel/ai#13167; Vertex's Claude docs defer to Anthropic's own JSON-schema-limitations page). That is consistent with passthrough or faithful replication of Anthropic-side validation — but there is no direct evidence, either direction, for nested `allOf`/`if`/`then` on any third-party provider. Plausible-but-unmeasured stays gated out. This research is the basis for widening the gate later if a live probe becomes possible.
- The #179 failure class makes unmeasured emission unacceptable: one provider-side 400 degrades every dimension the agent owns — three, for conventions-and-intent.

Why not "close as not-needed": `claude_md_rule` (convention) and `spec_text` (intent) are per-dimension unconditional in the contract — REQUIRED within the owning dimension, OMIT for siblings — the same emission discipline that earned the five #66 `requiredExtra` promotions. Only the flat-list limitation kept them prose-only.

## Mechanism

- `registry.js`: new per-row key `requiredWhenDimension` (`['claude_md_rule']` on convention, `['spec_text']` on intent, `[]` elsewhere). An entry is either a key of the row's own `schemaExtra` or a canonical field that is neither in `FINDING_REQUIRED` nor pipeline-stamped (`origin` is inadmissible — no agent emits it).
- `stages.js` `agentSpecs()` derives `spec.conditionalRequired` (sorted `{dimension, required}` entries, defensive reads for synthetic rows); `findingItemSchema` appends exactly the measured construct when non-empty; `findingSchema(spec, policy)` gates it through `conditionalSchemaActive(policy)`.
- `conditionalSchemaActive` = `pinsModelIds(provider) && !policy.gateway`. It is deliberately NOT the model-pin predicate alone: a gateway translates model names (the pin's risk terminates there) but forwards `input_schema` verbatim to whatever backend it fronts — a LiteLLM-class gateway pointed at Bedrock would deliver the construct to the exact surface we could not measure, and `CLAUDE_CODE_SUBAGENT_MODEL` cannot relieve schema risk.
- `policy.gateway` (boolean) is stamped by Phase 2 from `ANTHROPIC_BASE_URL` set non-blank (whitespace-trimmed); validated at the waist like `provider`. It rides alongside the provider enum rather than inside it because it answers a different question.
- Envelope: `resolvedPolicy` now reports `{subagentModel, provider, gateway, conditionalSchema}` — `conditionalSchema` is the same predicate the dispatch gate used, so the envelope can never disagree with the dispatch. The pre-existing exact-shape pin (the #178 follow-up test) was updated knowingly to the four-key object, not loosened to key-presence.
- Discover's null-dispatch gap line notes when the failed dispatch carried the conditional construct; Phase 8 methodology disclosure derives from `resolvedPolicy.conditionalSchema`.
- Docs lockstep: both report-format.md Required columns are now tri-state (`yes`/`conditional`/`no`) with registry-derived pytest lockstep on both tables; the contract prose uses the registered phrase "dimension-conditional dispatch requirement" with its own lockstep leg (claims must equal the registry union); both dispatch-schema phrases are now registered in docs/machine-parsed-strings.md (the pre-existing one had no row).

Why a sibling key instead of reusing `requiredExtra`: per-row reuse breaks that key's ecosystem — the sibling-parity guard hard-fails divergence, the intersection drops the fields from the flat list, the no-OMIT lockstep flags both fields (their contracts carry OMIT branches for sibling dimensions by design), and `all_required_extras()` assumes single-dimension rows. A conditional field needs the sibling-dimension OMIT branch that `requiredExtra` forbids.

## Design decisions of record

- **Dimension enum rejected.** An enum on `dimension` (to close the case-variant escape — `const` is case-sensitive and downstream consumers lowercase) was designed, implemented, and then removed on red-team evidence: filterFindings' own normalization sets tolerate `comment-accuracy`, `documentation`, and `doc-accuracy`, proving agents emit variant spellings. An enum would turn an observed, tolerated spelling into a whole-dispatch schema violation (all three dimensions degrade) to close a narrower fail-open escape.
- **Canonical fields admissible in the new key.** `claude_md_rule` is canonical (all seven contracts instruct it), so the naive "own-schemaExtra only" rule could not hold it; the canonical Required-column lockstep was extended to tri-state to keep the required-ness documented where that table's derivation looks.
- **Phrase chosen non-superstring.** A draft phrase ("conditionally required by the dispatch schema") contained the existing registered phrase as a substring and hard-failed the reverse-leg lockstep, which is a deliberately dumb substring scan; the shipped phrase shares no containment and needs no cross-exclusion, pinned by an `assertNotIn` regression test.
- **Gateway loses an enforcement it never had** — the fail-safe direction. Same class of acceptance as #179's model-ID decision, but through the schema gate's own signal.
- The agents/AGENTS.md byte-budget ratchet rose 20 761 → 21 004 for the new required-semantics bullet, per that test's own convention.
- ajv `strictRequired` note (record, not code): that non-default option flags a `required` naming a property undeclared in the same subschema — exactly this construct's shape. If a validator ever rejects it, that is the signature; do not pre-emptively add `properties` to the `then` blocks, which would un-measure the accepted spelling.

## Accepted risks

- **Relabel escape**: a model failing the conditional can satisfy the schema by relabeling a convention finding as comment_accuracy (routed to SUGGESTION, singleton confidence penalty applies). Structurally undetectable; likelier than a five-retry storm; mitigated by contract prose, measured by the wave-end paired mini.
- **Case-variant escape**: a `Convention`-spelled finding escapes the `const` match fail-open — today's exact behavior, contract enforcement only (see the enum rejection above).
- **Retry cost at cap 5**: the contract already mandates emission, the fields are quotable by construction, and the measured probe complied on the first retry (#66 requirement 3 weighed).
- **Two schema shapes**: both pinned by tests; the difference is one derived block behind one predicate.

## Measurement plan

- **Smoke B watch item** (observables only): conventions-and-intent dispatch non-null, zero degradation among convention/intent/comment_accuracy. A compliant smoke cannot distinguish enforcement from silent construct-dropping, so enforcement is not a smoke claim.
- **Wave-end paired mini** carries the enforcement/recall question (a convention finding the model drops rather than cite a rule for is recall loss, the #62 measured class). Precondition: assert `resolvedPolicy.conditionalSchema === true` in child envelopes before scoring — a bench machine with `ANTHROPIC_BASE_URL` set would otherwise measure the flat schema while looking green.

## Process

Recon (GitHub + codebase scouts) → live provider probes (both unmeasurable) + docs research → design memo → two opus red-team rounds (round 1: 2 breaks, 4 weakens; round 2 on the deltas: 2 breaks, 1 recommendation adopted — the enum removal) → implementation → 14 whole-mechanism mutations verified red (11 mechanism + 3 on the repaired contract guard), zero survivors → 6-dimension adversarial fan-out (sonnet finders, opus skeptic per finding): 5 findings, 5 confirmed, collapsing to 3 distinct defects, all fixed in-branch (vacuous OMIT-half contract guard, stale machine-parsed-strings note, `policy.gateway` missing from the waist-shape docs plus a new structural pin for both docs following the contextLines precedent).

## Gates

pytest tests/ 1685+ passed (93.74% ≥ 92.7 floor), bench 553+ passed (88.48% ≥ 87.5), node --test 890 passed (98.96 / 86.77 / 98.48 ≥ 98 / 85.7 / 97.4, presence check clean), bundle rebuilt byte-identical, pre-commit 14/14. Floors untouched (headroom within policy).

## Adjacent

One follow-up issue to file: registry-generated contract sentences — the OMIT/REQUIRED-per-dimension facts live in prose in the contract template and are now restated in the registry with a lockstep test keeping the restatement honest; the one-edit shape is a single declaration with the contract sentence generated the way the CLAUDE.md twins are.

https://claude.ai/code/session_01YbTRxvFWjQkZnEvSykrySi

---

> [!NOTE]
> **Medium Risk**
> Changes discovery JSON Schema shape and enforcement for a multi-dimension agent on first-party runs; mis-stamped `policy.gateway` could silently skip schema enforcement through a proxy, though contract prose still applies.
> 
> **Overview**
> Adds **`requiredWhenDimension`** on registry rows so **`claude_md_rule`** (convention) and **`spec_text`** (intent) can be schema-required only for their owning dimension on the shared conventions-and-intent dispatch, without forcing sibling dimensions to fabricate them.
> 
> On **first-party-direct** runs (`conditionalSchemaActive`: first-party provider and **`policy.gateway` false**), discovery builds nested **`allOf`/`if`/`then`** on finding items from `agentSpecs().conditionalRequired`. **Bedrock/Vertex/Foundry** and **gateway** sessions (`ANTHROPIC_BASE_URL` → **`policy.gateway`**, validated in **`args.js`**) keep the prior flat schema; agent contract prose remains the enforcement floor.
> 
> Phase 2 skill/triage docs now stamp **`policy.gateway`**; **`resolvedPolicy`** exposes **`gateway`** and **`conditionalSchema`**; methodology and discover failure gaps call out when conditional schema was in play. Docs and tests add tri-state **required** (`yes` / **conditional** / `no`), the phrase **"dimension-conditional dispatch requirement"**, and lockstep guards mirroring **`requiredExtra`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d3ab9f5f4222313253aafe461cfe1eeff3db914. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


workflows/src/stages.js
```diff
@@ -320,6 +320,17 @@ export function agentSpecs(dims = DIMENSIONS) {
   return order.map((a) => {
     const spec = byAgent.get(a);
     spec.requiredExtra = intersectRequiredExtra(spec.rows);
+    // conditionalRequired (issue #218): one { dimension, required } entry per row that
+    // declares a non-empty requiredWhenDimension, sorted by dimension so the derivation
+    // never depends on DIMENSIONS row order. Defensive `|| []` on the read — synthetic rows
+    // built for tests (registry.test.js) omit the key entirely, and this must not throw.
+    // Empty on every single-dimension spec (and on the two of conventions-and-intent's three
+    // rows that carry none) by construction, so findingItemSchema below emits nothing extra
+    // for them regardless of policy.
```
leehopper: **💡 [LOW] agentSpecs() comment miscounts conventions-and-intent rows with empty requiredWhenDimension**

The comment claims conditionalRequired is empty 'on the two of conventions-and-intent's three rows that carry none' (i.e. carry an empty requiredWhenDimension), but the DIMENSIONS array in the same file shows only comment_accuracy has an empty requiredWhenDimension — convention (['claude_md_rule']) and intent (['spec_text']) both carry non-empty entries. The correct count is one row, not two.

**Suggested fix:**
Correct the comment to say 'the one of conventions-and-intent's three rows — comment_accuracy — that carries none', matching the actual DIMENSIONS data.

```suggestion
    // Empty on every single-dimension spec (and on the one of conventions-and-intent's three
    // rows — comment_accuracy — that carries none) by construction, so findingItemSchema below
    // emits nothing extra for them regardless of policy.
```

workflows/src/args.js
```diff
@@ -620,6 +630,14 @@ export function validateArgs(args) {
...
     errors.push(`invalid policy.provider: ${args.policy.provider} (expected one of ${POLICY_PROVIDERS.join(', ')})`);
   }
+  // policy.gateway gates registry.js's conditionalSchemaActive (issue #218) — a non-boolean
+  // would silently coerce (e.g. the string "false" is truthy), so shape-check it the same
+  // way provider is above. Absent/null both mean "no gateway", same tolerance as provider.
+  if (args.policy && typeof args.policy === 'object' && !Array.isArray(args.policy)
+    && args.policy.gateway !== undefined && args.policy.gateway !== null
+    && typeof args.policy.gateway !== 'boolean') {
+    errors.push(`invalid policy.gateway: ${args.policy.gateway} (expected a boolean)`);
+  }
```
leehopper: **💡 [LOW] policy.gateway validation duplicates the inline object-shape check instead of reusing isPlainObject**

The new policy.gateway guard re-derives the plain-object check inline (args.policy && typeof args.policy === 'object' && !Array.isArray(args.policy)) instead of calling the isPlainObject helper already defined at line 310 and already used at line 333 for the exact same purpose, so the file now carries the same three-clause check written out twice (once pre-existing for provider, once newly added for gateway).

**Suggested fix:**
Replace the inline shape check with the existing helper: Before: `if (args.policy && typeof args.policy === 'object' && !Array.isArray(args.policy) && args.policy.gateway !== undefined && args.policy.gateway !== null && typeof args.policy.gateway !== 'boolean') {`. After: `if (isPlainObject(args.policy) && args.policy.gateway !== undefined && args.policy.gateway !== null && typeof args.policy.gateway !== 'boolean') {`

```suggestion
  if (isPlainObject(args.policy)
    && args.policy.gateway !== undefined && args.policy.gateway !== null
    && typeof args.policy.gateway !== 'boolean') {
    errors.push(`invalid policy.gateway: ${args.policy.gateway} (expected a boolean)`);
  }
```

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: ef2bc8f3d1073952e5adaf7f25f8fa14302aded1

workflows/src/stages.js
leehopper: Fixed in 0d3ab9f — comment now names comment_accuracy as the single row carrying no entry.

workflows/src/args.js
leehopper: Fixed in 0d3ab9f — the gateway guard now reuses isPlainObject.

Issue #248, opened 2026-08-26
A malformed (non-object) replayed challenge checkpoint yields a silent ok:true empty review


leehopper:
## Problem

`validateArgs` never shape-checks `args.checkpoints`, and `runPhase` returns a replayed
checkpoint verbatim. A non-object `checkpoints.challenge` (string, number, boolean) flows
through the challenge phase untouched: `(challengeOut.findings || [])` evaluates over a
primitive, every downstream consumer sees zero findings, and the run returns `ok:true` with an
empty review and a normal artifact set. A corrupt or truncated checkpoint file therefore
reads as "review complete, nothing found".

Surfaced by the #213 round-2 review: the replay belt's in-place rewrite briefly turned this
route into a raw `TypeError` (`ok:false`, no gap), which was reverted to preserve the existing
envelope. The preserved behavior is itself the defect this issue tracks.

## Fix shape

Same family as #178 (all-degraded must not return `ok:true`): fail loud with a named gap when
a replayed checkpoint is not the shape the phase contract requires (e.g.
`challenge checkpoint is not an object — refusing to replay it`), rather than degrading into a
silent empty review. Needs a decision on which phases' checkpoints get shape-gated and what
the failure envelope carries. Suites-only.

## Evidence

Measured on main and on the #213 branch (2026-08-26): `validArgs({checkpoints: {challenge:
'corrupt-string'}})` through `runWith` → main: `{ok: true, highConfidence: 0, artifactPaths:
{...}}`. The #213 branch pins this envelope with a regression test; this issue is the release
valve for changing it deliberately.

leehopper:
## Executed 2026-08-31 — PR #264 MERGED (with #250 and the #247 decline in one round)

Decision: shape-gate at the pre-dispatch boundary, one policy for the whole checkpoint-shape class, decided with #250 as this issue asked.

[...]

Issue #247, opened 2026-08-26
apply_exclusions never scans claude_md_rule/spec_text, so the user kill-switch misses vocabulary that lives only in the citation


leehopper:
## Problem / decision

The #213 pass (citation-field injection strip) deliberately did NOT widen `apply_exclusions` to
`claude_md_rule`/`spec_text`. Decision of record: an exclusion match eliminates the whole
finding, and user patterns matching common rule phrasing (`MUST`, `CLAUDE.md`) would
mass-eliminate the conventions dimension — a strictly worse failure mode than the strip's
citation-only cost.

The counter-case is real, though. Both twins justify including `suggestion` in the exclusion
combine with "user-authored ignore patterns are the user's kill-switch over everything that
gets rendered" — and the citation fields are rendered. A conventions finding titled "Violates a
documented project convention" can carry its identifying vocabulary *only* in
`claude_md_rule` (e.g. `"Every public function must have a docstring"`). A user who stops
enforcing docstrings adds `docstring` to the REVIEW.md ignore list and it silently does
nothing: the finding matches in no scanned field.

## Fix shape (if adopted)

Add the two fields to the exclusion combine in both twins (the scanned-field surface is now
the shared strip-field list from #213, so this is a one-line widening per twin), re-record the
affected parity fixtures, add a citation-only-match golden. `re.ASCII` stays off this path
(user patterns, unicode folding pinned by `case_fold_unicode_cafe`).

## Why deferred

Changes which findings are eliminated — recall/noise-sensitive, rides a paired measurement
window, same class as #215/#117. Needs the whole-finding-elimination cost weighed against the
kill-switch gap, ideally with a measurement of how often real ignore patterns would collide
with citation text.

Surfaced by the #213 red-team (adjacent-issue discipline).

leehopper:
## Decided 2026-08-31: DECLINED (ratified on measurement), pinned structurally in PR #264

The #213 pass declined this widening on the cost asymmetry; this pass ran the measurement the issue asked for and the measurement ratifies the decline. Both twins' tests now pin the surface (a citation-only match keeps the finding), so the decline is enforced, not prose.

### Measurement of record (local bench corpus: 512 real findings / 48 files, 2026-08-31)

[...]

PR #249, opened 2026-08-26
feat(filter): scan claude_md_rule/spec_text with the injection sets, strip on match (#213)

### LINKED ISSUE #213: claude_md_rule and spec_text are rendered into posted comments without passing the content filters
## Problem

`post_review.py` renders `claude_md_rule`/`spec_text` under **Cited rule:** (sanitized + redacted
+ capped at 500), but neither filter twin scans them — same gap class as #62's `suggestion`, and
these two fields are repo-derived (CLAUDE.md / spec text under review), a strictly higher
injection-risk source than agent-authored advice. `_sanitize_outbound_prose` strips HTML
comments/invisibles/backtick runs but not "visit https://…" lures or bypass instructions.

## Fix shape

The field-level strip mechanism #62 introduced (#210) extends naturally: scan the two fields
with the same seven sets, strip on match, keep the finding. Recall/noise-sensitive → Wave 5+ class.

leehopper:
Closes #213.

## Why

`claude_md_rule` and `spec_text` quote repo text verbatim into posted comments under **Cited rule:**. The render path sanitizes, redacts, and caps them, but none of the seven deterministic injection sets ever scanned them, and the report writer received them with no filter at all (`REPORT_EXCLUDED_FIELDS` omits both; `agents/report-writer.md` carries no injection guidance). On a fork PR the working tree's CLAUDE.md is attacker-authored, so the citation is an attacker-controllable channel into a posted comment. Same gap class as #62's `suggestion`; the #210 strip mechanism extends to close it.

## What

- Both filter twins now scan the shared field list `_INJECTION_STRIPPED_PROSE_FIELDS = ("suggestion", "claude_md_rule", "spec_text")` with the same seven sets: strip the matching field, keep the finding, stamp `{field}_removed_by="injection"` and `{field}_removal_reason`. Non-string values strip unconditionally. All matching fields strip; suggestion-path bytes are unchanged.
- `suggested_fix_code` strip now propagates on a pattern-strip of ANY scanned field (first field in list order names the reason `{field} carried {phrase}`). The review round established that stripping a hostile citation while keeping its one-click patch deletes the disclosure, not the payload; the propagation closes that.
- Replay belt in `stages.js`: findings replayed from a challenge checkpoint pass through the strip before delivery selection, report input, `writeArtifacts` (which `findings.json` and the derived post-review projection are built from), persisted checkpoints, and the eliminated set. The strip is pure and idempotent (fuzzed 0/2400 non-idempotent), so live runs are unaffected. A non-object replayed checkpoint keeps main's tolerant envelope; that pre-existing silent-empty-review behavior is now pinned by a test and tracked in #248.
- Receipt stats gain `claude_md_rules_removed`/`spec_texts_removed` in both twins, spliced from a map derived off the field list, guarded by an emission assertion that loops the list (a future field is covered with one edit).
- `parity.test.js`'s removal-reason tolerance generalized from the hardcoded `suggestion_removal_reason` to every listed field (Python `repr(rx.pattern)` vs JS `JSON.stringify(rx.source)` diverge by construction).
- Four new parity goldens (`claude_md_rule_scan`, `spec_text_scan`, `citation_non_string`, `citation_phrase_propagates_fix_code`); every pre-existing golden byte-identical.
- Docs: module docstring receipt schema, `validation-pipeline.md` filter bullet, exclusion-combine comments in both twins, boundary-parity seed pinning.

## Decisions of record

1. **All seven sets scan the citation fields** (the issue's fix shape). Measured before choosing: 775 rule-text units segmented from 12 real CLAUDE.md/AGENTS.md files fired 5/698 prose units (0.72%) against the full union while catching 14/14 hand-written lures; every narrowed policy missed 5 to 7 lures. Caveat: the corpus contains zero literal URLs, so the url set's 0% false-fire is under-tested. The wave-end paired mini carries the aggregate recall/noise question, as the issue already planned.
2. **Propagation to `suggested_fix_code` from any scanned field** (reverses the initial design after two independent red-teams constructed the same attack). Strip cost is bounded: the finding always survives; only the citation blockquote and the fence drop.
3. **`apply_exclusions` deliberately NOT widened** to the citation fields: an exclusion match kills the whole finding, and user patterns matching common rule phrasing would mass-eliminate the conventions dimension. The kill-switch counter-case is filed as #247, not silently declined. `apply_exclusions` behavior is byte-unchanged in this PR.
4. **`REPORT_EXCLUDED_FIELDS` unchanged.** Removal-reason stamps embed `rx.pattern`/`rx.source` (first-party regex text), never attacker bytes, so there is nothing to exclude on injection grounds.
5. Stripping `claude_md_rule` activates the `spec_text` render fallback for findings carrying both; the fallback value is itself scanned, so nothing unscanned can surface through it.
6. Belt strips on a replayed checkpoint are disclosed by the per-finding stamps but not counted in `stats.filter` (same pre-existing shape as the #62 suggestion strip; noted in a comment at the belt).

## Validation

- 2-sonnet recon, 2-opus design red-team (REVISE, adopted: parity-comparator blocker, propagation reversal, measurement requirement), sonnet TDD implementation, two 2-opus adversarial review rounds (round 1 REVISE: the persist-path blocker where the derived post-review projection read unstripped `findings.json`; round 2 APPROVE x2 with polish items, all applied).
- Mutation ledger: 23/23 mutations red, one at a time, `__pycache__` purged, restores byte-verified.
- Gates re-verified independently on the final branch state: pytest 1765 passed (+723 subtests), bench 553 passed, node 944 passed, bundle byte-identical to a fresh build, `record_parity.py --check` all goldens fresh, pre-commit green.
- Local coverage: Python scripts 93.92 (floor 92.9), bench 88.48 (floor 87.5), JS 99.01/87.25/98.48 (floors 98/85.7/97.4). JS floors ratcheted in-PR from this PR's CI measurement (99.01/87.25/98.48): lines 98 to 98.1, branches 85.7 to 86.3, functions 97.4 to 97.5. Python scripts (CI 93.85, headroom 0.95 pp) and bench floors unchanged per the 1.0 pp band.

## Recall/noise deltas for the wave-end paired mini

- Citation fields on delivered findings can now be stripped (finding kept, blockquote dropped); the url-set false-fire rate on URL-citing rule text is the watched axis.
- A citation-field pattern match now also drops the finding's suggestion fence.

Adjacent issues filed during this pass: #247 (exclusions kill-switch over citation text), #248 (malformed replayed checkpoint yields silent ok:true empty review).

https://claude.ai/code/session_01YbTRxvFWjQkZnEvSykrySi

---

> [!NOTE]
> **Medium Risk**
> Changes what text reaches posted PR comments and one-click fixes, with nontrivial checkpoint-replay paths, but behavior is twin-parity tested and idempotent on the belt.
> 
> **Overview**
> Extends the **#62 injection strip** from `suggestion` alone to **`claude_md_rule` and `spec_text`**, repo-derived citation fields that render into PR comments. Python and JS filter twins share `_INJECTION_STRIPPED_PROSE_FIELDS` and run the same seven pattern sets per field: **strip the field, keep the finding**, with `{field}_removed_by` / `{field}_removal_reason` stamps. A pattern hit on any scanned field still **drops `suggested_fix_code`** (first field in scan order names the propagation reason).
> 
> Adds a **replay belt** in `stages.js`: challenge checkpoints replayed without re-running filter get **`applyInjectedProseStrip`** on `findings`, `unverified`, and `eliminated` so delivery, report prompts, `findings.json`, and derived post-review projection cannot leak unstripped citations. Stats add **`claude_md_rules_removed`** / **`spec_texts_removed`**; parity tests and goldens cover the new behavior. **JS coverage floors** and AGENTS.md docs are ratcheted to match CI; **`apply_exclusions` is unchanged** for citation fields (documented rationale).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0c1bb10b8b99db19a47a308524703d5357506ff. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:
Both Bugbot findings addressed in `ed549a9`:

- **Belt throws on malformed nested arrays** (low) — fixed at the belt call site: a `beltStrip` helper mirrors `stripReportExcludedFields`'s skip predicate at both levels, so a non-array field and a non-object element pass through unchanged instead of throwing. The twin-paired strip functions are untouched (no parity impact). Two new tests pin the tolerance; mutation-verified red against the bare `.map`. The probe also exposed a pre-existing, unrelated crash — a null element in a replayed findings list dies later in `rankFindings` on main too — filed as #250 alongside #248's checkpoint-shape family.
- **Budget pin undercounts AGENTS.md growth** (medium) — refuted by measurement: the AGENTS.md set totals exactly 21061 bytes, equal to `AGENTS_SET_BUDGET_BYTES`, and `test_canonical_set_fits_under_the_codex_cap` is deliberately pinned AT the measured size (`assertLessEqual`; its own comment states the current tree passes exactly and one added byte does not). The suite passes locally and in CI at both commits. No change.

tests/test_filter_findings.py
```diff
@@ -2853,5 +3267,136 @@ def test_fix_max_lines_and_chars_agree_across_all_three_homes(self):
...
+        # file) -- an unbounded slice-to-EOF would also (wrongly) pass if the splice
+        # landed in a later function instead of this one.
+        js_fn_next = js_src.find("\nexport function ", js_fn_start + 1)
+        js_fn_body = js_src[js_fn_start : js_fn_next if js_fn_next != -1 else None]
+        self.assertIn(
+            "...proseFieldsRemoved",
+            js_fn_body,
+            "workflows/src/filterFindings.js's applyFilterPipeline no longer "
+            "splices proseFieldsRemoved into its stats object",
+        )
```
leehopper: **💡 [LOW] New lockstep test guards a source-code substring instead of behavior**

test_both_twins_splice_the_derived_stats_map_into_the_receipt asserts that the literal substrings "**prose_fields_removed" (Python) and "...proseFieldsRemoved" (JS) appear in the relevant function bodies. This is a phrase/string-presence guard over trusted source code, and REVIEW.md's rule against phrase-count guards applies literally: any equivalent, correct refactor of the splice (e.g. stats.update(prose_fields_removed) or a differently named local) would fail this test even though the actual emitted-stats behavior (already covered by the sibling test_both_twins_emit_a_removed_stat_for_every_scanned_field, which drives the real pipeline) would be unaffected.

**Suggested fix:**
Rely on the existing behavioral test (test_both_twins_emit_a_removed_stat_for_every_scanned_field / the JS analog in filter_unit.test.js) as the structural proof that the mechanism generalizes, and drop or clearly demote this substring check to a non-blocking style note rather than a hard assertion, since a correct rewording of the splice construct would otherwise be treated as a regression.

**Cited rule:**
> REVIEW.md Rules: "CRITICAL: Never accept a phrase count as a guard. A check that greps for wording is defeated by rewording. Flag any guard that asserts on the presence of a string where a structural property was available."

leehopper (commented):
---

1 inline comment(s) were posted; the following 1 finding(s) reference lines outside this diff and are included here instead: A finding listed here may not have an anchoring problem of its own — a consolidation group whose primary could not be anchored inline is listed here in full, corroborators included.

#### `scripts/filter_findings.py:1552`

**💡 [LOW] No regression test pins that apply_exclusions deliberately skips claude_md_rule/spec_text**

This PR adds an explicit design decision (documented in a new comment) that apply_exclusions must NOT scan claude_md_rule/spec_text even though it scans suggestion for the same 'rendered into comments' reason, because doing so could mass-eliminate the conventions dimension on common phrasing like 'MUST'. Neither the Python nor JS TestApplyExclusions/applyExclusions test suites include a case proving a claude_md_rule- or spec_text-matching exclusion pattern does NOT eliminate the finding, so this documented contract has no regression guard in either runtime.

**Suggested fix:**
Add a test in both TestApplyExclusions (Python) and the JS applyExclusions suite that sets claude_md_rule (or spec_text) to a string matching a user exclusion pattern, and asserts the finding is NOT eliminated (passed length 1, eliminated length 0) -- pinning the current 'exclusions do not scan citation fields' contract in both twins.

---
Generated by code-gauntlet | Reviewed up to: ed549a90d36c63d70392aa414e4adb2138d30228

tests/test_filter_findings.py
leehopper: Dropped the substring test in 1a717b2; the behavioral emission test (test_both_twins_emit_a_removed_stat_for_every_scanned_field + the JS analog) is the structural proof, and it loops the shared field list so a future field is covered without a new test.

Issue #215, opened 2026-08-18
six of the eight injection pattern sets never scan the finding title


leehopper:
## Problem

In both twins, only the shell set scans `title\ndescription` and the placeholder set scans
`title`; the url/encoded/bypass/instructional/vuln-intro/body-marker sets scan `description`
alone. A payload placed only in `title` is invisible to six of the eight regex heuristics — and
per `post_review.py`'s own docstring, every finding's title reaches the wire RAW (only
`suggestion`/`claude_md_rule`/`spec_text` go through `_sanitize_outbound_prose`).

Surfaced by the #62 research pass (#210). Changing which findings are eliminated →
recall/noise-sensitive, rides a paired measurement.

leehopper:
## Executed 2026-08-26 — PR #251

**Decision of record.** The title scan ships at PATTERN granularity, not set
granularity: all six content sets (url/encoded/bypass/instructional/vuln-intro/
body-marker) scan `title`, minus a four-entry exclusion list
(`_TITLE_SCAN_EXCLUDED_PATTERNS` / `TITLE_SCAN_EXCLUDED_PATTERNS`, byte-copied from
the home lists, parent lists untouched so description scans stay byte-identical). A
title match eliminates (the shell/placeholder title precedent; a title cannot be
stripped — undeliverable, and the dedup signature keys on it).

**Measurement decided the policy (the #213 recipe).** 1,895-title corpus (1,745 real
from bench runs + all 116 repo PRs' review comments; 150 synthetic hard negatives),
54 blind title lures, patterns imported from the shipped module compiled
`re.IGNORECASE | re.ASCII`. Every false-fire localized to single patterns:
- encoded base64-run `\b[A-Za-z0-9+/]{40,}={0,2}\b`: 8/1,745 real titles (0.46%),
  all slash-joined identifier pairs — EXCLUDED.
- bypass `\bauto[-\s]?approve\b`: 2 fires, both synthetic hard negatives (0/1,745
  real) — EXCLUDED as a stated precaution on that weaker basis.
- url `\bnavigate to\b`: zero corpus exercise + generic verb phrase — EXCLUDED
  pending evidence.
- url long-bare-URL `https?://…{20,}`: review-executed counterexamples (realistic
  SSRF/open-redirect/origin-check titles quoting the offending URL) + 2 real
  bench-golden security comments fire it — EXCLUDED (adopted from the round-1
[...]

PR #255, opened 2026-08-27
fix(filter): directive-gate five injection patterns, empty the title-scan exclusion list (#252)

### LINKED ISSUE #252: four injection patterns false-fire on (or cannot exclude) legitimate text; fixing their shape unblocks the #215 title carve-outs and removes live description-side elimination noise
The #215 pass measured every pattern of the six content sets against 1,895 legitimate
finding titles (1,745 real from bench runs + repo PR reviews, 150 synthetic security
hard negatives) and against real bench finding descriptions. Four patterns false-fire
on legitimate text or cannot rule out legitimate title uses; #215 excluded exactly
these four from the new title scan via `_TITLE_SCAN_EXCLUDED_PATTERNS` /
`TITLE_SCAN_EXCLUDED_PATTERNS`:

1. encoded `\b[A-Za-z0-9+/]{40,}={0,2}\b` — fires on slash-joined identifier pairs
   (`getAuthorizationConditions/getFilterConditions`), 8/1,745 real titles (0.46%).
   The same pattern eliminates findings on the DESCRIPTION side today: 72/3,061 real
   bench findings (2.35%) match on description and are eliminated by the shipped
   filter, across three shapes — slash-joined identifier runs (65), single ≥40-char
   identifiers with no slash (`SubjectImpersonationTokenExchangeProvider`, 41 chars),
   and long dot-free path runs. This description-side leak is larger than the
   title-side gap #215 closed and is live recall loss now.
2. encoded hex `(?<!\w)(?:0x)?[0-9a-fA-F]{32,}(?!\w)` — clean on titles (0/1,895,
   title-enabled by #215), but fires on 40-hex git SHAs in descriptions that quote a
   commit (`da09bc0808e7c6df1bc10e00d5acf764d94c7cc7`) — part of the 2.35% above.
3. bypass `\bauto[-\s]?approve\b` — bare noun phrase; auto-approve is a feature name
   under review, so plausible CI-security titles ("Auto-approve enabled for CI
   workflow runs triggered by fork pull requests") fire it. Observed fires were
   synthetic hard negatives (0/1,745 real), carved from the title scan as a
   precaution; the description-side exposure is the same class.
4. url long-bare-URL `https?://…{20,}` — realistic SSRF/open-redirect/origin-check
   titles quote the offending URL ("SSRF: host validation allows
   http://169.254.169.254/latest/meta-data/ fetches") and would be eliminated; two
   real bench-golden security comments fire it. Needs a directive-context or
   reputation-shaped narrowing before title enablement; note it is also the primary
   title exfil channel, so the carve leaves bare exfil URLs in titles deliverable.
   Also carved on corpus silence rather than a measured fire: url `\bnavigate to\b`
   (zero corpus exercise; realistic frontend titles like "Router fails to navigate to
   the error boundary on a 500 response" would fire).

Fix shape: narrow the pattern shapes (e.g. the base64-run pattern needs a way to
reject identifier-morphology runs — slash-density/dual-case-run heuristics; the hex
pattern needs to reject exact-40-hex commit-SHA morphology in prose context; the
auto-approve and navigate-to phrases need imperative context, not bare noun/verb
phrases). Each fix that lands removes its entry from the two exclusion lists — one
edit per twin, guarded by the #215 membership lockstep and carve-out goldens (which go
stale on the re-record, forcing the deliberate re-decision).

Recall/noise-sensitive on both sides (description false-fire removal RAISES recall;
title widening changes elimination) — rides a paired measurement.

Surfaced by the #215 pass (PR #251).

leehopper:
Closes #252.

## Problem

Four injection patterns false-fire on legitimate finding text, and one (hex) eliminates real findings on the description side today. Measured on the local corpus:

- **base64-run** `\b[A-Za-z0-9+/]{40,}={0,2}\b` fires on slash-joined identifier pairs (`getAuthorizationConditions/getFilterConditions`), long file paths, single ≥40-char identifiers, and this repo's own GitHub blob-permalink citations (`.../blob/<40hex>/<path>`). #252 reports 2.35% of real bench findings (72/3,061) wrongly eliminated on description.
- **hex** `(?<!\w)(?:0x)?[0-9a-fA-F]{32,}(?!\w)` fires on 40-char commit SHAs quoted in prose — every real fire was a SHA inside a permalink citation.
- **auto-approve** `\bauto[-\s]?approve\b` fires on the feature-name noun ("Add an auto-approve toggle for trusted branches").
- **long-bare-URL** `https?://…{20,}` fires on SSRF/redirect findings that legitimately quote the offending URL.

These four were carved out of the #215 title scan via `TITLE_SCAN_EXCLUDED_PATTERNS`; hex was left title-enabled but still leaks on the description side.

## Decision — directive/grammatical context gating (measured before chosen)

The discriminator is the same for all five: real false-fires are descriptive or citation text; a true injection payload carries an imperative. The file already proves this design — `\bvisit\s+https://` and `\bdownload from\s+https://` ship enabled because an imperative is present, while the bare-URL and `navigate to` entries were the two missing it. So every narrowing requires a directive:

| pattern | narrowed shape |
|---|---|
| base64-run | decode-family verb (`decode\|base64\|atob\|b64decode`) before the blob, **or** blob then shell-sink (`\| sh`, `base64 -d`, `then/and run`) |
| hex | `decode\|unhex\|xxd\|fromhex\|unhexlify` before, **or** hex then sink |
| auto-approve | `auto-approve` + mandatory determiner + `PR\|MR\|change\|commit`, **or** `auto-approve and (merge\|skip\|…)` |
| long-bare-URL | reader-imperative (`browse/go/navigate to`, `fetch/retrieve/pull from`) + URL, **or** exfil-verb + secret-object + URL |
| navigate-to | folded into long-bare-URL (requires a following URL) |

A bare blob, SHA, quoted URL, or feature-name noun no longer matches. **Measured 0 false-fire** across 1,257 bench descriptions, 1,312 titles, 493 real PR-review comments, and an adversarial negative set (git-permalink repro sentences, camelCase identifiers, prose SHAs/MD5s, SSRF-quoted URLs, "auto-approve toggle", routing "navigate to the error boundary"). Directive-lure catch: base64 7/12, hex 7/10, auto-approve 8/11, url 15/18 effective; the sink after-branch remains the unweakened primary catch for the real fetch-then-execute shape.

## Exclusion list → empty, machinery removed

With every pattern clean, `TITLE_SCAN_EXCLUDED_PATTERNS` / `_TITLE_SCAN_EXCLUDED_PATTERNS`, the `_title_suggestion_sets` filtering, and the three exclusion-guard tests are removed; all content sets title-scan directly. The four `*_kept_carveout` goldens and `encoded_title_hex_eliminated` are re-decided against the narrowed patterns (the base64/hex/URL/navigate/auto-approve legit titles now stay KEPT on the patterns' own merit; the commit-SHA title flips KEPT).

## Combined-scan (adversarial-review finding, fixed in-PR)

Narrowing to directive **adjacency** opened a cross-field split: a decode directive in the title + a bare blob in the description evaded per-field scanning. The url and encoded sets now scan the **combined** title+description string (as the shell set already did) and are removed from the separate title pass, closing the split. A far-gap split (directive padded >40 chars from the blob) and the per-field-only advisory sets (bypass/instructional/vuln/body) remain documented residuals — filed as a follow-up. A cross-field-split regression test guards the fix.

## Threat model

Finding title/description reach the posted PR comment (a human), the blind challenger LLM (tools: Read/Grep/Glob/LSP), and the report-writer LLM. Nothing auto-decodes base64/hex or auto-fetches a URL, so a directive-free blob/URL is inert (quoted, not executed); the gating keeps exactly the imperative forms that can manipulate a human or LLM. The narrowing also applies to the #62/#213 prose-field strip path, consistent with the accepted "bare URLs deliverable" residual from #215.

## Validation

- Twins byte-identical (unicode-guard green); bundle regenerated (`git diff --exit-code workflows/pipeline.js` clean).
- Python 1,789 + 741 subtests; bench 553 + 31; JS 989 — all green; parity `--check` fresh.
- Six isolating regression tests for the new directive branches, each red against its own deletion; mutation ledger (11 entries) red-first per narrowing with byte-verified restores.
- Three adversarial review rounds (design red-team, implementation review, final sweep); all findings adopted or refuted on executed evidence.
- Coverage above floors: Python 93.94/92.9, bench 88.52/87.5, JS 99.02/87.60/98.50 vs 98.1/86.5/97.6. Floors ratcheted below per the PR CI measurement.

## Recall/noise delta for the wave-end paired mini

Description-side false eliminations removed (RAISES recall); title scanning of the five patterns now enabled with directive gating (0/1,895-class measured false-fire). Residuals reaching the wire: far-gap cross-field splits, bare directive-free payloads (inert), and the per-field advisory sets' cross-field splits.

---

> [!NOTE]
> **Medium Risk**
> Alters security-review output by changing which findings are eliminated or field-stripped; recall improves for legit findings but adjacency-gated patterns and description-only advisory scans leave documented injection bypass residuals.
> 
> **Overview**
> Tightens **prompt-injection filtering** in `apply_injection_filter` (Python + mirrored JS pipeline) so legitimate security findings are less often dropped, while still catching imperative payloads.
> 
> **Pattern changes:** Encoded base64/hex matches now require a **decode-family verb or execute sink** near the blob (not bare SHAs, tokens, or permalink citations). **Auto-approve** matches only imperative shapes (determiner + PR/commit object, or `and merge/skip/…`), not feature-name prose. **URL** detection keeps only **`visit` / `download from` + URL** — bare long URLs, `navigate to`, and the #252 directive-gated bare-URL branches are **removed** (#255) because they matched real routing, SSRF, and exfil descriptions.
> 
> **Scanning model:** Shell, URL, and encoded heuristics scan **`title` + `description` together** so directives split across fields still eliminate; the separate title-only pass is reduced to four content sets (bypass/instructional/vuln/body). **`TITLE_SCAN_EXCLUDED_PATTERNS` and its guards are deleted** — carve-outs are replaced by narrower patterns.
> 
> **Tests/docs/CI:** Parity fixtures, unit tests (cross-field splits, isolating branches, legit-URL negatives), `AGENTS.md` / CI JS branch coverage floor **86.5 → 86.6**, and committed `workflows/pipeline.js` bundle updated to match `workflows/src`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf3885e096b59af61565c85d740b1686a1ab2702. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/filter_findings.py
```diff
@@ -879,10 +868,9 @@ def _strip_suggested_fix_code_if_needed(finding, first_pattern_strip):
             reasons.append(f"description matches injection marker: {m!r}")
 
         # Title scan: each of the six sets above minus shell (already
-        # scanned at heuristic 1) is re-scanned against `title` alone, minus
-        # _TITLE_SCAN_EXCLUDED_PATTERNS -- a payload placed only in the title
-        # reaches the wire and downstream model prompts untouched by the
-        # description-only scans above.
+        # scanned at heuristic 1) is re-scanned against `title` alone -- a
+        # payload placed only in the title reaches the wire and downstream
+        # model prompts untouched by the description-only scans above.
```
leehopper: **💡 [LOW] Stale "six sets" title-scan comments not updated after narrowing to four sets**

This PR moved url/encoded off the separate title-only pass (now scanning `combined` at heuristics 2a/2b instead), shrinking the title-only pass from six content sets to four (`_SUGGESTION_SETS[3:]`). The docstring at lines 636-641 was correctly updated to say "four content sets", but three other comments describing the same mechanism were left saying "six": the stats-field comment at line 41, the inline comment right above the title-scan loop at line 870, and the comment above `title_matches_eliminated` at line 1794. A reader of any of these three would believe six sets participate in the title-only pass, contradicting the (correct) docstring three lines above.

**Suggested fix:**
Update all three stale "six-set"/"six sets" comments (lines 41, 870-873, 1794 in scripts/filter_findings.py, and the matching pair at lines 601-604 and 1192 in workflows/src/filterFindings.js) to say "four-set"/"four sets", consistent with the already-corrected docstring at lines 636-641.

---

**Corroborating finding — code-simplifier (simplification, confidence 95):**

**Title-scan inline comment says 'six sets' but the code now scans only four**

The docstring above this loop (lines 636-641, updated in this diff) correctly says the title-only pass now covers the same four content sets (bypass/instructional/vuln-intro/body-marker) since url/encoded moved to scan combined. But the inline comment directly above the loop that iterates _title_suggestion_sets still says 'each of the six sets above minus shell', true before this diff (7 sets minus shell = 6) but wrong now since _title_suggestion_sets is _SUGGESTION_SETS[3:], only 4 sets. A maintainer reading the comment closest to the code would misjudge which sets participate.

workflows/src/filterFindings.js
```diff
@@ -585,10 +599,9 @@ export function applyInjectionFilter(findings) {
     if (m) reasons.push(`description matches injection marker: ${JSON.stringify(m)}`);
 
     // Title scan: each of the six sets above minus shell (already scanned at
-    // heuristic 1) is re-scanned against `title` alone, minus
-    // TITLE_SCAN_EXCLUDED_PATTERNS -- a payload placed only in the title
-    // reaches the wire and downstream model prompts untouched by the
-    // description-only scans above.
+    // heuristic 1) is re-scanned against `title` alone -- a payload placed
+    // only in the title reaches the wire and downstream model prompts
+    // untouched by the description-only scans above.
```
leehopper: **💡 [LOW] Title-scan inline comment says 'six sets' but the code now scans only four (JS twin)**

Same staleness as the Python twin: the doc comment above TITLE_SUGGESTION_SETS (lines 388-394, correctly updated in this diff) says the title-only pass now covers only the four still-description-only sets (bypass/instructional/vuln-intro/body-marker), since shell/url/encoded now scan combined. But the inline comment directly above the applyInjectionFilter loop that iterates TITLE_SUGGESTION_SETS still reads 'each of the six sets above minus shell', which is stale now that TITLE_SUGGESTION_SETS is SUGGESTION_SETS.slice(3), only 4 sets, not 6.

**Suggested fix:**
Before: '// Title scan: each of the six sets above minus shell (already scanned at heuristic 1) is re-scanned against title alone'. After: '// Title scan: the four sets that are still description-only (bypass/instructional/vuln-intro/body-marker) are re-scanned against title alone -- shell/url/encoded already scan combined (which includes title), so re-scanning them here would double-report the same match.'

```suggestion
    // Title scan: the four sets that are still description-only
    // (bypass/instructional/vuln-intro/body-marker) are re-scanned against
    // `title` alone -- shell/url/encoded already scan `combined` (which
    // includes `title`), so re-scanning them here would double-report the
    // same match. A payload placed only in the title reaches the wire and
    // downstream model prompts untouched by the description-only scans above.
```

scripts/filter_findings.py
```diff
@@ -831,15 +817,18 @@ def _strip_suggested_fix_code_if_needed(finding, first_pattern_strip):
         if m:
             reasons.append(f"contains shell command pattern: {m!r}")
 
-        # 2a. URLs to visit in description
-        m = _first_match(url_re, description)
+        # 2a. URLs to visit -- combined title+description (#252 Finding 1:
+        # a decode/exfil directive in title and the URL/blob in description
+        # must be caught together, since the rendered comment concatenates
+        # them into one instruction).
+        m = _first_match(url_re, combined)
```
leehopper, line 827: **💡 [LOW] Cross-field split test coverage exists only for the encoded heuristic, not the parallel URL heuristic**

Heuristics 2a (URL) and 2b (encoded) were changed identically from scanning `description` alone to scanning `combined` (title+description) so a directive/blob split across fields is still caught, but only the encoded case has a cross-field-split test (`test_cross_field_split_encoded_directive_and_blob_eliminates` and its JS mirror); no test exercises a URL directive in title with the URL itself in description (or vice versa).

scripts/filter_findings.py
```diff
@@ -639,15 +616,29 @@ def apply_injection_filter(findings):
...
+    `description` -- would satisfy neither field scanned alone even though
+    the rendered PR comment concatenates them into one coherent
+    decode-then-run instruction. Scanning combined closes that split without
+    reopening the bare-blob/bare-URL false-fire the directive-gating exists
+    to fix (a decode verb sitting within the adjacency window of an
+    unrelated identifier at a title/description boundary was measured and
+    not observed in the real corpus -- see `SCRATCHPAD/final_measure.py`
+    Part 3). A far-apart split (directive and blob more than the adjacency
+    window apart) still evades by design -- adjacency-gating is inherently
+    local, and that residual is accepted.
```
leehopper: **💡 [LOW] "roughly 40 characters" adjacency claim overstates precision for the URL exfiltration pattern**

The new docstring generalizes the encoded/url directive-gating as an adjacency requirement of "roughly 40 characters" between verb and blob/URL. That is exact for the two encoded-payload patterns (`[^\x00]{0,40}`), but the URL exfiltration pattern chains two `[^\x00]{0,30}` gaps (verb→secret-noun, secret-noun→URL), allowing up to 60 characters total, while the reader-imperative URL pattern allows only whitespace (effectively 0) between verb and URL. "Roughly 40" is not an accurate summary for either of those two url sub-patterns.

**Suggested fix:**
Either state the actual per-pattern windows (40 for encoded, 0/60 for the two url sub-patterns) or soften the claim to avoid implying a single shared numeric bound across both content sets.

scripts/filter_findings.py
```diff
@@ -448,24 +448,47 @@ def apply_threshold_filter(findings, config):
...
     r"\bdownload from[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]+https?://",
-    r"\bnavigate to\b",
+    r"\b(?:browse[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]+to|go[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]+to|open|navigate[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]+to|fetch[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]+from|retrieve[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]+from|pull[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]+from)[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]+https?://",
+    r"\b(?:send|post|upload|exfiltrate|leak|forward|transmit|beacon|report)\b[^\x00]{0,30}(?:tokens?|secrets?|cookies?|credentials?|api[-\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]?keys?|passwords?|sessions?|env)\b[^\x00]{0,30}https?://",
```
leehopper: **💡 [LOW] No negative test proves the new exfiltration-URL pattern spares a legitimate data-exfiltration vulnerability finding**

The new URL pattern eliminates any finding whose combined text has an exfil verb (send/post/exfiltrate/leak/...) near a secret-object noun (token/secret/cookie/session/env/...) followed by a URL; every existing test/fixture for this branch (isolating_url_exfil_eliminated and its unit-test twin) treats that exact shape as correctly-eliminated injection, but none demonstrates that an analyst's accurate description of a real exfiltration vulnerability in the reviewed code — which necessarily uses the same vocabulary — is kept rather than silently dropped.

AGENTS.md
```diff
@@ -76,7 +76,7 @@ LCOV="$(mktemp -d)/js-coverage.lcov" && node --test --experimental-test-coverage
 Floors: Python 92.9 / 87.5 (scripts raised 2026-08-19 from the #219 PR CI measurement:
 93.37, then 2026-08-24 to 92.6 (#231: 93.54), to 92.7 (#236: 93.67), and 2026-08-25 to 92.9
 (#238: 93.84); bench raised 2026-08-19 from the #219 run:
-88.29); JS 98.1 / 86.5 / 97.6 (lines pinned
+88.29); JS 98.1 / 86.6 / 97.6 (lines pinned
```
leehopper, line 79: **💡 [LOW] JS branches floor bumped to 86.6 without the measurement citation the file's own convention requires**

The floor-history paragraph documents every prior branches-floor change with an explicit "(from the #NNN PR CI measurement: X)" citation (e.g. #62, #249, #251). This diff bumps the summary number from 86.5 to 86.6 but the narrative clause at lines 80-83 still only documents history up through the #251 raise to 86.5, with no added citation for the new 86.6 value, breaking the file's own established citation pattern that REVIEW.md requires ("Claims must name their measurement").

**Suggested fix:**
Append a clause documenting the PR and CI measurement backing the 86.5→86.6 raise, matching the format used for every prior floor change in the same paragraph.

**Cited rule:**
> REVIEW.md Rules: "Claims must name their measurement. Flag a comment or doc that asserts a behaviour, frequency, or limit without saying how it was established."

scripts/filter_findings.py
```diff
@@ -639,15 +616,29 @@ def apply_injection_filter(findings):
...
+    exfiltration phrase ahead of a URL) with an adjacency requirement (the
+    verb and the blob/URL within roughly 40 characters of each other), so a
+    payload split across fields -- the directive in `title`, the blob in
+    `description` -- would satisfy neither field scanned alone even though
+    the rendered PR comment concatenates them into one coherent
+    decode-then-run instruction. Scanning combined closes that split without
+    reopening the bare-blob/bare-URL false-fire the directive-gating exists
+    to fix (a decode verb sitting within the adjacency window of an
+    unrelated identifier at a title/description boundary was measured and
+    not observed in the real corpus -- see `SCRATCHPAD/final_measure.py`
```
leehopper: **💡 [LOW] New docstring cites a session-scratch file that never lands in the repo tree**

The new docstring text supporting the combined-scan change cites `SCRATCHPAD/final_measure.py` Part 3 as the evidence that a decode verb sitting within the adjacency window of an unrelated identifier was "measured and not observed in the real corpus". A grep of the whole repository finds no `SCRATCHPAD` path anywhere, and AGENTS.md's own rule states session scratch (design memos, implementation plans, handoff notes) "never lands in the tree". The citation is therefore permanently unverifiable to anyone reading the shipped comment.

**Suggested fix:**
Either drop the specific file citation (keep the prose claim without an unverifiable pointer) or move the measurement summary into the PR/issue thread as AGENTS.md's session-scratch rule directs, and reference the issue/PR number instead of a path that will never exist in the tree.

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: 3a1ad55fdd700099ec57205d5e14c2760ac30133

scripts/filter_findings.py
leehopper: Fixed in bf3885e — and this was a real recall issue, not just a missing test. I verified the exfil branch (and the reader-imperative branch) eliminate realistic legitimate findings that quote the same vocabulary: "an attacker can send the session cookie to https://evil…", "could exfiltrate the auth token to https://…", "Logs leak the bearer token to https://…", "the router should navigate to https://…", "docs tell developers to open https://…". A regex can't separate a legit finding that quotes an exfil/nav directive from an injection that issues one. Both new long-bare-URL branches are removed; url is now just `visit https://` / `download from https://`. Added `test_url_keeps_legit_nav_and_exfil_findings` asserting all five KEPT.

scripts/filter_findings.py
leehopper: Fixed in bf3885e. url still scans `combined`, so a directive/URL split is caught — added `test_cross_field_split_url_visit_directive_and_url_eliminates` (title="Reviewers should visit", description=bare URL). The exfil/reader-imperative branches are removed (see the exfil-finding reply).

scripts/filter_findings.py
leehopper: Fixed in bf3885e — the three stale "six sets" comments (~41, ~871, ~1796) now say "four sets".

workflows/src/filterFindings.js
leehopper: Fixed in bf3885e — JS twin inline comment updated to "four sets" with the double-report rationale, per your suggestion.

scripts/filter_findings.py
leehopper: Fixed in bf3885e. With both long-bare-URL branches removed, the docstring states encoded's ~40-char window and url's whitespace-adjacent visit/download-from form separately, no shared "roughly 40" claim.

scripts/filter_findings.py
leehopper: Fixed in bf3885e — removed the `SCRATCHPAD/final_measure.py` citation; provenance lives in the PR, not the shipped comment.

AGENTS.md
leehopper: Fixed in bf3885e — added the citation ("branches 2026-08-27 to 86.6 from the #255 PR CI measurement: 87.60") and raised `AGENTS_SET_BUDGET_BYTES` 21161→21234 to fit, with justification, per the layout guard's "raise it in the same commit" guidance. Command block stays byte-identical to ci.yml.

PR #257, opened 2026-08-28
feat(filter,pipeline): combined-scan unification, pattern-shape fixes, replay filtering belt (#256, #254, #253)

### LINKED ISSUE #253: replayed pre-#215 challenge checkpoints bypass the title scan; decide a filtering-belt analogue
`PERSISTED_RESUME_PHASES = ['challenge']`: a resume replays `challengeOut` verbatim
into delivery selection, report input, and re-persisted checkpoints. The #213 belt
(`applyInjectedProseStrip` in stages.js) closes this for the strip-and-keep prose
fields, but it cannot help the #215 title scan: a title match ELIMINATES, and the belt
is a per-finding field strip, not a list partition. A challenge checkpoint recorded by
a pre-#215 pipeline and replayed post-upgrade therefore delivers a title-borne
url/encoded/bypass/instructional/vuln-intro/body-marker payload to the wire, the
challenger prompt, and the report agent.

#215 scoped this out deliberately: a replay-time elimination is not loss-free (it
changes counts, delivery selection, and interacts with the all-degraded/empty-report
guards), and the vulnerable population — checkpoints recorded before the #215 version
— shrinks over time; fresh runs are clean by construction.

Decision to make here: extend the belt to a filtering partition (move title-matching
replayed findings from `.findings`/`.unverified` into `.eliminated` with the same
reason strings, with the same not-counted-in-stats disclosure the #213 belt already
carries), or ratify the scope-out permanently with the residual documented. Suites-only
if implemented (replay fixtures exist from #213).

Surfaced by the #215 pass (PR #251).

### LINKED ISSUE #254: injection pattern bigram brittleness — measured 66.7–100% per-set lure catch, misses are phrase-shape gaps
The #215 measurement ran 54 authored title lures against the shipped sets. Misses are
concentrated in literal-bigram brittleness that affects title AND description scanning
equally (the scan surface is not the problem; the pattern shapes are too tight):

- vuln_intro: "disabling TLS verification" — patterns require the bare stem
  (`disable\b`), inflected forms escape.
- body_marker: `<finding severity="low">` escapes the literal `<finding>`;
  `[INSERT FINDING TITLE HERE]` escapes the whitespace-only `[INSERT]` class.
- instructional: "requires you to run", "copy and paste this command into your
  terminal" escape the exact bigrams (`you should run`, `paste (this|the following)
  into`).
- url: `ftp://` escapes `https?://`-anchored patterns.
- bypass: "bypass security review" escapes `bypass (security )?controls?`.

Distinct from the false-fire issue (patterns too loose): this is the too-tight axis.
Any widening is recall/noise-sensitive and rides a paired measurement; candidates
should be corpus-measured with the #215 methodology (real-title corpus + blind lures)
before shipping.

Surfaced by the #215 pass (PR #251).

### LINKED ISSUE #256: Unify all injection content sets to combined-scan; retire the per-field title pass
## Context

Surfaced by the #252 adversarial review (PR #255).

The prompt-injection filter scans a finding's title and description with the content sets. Historically the shell set scans the **combined** `title\ndescription` string, while the other five sets scanned `description` alone plus a separate title pass (#215). #252 moved the **url and encoded** sets to combined-scan as well, because directive-adjacency gating opened a cross-field split (a decode directive in the title + a bare blob in the description evaded per-field scanning).

The **bypass, instructional, vuln-intro, and body-marker** sets still scan per-field (description + title pass). A multi-token pattern in any of them can likewise be split across the title/description boundary and evade — e.g. `you should` (title) + `run this command` (description), or `auto-approve this` (title) + `pull request` (description). The #252 review classified these as an accepted residual because the split yields only human/LLM-**advisory** framing carrying no decode-run/fetch payload of its own, unlike the machine-actionable shell/url/encoded sets.

## Proposal

Move all six content sets to a single combined-scan (retire the separate `_title_suggestion_sets` title pass entirely), so no multi-token pattern can be split across the field boundary, and the scan model is uniform with shell. This removes:

- the per-field title pass and the `title_scan_matched` / `title_matches_eliminated` observability stat (superseded — every match is a combined match),
- the description-vs-title reason-prefix distinction (all reasons become the bare shell-style form).

It requires re-recording the description/title parity goldens for the four moved sets and a combined-corpus false-fire re-measurement (the narrowed #252 patterns already measured 0 on combined; the four advisory sets are contiguous phrases with their own measured behavior).

## Notes

- Recall/noise-sensitive if it changes any elimination; rides the wave-end paired mini.
- Also consider whether the far-gap (>40 char) cross-field split that remains for the adjacency-gated sets is worth addressing, or is an inherent, accepted limitation of adjacency heuristics (documented in #252).

Suites-only.

### LINKED ISSUE #260: heuristic-7 placeholder-title patterns false-fire on legitimate findings — previously unmeasured surface
Heuristic 7's placeholder-title patterns (`_INJECTION_TITLE_PATTERNS`: TODO/FIXME/Placeholder/Example finding/...) false-fire on legitimate findings, and this surface has never been measured — the #215/#252/#254 measurement lineage covered only the seven content sets.

Baseline (2026-08-28 pass, 727-pair real corpus): 5/727 titles fire, 8/727 on the combined surface. All classified legitimate — findings about literal TODO comments or placeholder-string fallbacks in reviewed code (e.g. a real finding titled around "refresh_token fallback assigns the literal placeholder string ... instead of a real value"). Those findings are eliminated today.

That is a ~0.7% title false-fire rate on real findings, higher than anything the content sets measured. Candidates: context-gate the generic words (a title ABOUT a TODO comment vs a placeholder title), narrow to title-initial or whole-title matches, or ratify with the rate documented. Note the #254 pass already respelled this set's literal-space separators (`Example finding` etc.), which is orthogonal.

Recall/noise-sensitive; rides a paired measurement.

Surfaced by the corpus baseline of the #256/#254/#253 pass (PR #257).

### LINKED ISSUE #261: false-positive-exclusions.md describes description-only scanning — stale since #252, wrong after #256
`skills/code-gauntlet/references/false-positive-exclusions.md` (Prompt Injection Artifacts section, ~line 176) still says "Finding **description** contains shell commands…". That field framing has been stale since #252 moved shell/url/encoded to combined title+description scanning, and is wrong for every content set after #256 (PR #257): all seven sets now scan the combined string.

This file is the category source of truth the filter's docstring cites (`scripts/filter_findings.py` references it), and it is mirrored into the agents/*.md copies under the keep-all-agent-copies-in-sync marker, so the stale wording is duplicated across the agent contracts.

Nothing tests this prose against the code (the #256 pass grep-verified no test pins it), so it cannot break CI — it just misinforms agents doing self-filtering about which fields matter. Fix is a prose pass over the reference file plus the mirrored agent copies, per the duplication discipline in agents/AGENTS.md. Suites-only, doc-only, no recall sensitivity.

Surfaced by the #256 pass review (PR #257).

leehopper:
Closes #256. Closes #253. Closes #254. Closes #260. Closes #261.

Three Wave 5 items from the #215/#252 adjacent set, landed as one round because each changes the same scan surface: #256 makes the scan model uniform, #254 fixes measured pattern-shape gaps on that uniform surface, and #253 makes replayed checkpoints re-enter it.

## #256 — all content sets scan combined; the title pass is retired

The four remaining description-only sets (bypass, instructional, vuln_intro, body_marker) now scan the combined `title\ndescription` string, matching shell/url/encoded. The #215 per-field title pass, the `title_scan_matched` stamp (and its input-pop guard), and the `title_matches_eliminated` stat are deleted from both twins. Reason strings for the four moved sets drop the `description ` prefix and take the bare shell-style form.

Cross-field splits (`you should` in the title, `run this command` in the description) can no longer evade a multi-token pattern. Two structural guards keep the superset property (`fires(title) OR fires(description) implies fires(combined)`) true as patterns evolve: a per-entry synthetic property test in both runtimes, and an anchor-shape source test whose family list is AST-derived from `_CONTENT_PATTERN_SETS`, so a new content set is covered by default.

Decisions of record: field attribution in `elimination_reason` (title vs description) is deliberately lost for the four moved sets — uniform with shell/url/encoded. Reason ORDER changes when a title-borne moved-set payload co-fires with heuristic 4 or 7 (the moved sets now fire earlier; `reasons[0]` feeds the operator-facing stderr line) — pinned by a test. A forged incoming `title_scan_matched` key now passes through inert (zero consumers tree-wide, grep-verified).

Measured: junction-straddle delta 0/727 per set on the real corpus, reverse delta 0/727 (nothing eliminated today becomes kept). 11 parity goldens re-recorded; diffs are prefix drops and stamp removal only.

## #254 — four shipped shape fixes, four measured declines

Acceptance ran a four-gate rule per candidate: G1 corpus zero over a pinned recipe; G2 near-miss token density (a zero over unexercised vocabulary is absence, not evidence — the rule #251 stated and the #252 url entries proved); G3 a blind-authored legit-voice negative set that must not fire; G4 twin-divergence probes (avoided by using only unbounded delimiter-terminated classes — no bounded windows, no bare `\s`).

Shipped:
- url `visit`/`download from` widened from `https?://` to scheme-general `[a-z][a-z0-9+.\-]{1,15}://` — the imperative is the discriminator, not the scheme; closes ftp/ftps/sftp/scp in one edit. shell's curl/wget keep `https?://`: not a measured gap, no lure missed it.
- body_marker `<finding>`/`<example>` widened to attribute-tolerant `<finding(?:<WS>[^>]*)?>` form. `</finding>` declined: an injected block always opens, and the closing tag false-fires on legit unbalanced-tag findings.
- body_marker gains a placeholder-noun-gated entry `\[<WS>*INSERT\b[^\]]*\b(?:FINDING|TITLE|TEXT|PLACEHOLDER|HERE)\b[^\]]*\]`, appended after the existing `[INSERT]` entry so first-match reasons stay stable. The issue's bounded-interior spelling was declined: it fires 20/23 on blind legit SQL-privilege texts (`[INSERT, UPDATE, DELETE]` grants), and INSERT has real corpus exercise (27/727 bench, 5/718 PR comments).
- Six patterns still using a literal ASCII space respelled to the union whitespace class (url `download from`, body `lorem ipsum`, the four title-placeholder `<word> finding` entries) — the separator-brittleness axis the `bypass_separator_*` fixtures already treat as an invariant; strictly safer (matching changes only for strings already containing the phrase).

Declined, with executed legit-voice fires recorded on #254:
- instructional "requires you to run" (fires on real conventions/runbook findings) and the paste-object loosening (fires on a real docs-review finding).
- bypass "bypass security review": all 161 corpus `bypass*` occurrences are descriptive third-person about real vulnerabilities — the collision voice, measured. The already-shipped `skip<WS>+review` bigram has the same problem today (2 executed legit fires); filed as an adjacent issue rather than widened further.
- vuln_intro inflections ("disabling TLS verification"): the bare stem IS the discriminator — imperative is injection voice, inflection is the voice of a legitimate finding describing a real vuln. Class rule of record: no inflection-tolerance for the imperative-gated sets (vuln_intro/bypass/instructional); tolerance only for structural markers where no descriptive voice exists.

Residual risk. The false-fire corpus is local, untracked scratch (`bench/workspace/`, gitignored): 727 title/description pairs from 30 run directories (findings files + checkpoint filter/challenge buckets, exact-pair dedupe) plus 718 PR/issue comment bodies from this repo, 63 authored hard negatives, 72 blind lures (the lure brief named `ftp://` as an example scheme, so the url catch number is partially non-blind). The corpus under-represents the colliding vocabulary for several candidates (TLS/SSL, ftp, terminal: 0 occurrences) — those were decided on the blind legit-voice gate, which is adversarial but finite. A shipped widening can still eliminate a legitimate finding whose phrasing this round did not anticipate; the failure mode is a lost finding, not a leaked payload, and is recoverable by narrowing in a follow-up. C1's PR-comment leg shows 3/718 fires — all pre-existing meta-discussion comments quoting the filter's own `visit https://` vocabulary, not the widening's delta. C2's blind finding-markup negatives fire 19/23, of which 15–16 already fire at HEAD on the bare `<finding>`/`<example>` literals (the tool's own wire format colliding with findings that discuss it — filed as an adjacent issue); C2's net-new is 3/23 inside that same already-condemned class, accepted for uniform marker detection (a marker set that catches `<finding>` but not `<finding severity="x">` is trivially evadable).

## #253 — replayed challenge checkpoints re-enter the current filter

`PERSISTED_RESUME_PHASES` replays a challenge checkpoint verbatim, so a checkpoint recorded by an older pipeline bypassed every filter improvement since. The #213 strip belt closed that for the prose fields only. This PR replaces it with a filtering belt: replayed `.findings`/`.unverified` run through `applyReplayInjectionScan`, a new twin-paired export that runs the injection scan minus heuristic 4 (short-description + high-confidence). h4 is excluded by construction: it reads `confidence`, which `detectDisagreement` boosts (+10) after the record-time scan, so re-running it would eliminate legitimately corroborated findings — the round's red-team proved that with a live construction, and a runWith-level test now pins the call site.

Mechanics: `normalizeFieldNames` runs first (a v2-era checkpoint carries `body`/`line`); non-object elements keep their positions (the #213 tolerance tests are the oracle); survivors get the kept-path prose strip inline (byte-equal to `applyInjectedProseStrip`); eliminations append to `.eliminated` with the standard stamps plus `replay_belt: true`, and the whole `.eliminated` array is stripped AFTER the append so a newly-eliminated finding never re-persists a raw citation field. `stats.final_count`/`skipped` are assigned from the arrays the belt just wrote (correct by construction on malformed input); `stats.replay_belt_eliminated` carries this call's count. Disclosure is one `replay-filter:` gap line derived from the marked entries in `.eliminated`, so it survives resume-of-a-resume; the malformed-`.eliminated` drop corner accumulates a durable `stats.replay_belt_dropped` and derives its gap from that. `replay-filter:` is registered in docs/machine-parsed-strings.md. The all-degraded and empty-report guards are untouched; the belt comment documents both empty-report corners (an all-eliminated replay now legitimately yields an empty ok:true report, disclosed via the gap line). On a fresh run the belt is a no-op by construction (this run's filterStage already ran the same content scan).

Measured: 0/449 eliminations across the kept challenge buckets of all 47 real checkpoints in the local bench corpus, every heuristic — the belt changes nothing for any checkpoint a shipped version has recorded; it exists for the version-delta and hand-edited cases.

The challenge outcome buckets (`challenge_survived`/`contested`/`downgraded`) remain pre-belt counts, stated in the adjustment comment.

## Validation

Suites-only per the wave rule; the recall/noise-sensitive deltas ride the wave-end paired mini. Both suites green (pytest 1798 passed + 829 subtests; node 1010/1010), goldens fresh (`record_parity.py --check`), bundle byte-fresh, pre-commit clean. CI coverage on this PR: Python scripts 93.86 (floor 92.9, headroom 0.96pp), JS 99.05/87.60/98.52 (floors 98.1/86.6/97.6, headroom 0.95/1.00/0.92pp) — no gate's headroom exceeds 1.0pp, so no floor moves in this PR, per policy.

Process: three-reviewer adversarial red-team of the design before implementation (executed evidence, not argument — it rejected the belt's first shape and rewrote the #254 acceptance rule), per-stage independent verification, then a four-lens fan-out review with per-finding adversarial refutation: 19 findings, 12 confirmed (all test-coverage/hardening/record class, no live behavior defect), 7 refuted, all 12 fixed in the final commit with every review mutation re-run red-first. Mutation ledger: every shipped mechanism reverted red-first with byte-verified restores, including the five pattern edits the first sweep missed.

## The adjacent round (#258–#261), resolved in this same PR

The four issues the first round surfaced were each taken through the same design → measure → adversarial re-measurement cycle. Outcomes:

- **#260 — executed.** The placeholder-title set now keys on stub vocabulary: the three bare-word entries (`TODO`/`FIXME`/`Placeholder`) are dropped — in 30 recorded runs their real fires were 100% false positives (5/727 titles, all genuine findings about residue) with zero observed stub catch — and a `Placeholder finding` entry joins the four two-word stub entries. A first candidate (title-initial anchor + 6-word bound) was falsified by re-measurement before shipping: the stub and legit populations are inverted on both length and position, so only the vocabulary discriminator separates them. Regression tests pin both directions; a bare `TODO: fix the null check in auth.py` title is now kept.
- **#261 — executed.** The injection-artifact reference (canonical file + seven mirrored agent copies) now states the combined title-and-description scope once at the section intro; the short-description bullet stays description-only because heuristic 4 genuinely is.
- **#258 — ratified, closed without code change.** The directive-gated `skip review` replacement was designed and then rejected by independent re-measurement: the gated grammar still eliminates 12/14 realistic legit review-gate findings while losing 9/20 independently-authored injection phrasings, because a legitimate finding states "skip review for this PR" about its own subject — the #252 un-narrowable class. The entry stays; the drop option is pre-measured in the issue record (siblings catch 2/20).
- **#259 — ratified, closed without code change.** Every code-span exemption form is a two-character evasion (20/22 caught lures escape by backtick-wrapping); the alternative of deleting the tag entries loses a fabricated-finding forgery class 10/10 end-to-end, and every measured recovery entry false-fires 20/20 or 11/12 in legit voice. Rule of record: narrowing a deny-pattern costs the attacker capability; an exemption keyed on formatting of the same attacker-controlled string costs two characters. Rates documented in the issue record.

https://claude.ai/code/session_011wE2DMoAzAdeVCp4xTUWuS

---

> [!NOTE]
> **Medium Risk**
> Changes what findings are eliminated or stripped on record and on challenge replay/resume; mis-tuned patterns or the replay belt could drop legitimate review output, though extensive parity tests and heuristic-4 exclusion mitigate confidence-related false eliminations.
> 
> **Overview**
> **Unifies prompt-injection detection** so bypass, instructional, vuln-intro, and body-marker heuristics scan `title` and `description` together (like shell/url/encoded already did), drops the separate title pass, and removes `title_scan_matched` / `stats.title_matches_eliminated`. Elimination reasons use bare phrases (no `description ` / `title ` prefixes). **Tightens patterns** (#254/#260): scheme-general URLs, attribute-tolerant `<finding>`/`<example>`, noun-gated `[INSERT …]` placeholders, union whitespace on stub titles (`Example finding` etc.) while **stopping** elimination on bare TODO/FIXME/Placeholder titles that describe real code residue.
> 
> **Adds a replay belt (#253)** after challenge: replayed or fresh `challengeOut.findings` / `.unverified` go through `applyReplayInjectionScan` (same scan **without** heuristic 4, so post-filter confidence boosts cannot wrongly drop corroborated findings). Belt-removed items land in `.eliminated` with `replay_belt: true` and a top-level `replay-filter:` gap; Python/JS twins stay parity-tested.
> 
> **Docs and contracts** align seven discovery agents plus `false-positive-exclusions.md` on title/description vs field-strip behavior; `TestPromptInjectionArtifactsMirror` byte-pins the seven agent copies. Parity goldens and a large test expansion cover cross-field splits, isolating pattern mutations, and combined-scan superset properties.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3b43f74200ec99c3c4c2e3365ac85540bf5a6e91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/filter_findings.py
```diff
@@ -420,21 +419,45 @@ def apply_threshold_filter(findings, config):
...
+# "lorem ipsum" picked up the union whitespace class (previously a literal
+# space).
 _INJECTION_BODY_PATTERNS = [
-    r"<finding>",
-    r"<example>",
+    r"<finding(?:[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff][^>]*)?>",
+    r"<example(?:[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff][^>]*)?>",
     r"\[[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]*INSERT[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]*\]",
-    r"lorem ipsum",
+    r"\[[\t\n\x0b\x0c\r \x1c-\x1f\x85\xa0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff]*INSERT\b[^\]]*\b(?:FINDING|TITLE|TEXT|PLACEHOLDER|HERE)\b[^\]]*\]",
```
leehopper, line 459: **💡 [LOW] No negative test proves the noun-gated [INSERT...] pattern spares the SQL privilege-list false-positive it was designed to avoid**

The #254 widening added a placeholder-noun-gated `[INSERT ...]` body-marker pattern, and the code comment (both runtimes) explicitly states the noun gate exists specifically to avoid colliding with a real security finding quoting `[INSERT, UPDATE, DELETE]` (a SQL privilege list); no test in either tests/test_filter_findings.py or workflows/test/filter_unit.test.js asserts that such a finding is kept, not eliminated.

**Suggested fix:**
Add a kept-legit test on both runtimes, e.g. a finding whose description reads something like 'The stored procedure grants [INSERT, UPDATE, DELETE] privileges to the public role, which is broader than intended' and assert apply_injection_filter/applyInjectionFilter returns it in `passed`/`kept`, not `eliminated` -- mirroring the existing kept-legit carveout tests (test_title_base64_carveout_kept etc.) already in the same file.

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: a25eb3abeef7aec5f9da2a63881b84f88b7d100f

scripts/filter_findings.py
leehopper: Fixed in 3b43f74: kept-carveout tests in both runtimes pin the SQL privilege-list payload as kept, red-verified against the declined bounded-interior spelling.

PR #269, opened 2026-08-31
fix(filter,pipeline): type-coerce scanned finding fields, count exclusion eliminations, disclose the discarded resume (#266, #267, #268)

leehopper:
Closes #266. Closes #267. Closes #268. (#265 is deliberately NOT closed — it is re-scoped by comment; see below.)

## What this does

**#266 — typed-field coercion across the filter twins, fixed class-wide.** The issue named one site (`apply_exclusions` title/description: Python's f-string renders a stamped `None` as the literal string "None" while JS's `|| ''` renders nothing). Recon and two review rounds found the class was much wider, so the rule shipped is: *a scanned finding field contributes a value of its expected type or that type's default — never a stringified null, never a crash*. Shared helpers (`_as_text`/`_as_confidence`, `asText`/`asConfidence`) applied at every site, both twins:

- `apply_exclusions` title/description
- the injection scan core: title/description/file (Python previously **crashed** on a null title at heuristic 7; `file: null` produced a live `elimination_reason` divergence), confidence (Python crashed at heuristic 4; a numeric-string `"90"` bypassed it in JS)
- `detect_disagreement` suppression rules 1 AND 2 (Python crashed; JS silently suppressed on an array title — a live recall divergence)
- `_route_by_dimension` and its duplicate `_is_test_correctness_finding` (live report-destination divergence on a non-string title)
- `consolidate_cross_agent`'s winner key (`description: null` crashed **both** twins through the live `applyFilterPipeline` path)
- `apply_threshold_filter` confidence + severity (null confidence crashed the Python CLI before anything else ran; null severity crashed both twins)

`_as_confidence` excludes bools and maps NaN to 0 in both twins (guards mutation-pinned). Seven new parity fixture groups pin the crash oracles and the cross-language discriminators (array-typed companions where a bare null is inert in one twin); every pre-existing golden is byte-identical.

**#267 — `exclusions_removed` stat, both twins.** Placed adjacent to `injections_removed` in both stats literals and added to the Python module's Output-JSON-schema docstring (its only written contract). Verified against every consumer: the stats object flows opaquely to `reportInput.stats.filter` and the envelope; nothing enumerates keys. Tests drive the real entry points (`main()` with a temp REVIEW.md ignore pattern; `applyFilterPipeline` directly).

[...]

Issue #268, opened 2026-08-31
top-level garbage checkpoints silently discard the resume and re-run the whole review with no disclosure


leehopper:
## Problem

`readCheckpoints` resolves a garbage top-level `checkpoints` value (a string, an array —
arrays since PR #264 — or a wrapper whose `.phases` is not a plain object) to `{}`: the run
dispatches everything fresh, `ok:true`, correct answer — and the operator's resume intent is
silently discarded, with a full re-billed run and no gap saying so. The inertness itself is
the decided F4-3-adjacent behavior (a stamped `checkpoints: null` must stay disclosure-free);
the open question is whether a NON-null garbage value deserves a disclosure gap, since #248's
premise is that a malformed checkpoint must not fail invisibly, and PR #264's shape gate only
covers malformed values inside a well-formed map.

## Fix shape

Decide: either ratify the silent re-run (cheap, document it at `readCheckpoints`) or push a
`checkpoint-shape:`-family disclosure gap when a non-null `checkpoints` resolves to the empty
map. If disclosed, keep `checkpoints: null` exempt (the F4-3 decision). Suites-only.

Surfaced by the PR #264 round-2 red-team.

leehopper:
## Executed 2026-08-31 — PR #269 MERGED (ef2401a)

Decision: disclose, degrade-and-continue. One `checkpoint-discarded:` gap (registered in docs/machine-parsed-strings.md) when `args.checkpoints` was supplied, non-null, and unusable — string/array/corrupt-`.phases` shapes, none of which has a legitimate producer. The gap names the failing position (`checkpoints` vs `checkpoints.phases`) and shape, rides both exits and the checkpoint-shape reject envelope, fires exactly once, and never enters phaseOutputs (no resume-of-a-resume double count). `checkpoints: null`/absent stay silent — the F4-3 decision untouched, its test byte-identical. NOT a validateArgs refusal: PR #264's oracles pin garbage top-levels dispatching the full fresh pipeline ok:true, and a refusal would reverse that while missing wrapper shapes. `readCheckpoints`' unwrap hoisted to `unwrapCheckpointMap`, behavior unchanged. Residual: a fed-back `{completed, truncated: true}` return unwraps to an inert bare map and stays silent — filed as #270.

PR #271, opened 2026-08-31
fix(pipeline): disclose the no-op resume a fed-back compact or truncated return produces (#270)

### LINKED ISSUE #270: a fed-back truncated checkpoints return silently re-runs the whole review, invisible to the checkpoint-discarded disclosure
## Problem

`readCheckpoints` accepts EITHER a bare phase map or a `{phases, completed}` wrapper. A fed-back `{completed, truncated: true}` return — the one garbage-adjacent shape an operator plausibly produces, since SKILL.md documents it as a real return and headless auto-resume reads `return.checkpoints` directly — unwraps "successfully" to an inert bare map (its keys are not phase names), so every phase re-runs with no disclosure and a full re-billed run. The PR #269 `checkpoint-discarded:` disclosure cannot see it: its predicate keys on the unwrap FAILING, and this shape unwraps fine.

SKILL.md already instructs against re-feeding a truncated return, so the exposure is a skill mis-branch, not a documented flow.

## Fix shape

Decide between: (a) ratify (document at `readCheckpoints` that an all-unknown-keys map is an inert no-op resume, keep the silence); (b) extend the disclosure to a resolved map that contains zero recognized phase names while the caller supplied a non-empty object — must NOT fire for `{}`/`{phases: {}}` (legitimate empty) and must stay silent for `checkpoints: null`/absent (the F4-3 decision). Suites-only.

Surfaced by the PR #269 review round.

leehopper:
Closes #270.

## What this does

The `checkpoint-discarded:` disclosure (from #268/PR #269) gains the no-op-resume arms #270 asked for. Decision: **extend the disclosure** rather than ratify the silence — every shape the new arms catch is a documented operator mis-branch with no legitimate producer, and the one shape operators plausibly produce (a fed-back return envelope) was the one still silent.

- **Arm A — zero-recognized non-empty map** (`unwrapCheckpointMap` succeeded, ≥1 own key, none of the eight phase names): three fixed sub-case messages — a truncated failure return ("nothing is resumable, re-run without checkpoints"), a compact return with a non-empty `completed` (points at the on-disk `artifactPaths.checkpoints`), an empty-`completed` return envelope or an unrecognized map (fresh-run advice / the eight names listed).
- **Arm B — empty `.phases` masking**: a wrapper whose `.phases` is empty while recognized phase names sit ignored at its top level.
- **Every message is fixed pipeline-authored text.** The original design listed the caller's key names in the gap; the pre-implementation red-team executed that as an injection: a checkpoint key literally named after a bench G3 sentinel would have flipped gate `G3` for the run (`_DEGRADE_RE` scans every gap), and the checkpoint file is PR-author-reachable. No caller bytes reach the gap channel; a test drives all arm texts against the two forbidden sentinels case-insensitively.
- Silence preserved and pinned: absence, `checkpoints: null` (F4-3 — its test byte-identical to main), `{}`, `{phases: {}}`, `buildResumeCheckpoints({})`, a `slimPersistedCheckpoints`-shaped real resume (the primary legitimate producer — its silence pin kills a reviewer-found surviving mutation), every single-phase bare map (looped over the hard-coded `ALL_PHASES` list, independent of the table it guards), partial `.phases` masking (a stated non-goal — the resumed phases are real), and ctx-only checkpoints.
- Exactly-once by construction: the unwrap-fail arm and the new arms are mutually exclusive, the shape gate never inspects unrecognized keys, and the ctx-seam test pins that a garbage args value plus a ctx fallback yields one gap, not two.

## Producer enumeration (the evidence the decision rests on; reviewer-executed)

Fires (all documented operator mis-branches): the ok:true compact return `{completed: [...]}`; the truncated failure return `{completed, truncated: true}`; the checkpoint-shape refusal envelope's `{completed: []}`. Silent (all legitimate): the persisted slim checkpoint artifact (`{phases: {...}, completed, phaseReached, counts}`) — the documented resume input; `buildResumeCheckpoints` success and first-phase-crash shapes; every bare map with ≥1 phase key. A zero-recognized non-empty map has no legitimate producer.

The disclosure does not prevent the re-billed run (the gap rides the completed run's envelope); its value is attribution and a correct next attempt.

## Validation

One-opus pre-implementation red-team (REVISE: killed the key-listing injection, added arm B, the ALL_PHASES silence loop, the ctx-seam pin) → sonnet implementation → one-opus adversarial review (FIX_FIRST: the compact-return message wrongly pointed at `artifactPaths.checkpoints` for the artifact-less refusal envelope; three surviving mutations M8/M5/M3 now each killed by a named test) → fix round. Suites: 1171 JS / 1832 + 840 Python green; JS coverage 99.11 / 88.21 / 98.56 (floors 98.1 / 87.0 / 97.6); bundle byte-stable; pre-commit clean; all pre-existing tests byte-identical to main. Suites-only; the disclosure only fires on shapes with no legitimate producer, so no recall/noise delta.

Floors: branches headroom will exceed 1.0pp — the ratchet lands as a follow-up commit on this branch from the PR's own CI measurement, per policy.

https://claude.ai/code/session_01JnbCPJi3FWXb1JdrRy17HE


workflows/src/stages.js
```diff
@@ -3603,33 +3603,138 @@ export function readCheckpoints(ctx, args) {
...
+      // a path that was never written.
+      if (Array.isArray(resolved.completed) && resolved.completed.length > 0) {
+        return ['checkpoint-discarded: the supplied checkpoints value looks like a compact return rather '
+          + 'than a checkpoint artifact -- it carries no phase outputs to resume from; the on-disk resume '
+          + 'state from that run lives at artifactPaths.checkpoints.'];
+      }
+      return ['checkpoint-discarded: the supplied checkpoints value looks like a return envelope rather '
+        + 'than a checkpoint artifact -- it carries no phase outputs to resume from; re-run without the '
+        + 'checkpoints argument to start the review fresh.'];
+    }
```
leehopper, line 3711: **💡 [LOW] checkpointDiscardGap's non-array `completed` sub-case (explicitly called out in the comment) has no test**

The Arm A branch that inspects a resolved checkpoints map's `completed` key distinguishes a non-empty array (on-disk pointer message) from an empty-or-non-array value (return-envelope/fresh-run message), and the comment at line 3698 explicitly says 'An empty (or non-array) completed instead matches...' — but every test fixture in checkpoint_shape_gate.test.js sets completed to either [] or ['summarize'], never a non-array truthy value such as a string or number.

**Suggested fix:**
Add a test like `validArgs({ checkpoints: { completed: 'not-an-array' } })` asserting the gap text matches the return-envelope/fresh-run wording (not the on-disk-pointer wording), confirming the Array.isArray guard — not just `.length > 0` — drives the branch.

leehopper (commented):
code-gauntlet review: 1 low-severity suggestion found. See inline comment.

---
Generated by code-gauntlet | Reviewed up to: d93168cae416daeda08433b4f548202345b747fb

leehopper:
Review finding addressed in 5a09366: added the non-array-`completed` test (`{completed: 'not-an-array'}` → return-envelope wording, not the on-disk pointer) and mutation-verified it — weakening the `Array.isArray` guard to `.length > 0` alone turns exactly this test red (a string's truthy `.length` would otherwise take the pointer branch).

PR #240, opened 2026-08-26
fix(filter): pin the filter twins' unicode word-boundary/whitespace/fold semantics (#211, folds #214)

leehopper:
Closes #211. Folds #214.

## Decision

Option (a), generalized after measurement: pin the Python filter twin to engine-independent semantics rather than ratify the divergence, across the whole first-party pattern surface — not just the encoded set the issue names.

- `re.IGNORECASE | re.ASCII` on every first-party fixed-pattern compile/call site in `scripts/filter_findings.py` (8 injection sets, 2 routing keyword sets, 25 test-correctness patterns, 2 suppression rules). With `re.ASCII`, Python's `\w \b \B \W \d \D` and IGNORECASE folding equal JS non-`u` semantics exactly (measured; the folding delta for ASCII patterns is 94 ordered pairs over 82 codepoints, all Python-side, e.g. `ſkip review` eliminating in Python only).
- `\s` cannot be aligned by any flag (`re.ASCII` shrinks it to 6 chars while JS keeps 25), so every first-party `\s` in BOTH twins is respelled to an explicit 30-codepoint union class (the union of Python's 29 and JS's 25 whitespace sets), byte-identical between the twins. Membership pins in both engines' unit suites enumerate the class per engine and assert the exact 30-codepoint set.
- `_count_words`/`countWords` split on the same class and count non-empty parts (the `str.split()` vs `split(/\s+/)`/trim asymmetry was a measured divergence on the same six codepoints, and the first rewrite's `.replace(re)` non-global bug was caught and killed in adversarial review).
- The file-path template check's `.` respelled to `[^\n]` in both twins (JS `.` excludes CR/U+2028/U+2029; Python's excludes LF only — a JS-only shipped-behavior change on paths carrying those characters).
- **`apply_exclusions` is deliberately exempt** from `re.ASCII`: its patterns are user-authored, arbitrary-script text, where the flag breaks `café` ↔ `CAFÉ` matching that JS keeps (measured — the flag would create divergence, not close it). A fixture pins this so a future "fix" goes red.

Direction-of-authority: `record_parity.py` records Python's output as the parity golden, so every a-variant "edits the reference to match the implementation." Resolved by dissolving the question — with explicit classes plus `re.ASCII`, both engines express the same semantics and the pattern source is the authority, not either engine.

## Why not the alternatives

[...]

Issue #211, opened 2026-08-18
unicode word-boundary divergence between the filter twins on the encoded-payload set


leehopper:
## Problem

Python `\w`/`\b` are unicode-aware; JS's are ASCII-only. The encoded-payload patterns
`(?<!\w)(?:0x)?[0-9a-fA-F]{32,}(?!\w)` and `\b[A-Za-z0-9+/]{40,}={0,2}\b` therefore disagree
between the twins whenever a non-ASCII letter abuts the run.

Measured during the #62 review (#210): `description` = clean prose, `suggestion` (or
`description`) containing `é` + 32 hex chars → Python keeps the text unscathed; JS matches and
strips (post-#62) or eliminates (pre-#62 semantics on `description`). Two runtimes, two different
posted-comment bodies for one finding.

The class is pre-existing on the `description` path — #62 only extended the mismatched surface to
`suggestion`. The parity suite cannot see it: a fixture exercising it would be red today, which is
why none exists.

## Options

(a) pin the Python patterns to ASCII semantics (`re.ASCII` on the encoded set) so both twins match
the JS behavior — a behavior change on the live description path, so recall/noise-sensitive;
(b) accept and document the divergence as out-of-parity-contract. Either way, the decision should
land with a fixture or an explicit register entry.

Found by adversarial review on #210.

PR #273, opened 2026-09-01
fix(filter): pattern registry + generator, config-parser/homoglyph/whitespace twin convergence (#241-#244)

### LINKED ISSUE #241: pattern registry + generator for the filter twins (one source, both twins emitted)
## Problem

PR #240 (#211) aligned the filter twins' pattern semantics and added a structural guard
(`tests/test_filter_twins_unicode_guard.py`) that asserts element-wise byte-identity of every
first-party pattern family between `scripts/filter_findings.py` and
`workflows/src/filterFindings.js`, plus the three inline sites and the word splitter. The guard
asserts identity; nothing produces it. Every pattern edit is still a coordinated two-file change,
and the union whitespace class is an 88-character literal repeated at ~60 sites per twin.

## Fix shape

The #238 shape: one registry declaring each pattern family once, a generator emitting both the
Python lists and the JS literal arrays (marker-fenced blocks, `--check` mode, pytest freshness
test), with the union class expanded at generation time from a single declaration. The guard then
becomes a backstop rather than the only line of defense, and "extending costs one edit" holds.

## Notes

- The generator must preserve the twins' current byte-identical spellings (the guard pins them).
- `apply_exclusions` stays out of scope (user-authored patterns; deliberately not re.ASCII —
  see the exemption comment at its call site and the café parity fixture).

Surfaced by the #211 pass (PR #240).

### LINKED ISSUE #242: live homoglyph case-fold evasion of the shipped injection filter
## Problem

The shipped JS injection filter is evadable by unicode case-fold homoglyphs. Measured at HEAD
during the #211 red-team (all caught by no pattern set):

- `ſkip review` (U+017F LATIN SMALL LETTER LONG S) evades the bypass set
- `diſable TLS verification` evades the vuln-intro set
- `dısable TLS verification` (U+0131 dotless i) evades the vuln-intro set

Python's unicode IGNORECASE used to catch these, but that twin has no live consumer, and #211
(PR #240) pinned it to the same ASCII folding for parity.

## Constraints measured during #211

- `/u` is NOT a safe fix: it rejects the URL pattern (`Invalid escape` on `\"` identity escape)
  and moves JS folding toward full-unicode, away from the pinned twin semantics.
- The whole ASCII-target fold surface is small: 94 ordered pairs over 82 BMP codepoints (see
  docs/duplication-register.md row for the enumeration). Candidate fixes: explicit alternations
  for the three-codepoint core (ſ, K, ı) or a confusables normalization pre-pass before the
  pattern scan in both twins.

Recall/noise-sensitive: rides a paired measurement per bench/MEASUREMENT.md.

Surfaced by the #211 pass (PR #240).

### LINKED ISSUE #243: parse_review_md regex family: inconsistent /i flags and unicode-\d twin divergence
## Problem

`parse_review_md` and `load_exclusions` (both twins) were declared out of scope by the #211 pass,
but carry the same engine-dependent construct classes as the finding-text patterns it fixed:

- Inconsistent case-insensitivity across the five directive parsers: `severity_threshold` and the
  block patterns carry `re.IGNORECASE`/`/i`; `confidence_threshold`, `security_min_confidence`,
  and `ignore_section` carry no flag on either side.
- Unicode `\d` divergence, measured: `confidence_threshold: ٧٥` (Arabic-Indic digits) parses as 75
  in Python (unicode `\d` + `int()` accepts it) and is silently ignored in JS.
- `\s`/`[\s]` sites with the same 6-codepoint Python-vs-JS membership delta #211 closed elsewhere.

This parses REVIEW.md (repo-authored config), not attacker-echoed finding text, so severity is
lower — but the twins can disagree on what a user's config means.

## Fix shape

Audit the family against the #211 mechanism: explicit classes + `re.ASCII` where patterns are
first-party fixed syntax, flag-parity decisions per directive, and extend the structural guard's
scope (or the future pattern registry) to cover them. Decide deliberately whether directive
digits should be ASCII-only in both twins.

Surfaced by the #211 pass (PR #240; measured in the pre-implementation red-team).

### LINKED ISSUE #244: host-language whitespace/case skew in dedup signature, pyIntOrNull, and lower() keying
## Problem

Host-language string primitives — not regexes — still give the filter twins different notions of
whitespace and case, in keying/coercion positions. Registered in docs/duplication-register.md by
PR #240; this issue is the tracked home.

Measured mechanisms:

1. Dedup signature: `title.lower().strip()` (Python) vs `title.toLowerCase().trim()` (JS).
   Divergent on U+001C-U+001F and U+0085 (Python strips, JS does not) and U+FEFF (JS strips,
   Python does not). 186/2000 cases in the #211 round-2 seeded fuzz corpus; pre-existing,
   identical on main and post-#240.
2. `pyIntOrNull` (`/^\s*[+-]?\d+\s*$/`) vs Python `int()`'s own whitespace stripping in
   `_line_bucket` on the live detectDisagreement path: `U+FEFF + "12"` suppresses in JS only;
   `U+0085 + "12"` suppresses in Python only.
3. `.lower()` vs `.toLowerCase()` UCD-version skew: 95/55/28 divergent codepoints on CPython
   3.10/3.12/3.14 vs Node 24 — drifts with every runtime upgrade; unclosable by flag.

## Fix shape

For (1) and (2): respell both twins onto the #211 union whitespace class (Python
`strip(<chars>)`/regex, JS anchored replace; `pyIntOrNull` class respell) and extend the
cross-twin behavioral table. That changes shipped JS dedup/suppression behavior on freak
codepoints, so it rides a paired measurement window. For (3): register-only unless a fixture
ever depends on it; no flag can close UCD skew.

Surfaced by the #211 pass (PR #240, round-2 fuzz).

### LINKED ISSUE #272: Injection filter: confusables (NFKC lookalikes) + zero-width/format-char boundary breakers still evade after #242
## Summary

`#242` closes the **casefold-reachable** homoglyph class in the injection filter: the four
codepoints that case-fold to a plain ASCII letter (U+017F ſ, U+0131 ı, U+0130 İ, U+212A KELVIN K)
are now folded in a UNION scan (`_injection_scan_core` / `injectionScanCore` and the prose-field
strip) so a payload like `ſkip review` is eliminated instead of walking through.

That map is deliberately only four codepoints — it is the *complete* set reachable by case-folding.
Two strictly-cheaper evasion classes remain open, and both were **measured evading at HEAD and still
evading after #242's fold**:

### 1. NFKC / confusable lookalikes (need a lookalike glyph, but not a case-fold one)

924 single codepoints NFKC-normalize to a single ASCII letter; #242 folds none of them. Measured
KEPT at HEAD and after the #242 fold, driving the real `apply_injection_filter` / `applyInjectionFilter`:

- `ｓkip review` — U+FF53 FULLWIDTH LATIN SMALL LETTER S
- `ѕkip review` — U+0455 CYRILLIC SMALL LETTER DZE
- `𝗌kip review` — U+1D5CC MATHEMATICAL SANS-SERIF SMALL S
- `sₖip …` — U+2096 LATIN SUBSCRIPT SMALL LETTER K
- `dⅰsable TLS verification` — U+2170 SMALL ROMAN NUMERAL ONE
- `dｉsable …` — U+FF49 FULLWIDTH LATIN SMALL LETTER I

### 2. Format-character / zero-width boundary breakers (need no lookalike glyph at all)

These insert an invisible codepoint that is NOT in the union whitespace class (which stops at
U+FEFF) between letters, defeating every `\b`-anchored pattern without changing any visible letter.
Measured KEPT at HEAD and after the #242 fold:

- `skip​review` — U+200B ZERO WIDTH SPACE between the words
- `sk‍ip …` — U+200D ZERO WIDTH JOINER
- `sk­ip …` — U+00AD SOFT HYPHEN
- `sk̇ip …` — U+0307 COMBINING DOT ABOVE
- `skip︎ …` — U+FE0E VARIATION SELECTOR-15

## Why #242 does not (and structurally cannot) cover these

#242's fold is a hand-pinned case-fold map — it is version-immune precisely because it is not
derived from a Unicode runtime. A confusables/NFKC pre-pass would be a **twin-parity hazard**: the
two runtimes ship different Unicode versions (measured: CPython `unicodedata.unidata_version` =
16.0.0 vs Node `process.versions.unicode` = 17.0), so a runtime-normalized fold would diverge the
Python and JS twins. The format-char class is a different mechanism again (boundary insertion, not
glyph substitution). Both need their own design — hand-pinned confusables table and/or a
zero-width/format-char stripping pre-pass — not an extension of the case-fold map.

## Scope note

`closes #242` in the pattern-registry-cluster PR claims the **casefold-reachable class only**.
This issue tracks the confusables + format-char residual it deliberately leaves open.

Evidence: probes measured in the adversarial design review for #242 against the shipped
`apply_injection_filter` / `applyInjectionFilter` at HEAD and after the fold.

leehopper:
Closes #241, closes #242, closes #243, closes #244, closes #272.

Four staged commits (one per issue) off the #211 (PR #240) surface, plus a fifth closing the
adjacent evasion residual #242 filed as #272. Each stage was
implemented and independently adversarially verified before the next; a whole-branch
cross-stage review ran before this PR. Every "measured" claim below was checked against
the real twins by a probe, not asserted.

## ⚠ Measurement window (read first)

- **S1 (#241) and S2 (#243) are suites-only** — no shipped-behavior change to the review
  output (S1 is a pure source refactor behind a generator; S2 converges config parsing the
  live JS twin already gated).
- **S3 (#242) and S4 (#244) are live JS changes on the recall/noise-sensitive injection +
  delivery surface.** They ride the owner-triggered **paired-mini** window per
  `bench/MEASUREMENT.md`. **The paired-mini measurement was NOT run in this session.**

## S1 — pattern registry + generator (#241) · `feat(filter)`

One source of truth for the filter twins' pattern lists, with a generator that emits both
twins and a freshness check, so the #240 byte-identity guard becomes a backstop rather than
the only line of defense and "extending costs one edit" holds.

- New `scripts/filter_patterns_registry.py` (data-only) + `scripts/generate_filter_patterns.py`
  (`--check` mode). Per-block **id-keyed** fence markers (the single-spelling splice from
  `generate_contract_requirements.py` hard-fails at >1 open marker — measured; not inherited),
  plus an anchored-line-rewrite kind for the two in-function suppression regexes. Hand-written
  prose decision-records stay **outside** the fences.
- Emission is the ruff-format normal form (three pinned Python layouts, all measured fixed
  points; a `ruff format --check` test proves the generator and the format hook never fight)
  and the guard-required JS shapes (two-line `compile_single`, `\/`-escaped literals, `tuple(...)`
  -wrapped `_CONTENT_PATTERN_SETS`).
- The registry was **bootstrapped programmatically** from the current twins via the guard's own
  discovery/de-escape helpers — no pattern hand-typed. Regeneration reproduces both twins
  **byte-identically except three disclosed canonicalizations**:
  - `filter_findings.py` `_TEST_CORRECTNESS_PATTERNS` rows for `vacuous`, `tautolog`, `deadlock`:
    the only 3 collapsed one-liners of 25, exploded into the magic-trailing-comma form the other
    22 already carry (both are ruff fixed points; the generator emits one shape).
  - `_FUNCTIONAL_VIOLATION_KEYWORDS` / `_TYPE_SAFETY_BUG_KEYWORDS`: implicit two-adjacent-string
    concatenation merged into the single literal the parser already produces.
- Guard edit: the byte-identity walk now sources family names from the registry rows instead of
  an `_INJECTION_`-prefix rule + hardcoded singleton tuples — no name allowlist inside the test.
- Census correction of record: the union whitespace class occurs **135× per twin** (129 `[inner]`
  + 6 `[-inner]`), not the 129 the #240 record implied.

## S2 — config-parser twin convergence (#243) · `fix(filter)`

`parse_review_md` / `load_exclusions` were NAMED-EXEMPT from the #211 discipline but carried the
same engine-dependent constructs. Every divergence below was measured on both twins; all now agree.

1. **Digits + bound.** `(\d+)` → `([0-9]{1,3})` on `confidence_threshold` / `security_min_confidence`,
   accepted only when `<= 100` (`review-md-spec.md` `<0-100>`), else ignored → defaults, both twins.
   Closes the Arabic-Indic `٧٥` skew (PY 75 / JS ignored) and the `int()`-vs-`parseInt` `>2^53`
   divergence (`999…` → PY exact int / JS `1e+21`).
2. **Flags.** Uniform IGNORECASE on the five directive parsers; Python adds `re.ASCII` across the
   family — closes the `ſeverity_threshold` / `# deep-revıew` block-marker fold (Python unicode
   `/i` folded U+0131/U+017F into ASCII key letters; JS `/i` never did).
3. **Anchors.** `^`+MULTILINE → `(?:^|\n)` — closes the JS `/m`-breaks-after-`\r`/U+2028/U+2029 skew.
4. **Universal-newline splitting.** New `_split_review_lines` / `splitReviewLines` split on
   `\r\n | \r | \n` in both twins, reproducing Python `open()`'s universal-newline translation —
   closes the `splitlines()`-vs-`split('\n')` list-split skew (U+000B/U+0085/U+2028) **and** the
   lone-`\r` case a first draft false-converged on (caught in verification, fixed before this PR).
5. **Explicit classes.** whitespace-semantic `\s` → the union class; bullet tail `(.+)$` →
   `([^\n]+)$`; DOTALL `.` / `[\s\S]` → `[^\x00]` (see disclosure).
6. Hoist to module-level constants folded into the registry (`_REVIEW_*`); NAMED-EXEMPT carve-out
   deleted (dict, `_scope` arm, dead `continue`s, docstring paragraph); discovery pins re-derived by
   running `_discover_python` (call total 42→43 for the new `re.split`; `block_patterns` →
   `_REVIEW_BLOCK_PATTERNS` func=None). `apply_exclusions` stays forbidden-ASCII, untouched.

Disclosure: `[^\x00]` is cross-twin **symmetric** but not behavior-preserving vs HEAD — a NUL in a
config-block body now falls back to the whole-file scan in **both** twins (`args.js` validates
`reviewMd[i].text` only as a string). The `[0-9]{1,3}`+`<=100` pair accepts `1000` as `100` in both
twins (symmetric, contract-pinned).

## S3 — casefold-reachable homoglyph union scan (#242) · `fix(filter)`

Closes the measured live evasion of the shipped JS injection filter (`ſkip review`, dotless-i,
Kelvin-K).

- **UNION scan, not text replacement.** Scan the raw text first (byte-identical to HEAD's matcher),
  then — only if the text carries a mapped codepoint — scan a folded copy as a second pass; the raw
  match wins reason selection. This is load-bearing: every content pattern is `\b`/`(?<!\w)`-anchored
  and ſ/ı/İ/K are non-word under ASCII, so **replacing** the scanned text with its folded form is
  non-monotone — it flips those boundaries and destroys 6 detections HEAD already makes, opening a
  one-character evasion (`ſskip review`). The union only adds detections. Measured: `ſkip review`
  KEPT→ELIM while all 6 destroy-risk shapes stay ELIM, identical in both twins.
- **4-codepoint scope**, hand-pinned in the registry as `ASCII_CASEFOLD_REACHABLE`
  (U+017F→s, U+0131→i, U+0130→i, U+212A→k) — independently re-measured complete for
  casefold-reachability. Runtime NFKC is rejected as a twin hazard (CPython UCD 16.0.0 vs Node ICU
  Unicode 17.0). Data-only in the registry; the two fold helpers are hand-written per twin
  (`str.translate`, no `re.*`; `String.replace` over a `/[…]/g` literal, never `.test(`) so the
  guard census is unmoved (43 / 3).
- Applied only at the injection scan (`combined` for h1/2/3/5/6/8, `title` for h7) and the per-field
  prose strip. Dedup signature (h10), word count (h4), `apply_exclusions`, dimension routing,
  test-correctness, `parse_review_md`: untouched.
- Base64 rejoin bonus: `[A-Za-z0-9+/]{40,}` breaks on an embedded ſ/K today; the folded pass rejoins
  the run — a real catch, paid for with no loss on the hex branches (which replace-text would lose).

Scope of "closes #242": the **casefold-reachable class only**. NFKC lookalikes (Cyrillic ѕ, fullwidth
ｓ, mathematical 𝗌) and zero-width / format-char boundary breakers are a strictly cheaper evasion
class left open, filed as **#272**.

## S4 — dedup-signature, int-coercion + review-line whitespace skew (#244) · `fix(filter)`

Host-language whitespace/digit skews outside the regexes, in keying/coercion/trim positions.

- **(a) union-class trim, shared across four sites.** `title.lower().strip()` / `.toLowerCase().trim()`
  (the dedup signature) plus the three per-line strips inside `load_exclusions`/`parse_review_md` (the
  fenced-block line, the bullet fallback, and the `parse_review_md` ignore item) all trimmed the
  language-native whitespace set — Python `str.strip()` removes U+001C–1F/U+0085, JS `.trim()` does not
  (and JS removes U+FEFF, Python does not). All four now strip the union class off both ends via one
  registry-sourced constant `_WS_TRIM_RE` / `WS_TRIM_RE` (`^[union]+|[union]+$`, JS GLOBAL). This closes
  the "user's exclusion pattern carrying a trailing control char silently fails to apply in the JS-live
  pipeline while applying in Python" divergence a whole-branch review measured — the same
  keying-position skew as the signature, at the same call-site class. `line_start` (the signature's 3rd
  field) stays raw. Count-parity-symmetric (reuse by name adds no new class literals).
- **(b) int coercion, one shared helper per twin reaching ALL bucket paths.** Python had **two**
  nested bucket closures (`_line_bucket` in `detect_disagreement`, `_bucket` in `group_by_proximity`
  feeding the delivered `consolidation_key` + imported by `apply_challenges.py`); patching only one
  would have created a new divergence. Both are replaced by one module-level `_line_bucket` over a new
  `_py_int_or_none` mirroring JS `pyIntOrNull` branch-for-branch, with the string gate capturing the
  digits and `int()`/`parseInt` run on the **capture, not the raw value**. This also fixes a JS **NaN
  corruption** (`parseInt('\x1c12')`=NaN → `consolidation_key` `file:NaN`) and closes the digit-class
  skew (Python `int()` accepted `١٢`/`１２`/`1_2`; the `[0-9]` gate now rejects them in both twins),
  and type-guards Python against a non-str `line_start` (no `.strip()` AttributeError).

Skew probe (both bucket paths, both twins converge to the same value after): U+001C/1D/1E/1F, U+0085,
U+FEFF prefix on `"12"` → bucket 10; `١٢`/`１٢`/`1_2` → rejected. Raw-number path unchanged
(25.7→20, 20→20, None→0, True→0).

Register: row 54 → `intentional-and-documented`, residuals named (parseInt 2^53 cliff on absurd digit
strings and its `_line_bucket` float>2^53 sibling, neither reachable at real line numbers; row-53 UCD
`.lower()` skew, separate and untouched); tally 33/11 → 34/11. Census 43→45 (+2 inert module-level
constants); count-parity 155→159, balanced.

## S5 — confusables fold + zero-width strip (#272) · `fix(filter)`

The adjacent issue #242 filed for its residual, now closed on this PR. #242 folded only the 4
casefold-reachable homoglyphs; two strictly-cheaper evasion classes stayed open and were measured
evading: NFKC/cross-script lookalikes (ｓkip, ѕkip Cyrillic, 𝗌kip math, dⅰsable, dｉsable) and
zero-width/format-char boundary breakers (skip‹ZWSP›review, sk‹ZWJ›ip, sk‹soft-hyphen›ip,
sk‹combining-mark›ip, skip‹VS15›). All eleven now eliminate in both twins.

- Two hand-pinned static tables composed into the **same** #242 raw-first-additive union scan
  (never a text replace): **CONFUSABLE_FOLD** (1468 single-codepoint→ASCII-letter entries; NFKC from
  CPython unicodedata 15.0.0 + cross-script complement from Unicode confusables.txt v17.0.0 + the
  casefold-4, precedence casefold>NFKC>confusables so ſ→s not the confusables ſ→f) and
  **INVISIBLE_STRIP** (599 zero-width/joiner/bidi/variation-selector/combining-mark codepoints,
  incl. the astral range U+E0100–E01EF), deleted from the scanned copy. Fold-keys and strip-keys are
  disjoint. The dedup signature stays raw (#244's surface, untouched).
- **Twin-safe by construction**: the tables are checked-in static data, never a runtime NFKC/normalize
  call — the CI Python matrix ships different UCD versions, so a runtime fold or a re-derivation test
  would diverge the twins. A sibling generator (`generate_confusable_tables.py`, `--check`-gated) emits
  both twins' packed forms from one registry source. JS folds by **codepoint iteration** (919 fold
  sources + 240 strip codepoints are astral; a non-`/u` char-class would corrupt them) — byte-identical
  to Python `str.translate`.
- **Safe without a paired-mini measurement**: folding is raw-first-additive, so it structurally cannot
  reduce recall; a noise battery measured **zero** over-elimination across 50 injection fixtures (0
  flips) and 42 legitimate multilingual/math/code findings. Still a live JS injection-surface change, so
  the paired-mini flag stands (not run this session).

## Verification

A whole-branch cross-stage review (emergent twin-parity fuzzing, security/noise, gate readiness) ran
after the four stages; its one MAJOR finding — the review-line strip skew above — is folded into S4,
and the rest came back clean. #272 was derived (Unicode data), noise- and twin-safety-validated, then
implemented and independently verified under the same pattern.

All gates green at HEAD (`ef22529`): `pytest tests/` + `bench/tests/`, `node --test`, bundle byte-stable
against a fresh `build.js`, `generate_filter_patterns.py --check`, `record_parity.py --check`,
`pre-commit run --all-files`. Coverage: scripts 94.3% (floor 92.9), bench 88.5% (87.5),
JS 99.2 / 88.4 / 98.6 (98.2 / 87.3 / 97.6). Floors **not** ratcheted from local numbers (per policy —
ratchet from CI measurement). Every new regression test was mutation-verified red against the bug it
names. Adjacent issue **#272** filed for the residual evasion class.


scripts/filter_findings.py
```diff
@@ -311,42 +436,42 @@ def parse_review_md(path):
...
-    # security_min_confidence
-    m = re.search(
-        r"^[ \t]*security_min_confidence\s*[:=]\s*(\d+)", block_text, re.MULTILINE
-    )
+    m = _REVIEW_SECURITY_RE.search(block_text)
     if m:
-        config["security_min_confidence"] = int(m.group(1))
+        value = int(m.group(1))
+        if value <= 100:
+            config["security_min_confidence"] = value
```
leehopper, line 463: **💡 [LOW] confidence_threshold/security_min_confidence out-of-range guard silently truncates instead of rejecting 4+ digit values**

The new `_REVIEW_CONFIDENCE_RE`/`_REVIEW_SECURITY_RE` patterns bound the digit capture to `[0-9]{1,3}` with no trailing digit boundary (e.g. `(?![0-9])`), so for a value with 4+ digits the regex silently captures only the first three and ignores the rest. The subsequent `value <= 100` check then operates on the truncated prefix, not the actual written value, so any config value whose first three digits are <=100 (the common case: any value starting with '100' followed by more digits, e.g. a typo like 1000 or 10000) is silently ACCEPTED as confidence_threshold/security_min_confidence=100 instead of being rejected as out-of-range and falling back to the default, exactly contradicting the comment's stated intent ('a value above 100 is ignored (defaults apply)'). The identical bug exists in both the Python and JS twins since they share the same registry-generated pattern.

**Suggested fix:**
Add a negative lookahead requiring no further digit after the captured run, e.g. `([0-9]{1,3})(?![0-9])`, in both the registry pattern (scripts/filter_patterns_registry.py PATTERN_FAMILIES rows for _REVIEW_CONFIDENCE_RE/_REVIEW_SECURITY_RE) and regenerate both twins via scripts/generate_filter_patterns.py, so a 4+ digit run fails to match entirely (falls back to default) rather than silently truncating to its first 3 digits.

scripts/filter_findings.py
```diff
@@ -960,8 +1338,21 @@ def _strip_suggested_fix_code_if_needed(finding, first_pattern_strip):
...
+        # the union whitespace class via `_WS_TRIM_RE.sub` (#244 (a), the shared
+        # union-trim constant), so the two twins anchor-strip the same six
+        # codepoints Python `str.strip()` and JS `trim()` disagreed on;
+        # `line_start` stays RAW in the signature, NOT routed through
+        # `_line_bucket` (mechanism (b) is a separate site).
+        sig = (
+            _WS_TRIM_RE.sub("", title.lower()),
+            filepath,
+            finding.get("line_start"),
+        )
```
leehopper, line 1355: **💡 [LOW] No regression test locks in the intentional unfolded-title dedup signature (#242/#244 interaction)**

The dedup-signature heuristic (#10) is deliberately built on the raw, unfolded title even though heuristic 7 (title placeholder match) now scans the confusable-folded title, and the code carries a long comment explaining this split is intentional so two fold-identical (homoglyph-disguised vs plain) titles still hash to distinct dedup signatures. No test in tests/test_filter_findings.py or workflows/test/filter_unit.test.js exercises this specific interaction: none construct two findings at the same file/line whose titles differ only by a homoglyph that folds to the same ASCII text and assert both survive apply_injection_filter (neither is treated as a duplicate of the other).

**Suggested fix:**
Add a parity fixture (or unit test in both twins) with two findings sharing file/line where one title is plain ASCII and the other is the same text with one casefold-reachable homoglyph substituted (e.g. 'Reviewer note' vs 'Revieweѕr note'), and assert apply_injection_filter/applyInjectionFilter keeps both (neither eliminated as a duplicate of the other) — this pins the h7-folded/h10-unfolded split the comment documents.

leehopper (commented):
---
Generated by code-gauntlet | Reviewed up to: 8966f08557c6b389a38bb598ed512ca60be2de79

scripts/filter_findings.py
leehopper: Fixed in cd6597b. Added `(?![0-9])` after the `([0-9]{1,3})` capture in both `_REVIEW_CONFIDENCE_RE` and `_REVIEW_SECURITY_RE` (registry, regenerated into both twins). Now a 4+ digit run fails to match entirely and falls back to default instead of truncating: measured in both twins, `1000`/`10000`/`100000` → default, `100`→100, `99`→99. Red-first parity fixtures `confidence_threshold_4digit_ignored` (1000) and `security_min_confidence_5digit_ignored` (10000) added, mutation-verified (reverting the lookahead reds them by re-parsing 100). Guard census unchanged.

scripts/filter_findings.py
leehopper: Fixed in cd6597b. Added parity fixture `injection/dedup_unfolded_title_split`: two findings at the same file+line whose titles differ only by a Cyrillic е (U+0435, folds to e) — asserted both kept, neither deduped, in both twins. Mutation-verified: folding the dedup signature title in either twin collides them and reds the fixture, pinning the h7-folded / h10-unfolded split the comment documents.

leehopper:
**Paired-mini measurement (post-merge, of record)** — run `mini-20260901-170531-2cb9104` at merged tip `2cb9104`, `--tier mini --child-auth api`, 6 pre-registered PRs.

| metric | this run | baseline of record (`custom-20260723-102149-381e9ff`) | Δ |
|---|---|---|---|
| golden recall | **0.6333** (19/30) | 0.6333 | +0.0000 |
| noise rate | **0.1786** | 0.2233 | −0.0447 |
| precision (strict) | 0.2533 | — | |
| f1 (strict) | 0.3619 | — | |

Recall is unchanged and noise is slightly lower — consistent with the design: the #242/#272 confusables+invisible fold and the #244 respells all run as a **raw-first-additive union scan**, which by construction can only add eliminations of injected/homoglyph findings and cannot remove a real one, so recall cannot drop. The offline noise battery (0 over-elimination across 50 injection fixtures + 42 legitimate multilingual/math/code findings) is corroborated here.

Caveats: N=1; the baseline of record predates ~2 months of intervening changes, so this is a *cumulative* 'no-regression' check, not a clean A/B of this PR alone (a same-session base leg would isolate it). One of the six PRs (cal.com#14740) was first invalidated by a stale-plugin-cache identity mismatch and re-run cleanly after renaming the cache aside; all 6 are valid in the scored run.

PR #285, opened 2026-09-04
feat(validate): structured reachability replaces the validator 65 cap (#115)

### LINKED ISSUE #115: Reconcile validator hypothetical cap with the 55 confidence threshold
## Origin
Filed from #37 inventory ID R-021 (v2→v3 residue audit).

## Problem
`skills/code-gauntlet/references/validation-pipeline.md:72` says hypothetical findings are capped at confidence 65, described as below the non-security threshold of 70. The live default non-security threshold is 55, also stated in the same file at line 82, so a 65-capped hypothetical remains above the filter bar and the stated suppression intent no longer holds.

## Why not fixed in #37
Changing only the prose accepts changed behavior; lowering the cap changes validator output and may affect recall. This needs a policy decision and #101 wave placement, not a Wave 2 suites-only assignment.

## Acceptance
- Decide whether hypothetical findings should survive the default non-security filter.
- Align the validator cap, live filter thresholds, and validation-pipeline documentation.
- Add tests covering the chosen boundary behavior.
- Place this issue in the appropriate #101 wave before implementation.

leehopper:
Replaces the validator's prose rule "cap confidence at 65 for hypothetical findings" with a structured `reachability` classification and a code-owned demotion step in the Filter stage. The cap was written when both thresholds were 70; since the non-security bar moved to 55, a 65-capped hypothetical survived non-security filtering at whatever severity discovery gave it and died only on security. Validators also applied the cap inconsistently (measured 55 and 60). The number is gone from every live contract; confidence is scored on the merits and reachability is reported as its own field.

## Mechanism

- The validator returns `reachability` on each validation item: `current` (the claim is true of the committed tree, dead code included when dead code is the claim), `future_change_only` (the failure needs a new caller, config value or code path not in this PR), or `uncertain`. Optional in `VALIDATE_SCHEMA` so an omitted field can never degrade a whole batch; the prompt's return line is derived from the schema.
- Both apply-validations twins copy the value only when it is one of the three strings.
- Both filter twins run `applyReachabilityDemotion` / `apply_reachability_demotion` first: a `future_change_only` finding is set to `severity: low` (with `original_severity`, `demoted_by: reachability`, `demotion_reason`). Nothing is eliminated. Because the step precedes the threshold step, a REVIEW.md `severity_threshold` above `low` suppresses these findings by the user's own rule.
- Routing sends demoted findings to `suggestion` (`routed_by: reachability`) ahead of dimension and agent routing, so `main_only` deliveries withhold them. Stats gain `reachability_demoted`. The report renders a Reachability bullet on demoted findings.
- Registry, checkpoint gate, challenge and contestation are untouched. `reachability` is validator metadata like `validation_justification`, not a discovery field.

## Why demotion, not elimination

Measured from `checkpoint-all` artifacts of the five most recent bench runs (validator justification regex, then read by hand). Four findings were genuinely future-only:

| run | pr | id | sev | discovery | validator | final | challenge | outcome |
|---|---|---|---|---|---|---|---|---|
| mini-20260818 | discourse-graphite-10 | type-2 | low | 68 | 55 | 65 | 80 | delivered |
| mini-20260818 | sentry-93824 | type-1 | high | 85 | 60 | 45 | 82 | delivered, adjudicated valid_extra |
| smoke-20260812 | cal.com-22345 | type-1 | low | 46 | 55 | 40 | 20 | challenge removed |
| smoke-20260816 | cal.com-22345 | bug-1 | low | 50 | 55 | 65 | 20 | challenge removed |

Elimination would have removed one valid-extra and one unadjudicated finding and no adjudicated noise. The sentry `simplify-1` dead-code finding (validator 75, challenge 92, "no callers anywhere") shows a literal "reachable today" boolean would misclassify a present defect, hence the three-value enum with dead code defined as `current`. The sentry type-1 row also shows the old cap delivering a hypothetical as HIGH; under this PR it is LOW and suggestion-routed.

Red-teamed by one Claude opus and one Codex sol pass. Both rejected the first draft (eliminate at Filter, keyed to the contestation drop that deleting the cap removes). Opus proposed routing hypotheticals to Challenge unconditionally; declined because every filter survivor already reaches Challenge, which arbitrated all four cases above.

## Paired mini watch items

Recall, noise and valid_extra against the e8b3af7 baseline; `stats.reachability_demoted` per PR; the sentry#93824 type-1 class should appear as LOW suggestion, not HIGH.

## Not in scope

Closing `VALIDATE_SCHEMA` (`additionalProperties`); the challenge cap's overflow behavior; `docs/v3-residue-audit-2026-07.md` stays as a dated record (its R-021 row cites `validation-pipeline.md:72`; the cap sentence sat at line 75 before this PR removed it).

Gates: CI green on e282efc (16 checks, CodeQL python still queued at ready time). Local: pytest 1977 passed, bench 554 passed, Node 1223 pass, parity goldens fresh, generators current, bundle byte-identical, pre-commit clean. Coverage from the #285 CI run: scripts 94.63, bench 88.29, JS 98.71 / 89.57 / 97.72; floors ratcheted to 93.7 and JS branches 88.6 (headroom over 1.0 pp). Review: two Codex luna lenses (mutation ledger, 11 mutations all red; spec conformance D1-D9) with zero blocker or major, one docstring minor fixed in 58ea8ab.

Closes #115

https://claude.ai/code/session_01C4ZcxgaK8oBKLdcFnuGs2n

---

> [!NOTE]
> **Medium Risk**
> Touches validator output shape and filter routing/demotion in both Python and bundled JS pipeline twins; behavior is heavily fixture-tested but changes what reviewers see for hypothetical findings.
> 
> **Overview**
> Replaces the validator’s informal **65 confidence cap** for hypothetical issues with an explicit **`reachability`** field (`current`, `future_change_only`, `uncertain`). Confidence is scored on merit; reachability is merged in **apply_validations** (Python and JS twins) and drives filter behavior.
> 
> **Filter stage:** `future_change_only` findings are **demoted to low severity** (with stamps like `demoted_by: reachability`), run **before** threshold filtering, **routed to suggestions** ahead of dimension/agent rules, and surfaced in reports with a Reachability bullet. Stats add `reachability_demoted`.
> 
> **Contracts and docs** update `agents/validator.md`, validation pipeline references, validate-stage schema/prompt (optional `reachability` in JSON schema), and delivery/report formatting. CI coverage floors ratchet to **93.7** (Python scripts) and **88.6** (JS branches); `AGENTS.md` shortens the floor history text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e282efcf6c69450b1015a74276d997ac8d4c5fd4. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/filter_findings.py
+++ b/scripts/filter_findings.py
@@ -46,6 +46,7 @@ Output JSON schema:
             "consensus_boosted":       N,   # confidence boosted for co-location (same file + 10-line bucket, any agent)
             "singleton_penalized":     N,   # singleton findings penalized -15 confidence (non-core dims)
             "dimension_routed":        N,   # findings routed to suggestion by dimension (BF-15a)
+            "reachability_demoted":    N,   # findings stamped by reachability demotion
             "cross_agent_consolidated": N,   # cross-agent findings stamped (never dropped) with consolidation_key
             "test_analyzer_promoted":  N,   # test-analyzer findings promoted to main report
             "tagged_main":             N,   # tagged for main report
@@ -72,6 +73,9 @@ Each filtered finding includes:
     # "suggested_fix_code" itself may be absent after filtering -- it is
     # deleted, not nulled, whenever suggested_fix_code_removed_by is present
     # (#63/D8, mirrors the suggestion contract above).
+    "original_severity": <severity before a reachability demotion, when changed>
+    "demoted_by": "reachability"  # present for future_change_only validator findings
+    "demotion_reason": "validator found no code path that reaches this issue without a future change"
 
 REVIEW.md parsing:
     Looks for a fenced code block or YAML-style section containing:
@@ -585,6 +589,40 @@ def apply_threshold_filter(findings, config):
         passed.append(finding)
 
     return passed, eliminated, contested_count
+
+
+# ---------------------------------------------------------------------------
+# Filter: validator reachability demotion
+# ---------------------------------------------------------------------------
+
+
+_REACHABILITY_DEMOTION_REASON = (
+    "validator found no code path that reaches this issue without a future change"
+)
+
+
+def apply_reachability_demotion(findings):
+    """Demote future-change-only findings to low severity without eliminating them.
+
+    Every matching finding receives the reachability stamps, including a finding
+    already at low severity. ``original_severity`` is recorded only when the
+    severity value changes.
+
+    Returns (findings, demoted_count); the input list and its findings are
+    mutated in place.
+    """
+    demoted_count = 0
+    for finding in findings:
+        if finding.get("reachability") != "future_change_only":
+            continue
+        if finding.get("severity") != "low":
+            if "severity" in finding:
+                finding["original_severity"] = finding["severity"]
+            finding["severity"] = "low"
+        finding["demoted_by"] = "reachability"
+        finding["demotion_reason"] = _REACHABILITY_DEMOTION_REASON
+        demoted_count += 1
+    return findings, demoted_count
 
 
 # ---------------------------------------------------------------------------
@@ -1988,13 +2026,16 @@ def tag_findings(findings):
     consolidation_key and one consolidation_primary, chosen by core-dimension,
     then confidence, then description length (see consolidate_cross_agent).
 
-    Step 2 — Dimension-based routing (BF-15a): Check the finding's dimension field
+    Step 2 — Reachability routing: a reachability-demoted finding always routes to
+    suggestion and receives ``routed_by="reachability"``.
+
+    Step 3 — Dimension-based routing (BF-15a): Check the finding's dimension field
     first. Dimensions like bug/security/cross_file_impact/intent always route to main.
     Dimensions like comment_accuracy always route to suggestion. Conditional dimensions
     (test_coverage, convention, type_design) route to suggestion unless functional-
     violation keywords are present.
 
-    Step 3 — Agent-based routing (fallback): If dimension routing returned None
+    Step 4 — Agent-based routing (fallback): If dimension routing returned None
     (unknown or missing dimension), fall back to agent-based rules:
 
       Agent routing:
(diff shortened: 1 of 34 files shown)


tests/test_filter_findings.py
```diff
@@ -5112,5 +5113,107 @@ def test_every_content_set_pattern_has_a_covering_synthetic(self):
...
+    def test_live_docs_do_not_retain_the_removed_confidence_cap(self):
+        for rel_path in (
+            "agents/validator.md",
+            "skills/code-gauntlet/references/validation-pipeline.md",
+        ):
+            text = (Path(__file__).parents[1] / rel_path).read_text()
+            self.assertIsNone(
+                re.search(r"\bcap(?:ped|ping|s)?\b[^\n.]{0,40}\b65\b", text, re.I),
+                rel_path,
+            )
```
leehopper: **💡 [LOW] New regression test guards prose docs with a wording regex, matching REVIEW.md's phrase-count-guard rule**

REVIEW.md's Rules section marks 'Never accept a phrase count as a guard... Flag any guard that asserts on the presence of a string where a structural property was available' CRITICAL. The new test_live_docs_do_not_retain_the_removed_confidence_cap asserts that agents/validator.md and validation-pipeline.md do not match the regex r"\bcap(?:ped|ping|s)?\b[^\n.]{0,40}\b65\b", which is a phrase-count guard defeated by rewording (e.g. 'ceiling of 65' or 'confidence tops out at 65' would silently evade it while reintroducing the same stale claim it exists to prevent).

**Suggested fix:**
Since the target is prose documentation with no structural representation, consider narrowing scope/expectations of this guard explicitly as a doc-drift heuristic rather than a hard regression test, or accept the residual risk explicitly in a comment — but per REVIEW.md's own CRITICAL rule this pattern should be flagged for author awareness.

**Cited rule:**
> CRITICAL: Never accept a phrase count as a guard. A check that greps for wording is defeated by rewording. Flag any guard that asserts on the presence of a string where a structural property was available. (REVIEW.md → Rules)

⚔️ *Code Gauntlet*

scripts/filter_findings.py
```diff
@@ -2038,42 +2079,46 @@ def tag_findings(findings):
...
+            # Step 3: Try dimension-based routing first (BF-15a)
+            dim_route = _route_by_dimension(finding)
+            if dim_route is not None:
+                destination = dim_route
+                if dim_route == "suggestion":
+                    finding["routed_by"] = "dimension"
+            else:
+                # Step 4: Fall back to agent-based routing
+                if agent in _MAIN_REPORT_AGENTS:
                     destination = "main"
```
leehopper, line 2096: **💡 [LOW] Security-dimension findings bypassing always-main routing via reachability demotion is untested**

tag_findings now checks demoted_by=='reachability' before dimension-based routing, so a security-dimension finding marked future_change_only is routed to suggestion instead of main even though _MAIN_DIMENSIONS always sends security there otherwise; no test exercises a security-dimension finding through this new precedence in either the Python or JS implementation.

**Suggested fix:**
Add a parity fixture / unit test (Python tests/test_filter_findings.py and workflows/test/filter_unit.test.js) with dimension='security' + demoted_by='reachability' asserting report_destination=='suggestion' and routed_by=='reachability', so a regression that special-cases security back to 'main' (or that fails to demote security findings the same way) is caught.

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet


---

### ⚠️ 2 finding(s) could not be anchored inline

2 inline comment(s) were posted; the following 2 finding(s) reference lines outside this diff and are included here instead: A finding listed here may not have an anchoring problem of its own — a consolidation group whose primary could not be anchored inline is listed here in full, corroborators included.

#### `scripts/filter_findings.py:2070`

**💡 [LOW] Inline 'Step 2 & 3' comment in tag_findings() was not updated for the new reachability-routing step**

The docstring above tag_findings() was updated by this diff to renumber the steps as 1=consolidation, 2=reachability routing, 3=dimension routing, 4=agent fallback, but the inline code comment right before the per-finding loop still reads '# Step 2 & 3: Dimension-based routing, then agent-based fallback', which now describes the wrong steps (the code immediately below it starts with a reachability check labeled '# Step 2' at line 2082, followed by dimension routing labeled '# Step 3' at line 2087).

**Suggested fix:**
Update the comment to '# Step 2, 3 & 4: Reachability routing, dimension-based routing, then agent-based fallback' to match the renumbered docstring and the code beneath it.

#### `scripts/filter_findings.py:2323`

**💡 [LOW] 'Pipeline:' summary comment omits the new reachability-demotion step that now runs first**

The banner comment above main()'s filter pipeline still reads '# Pipeline: threshold -> exclusions -> injection -> disagreement -> tag', but this diff inserts a new Step 1 (apply_reachability_demotion) that runs before the threshold filter, so the comment no longer lists the actual first stage of the pipeline it introduces.

**Suggested fix:**
Update the banner to '# Pipeline: reachability demotion -> threshold -> exclusions -> injection -> disagreement -> tag'.

---
Generated by code-gauntlet | Reviewed up to: e282efcf6c69450b1015a74276d997ac8d4c5fd4

tests/test_filter_findings.py
leehopper: Fixed in 5182eac81ba906a1467070a0542f5859617f52cd: the test is deleted rather than reworded. Prose has no structural property to assert on, and a wording regex is exactly the phrase guard the rules forbid; git history records the cap's removal.

scripts/filter_findings.py
leehopper: Fixed in 5182eac81ba906a1467070a0542f5859617f52cd: security-dimension + demoted_by=reachability case added to tests/test_filter_findings.py, workflows/test/filter_unit.test.js and a new parity fixture (reachability_routing/demoted_security_precedes_main_dimension), asserting suggestion + routed_by=reachability in both runtimes.

leehopper:
Gauntlet fix map (5182eac81ba906a1467070a0542f5859617f52cd): inline LOW 1 (regex doc guard) -> test deleted, phrase guard on prose; inline LOW 2 (security precedence untested) -> tests + parity fixture in both runtimes; unanchored LOW 3 (Step 2 & 3 comment) and LOW 4 (Pipeline banner) -> comments corrected in scripts/filter_findings.py.

PR #287, opened 2026-09-04
fix(verify): hand the slice-input to the executor inline, off the artifact-writer (#275)

### LINKED ISSUE #275: Verify slice-input still transits the artifact-writer; writer paraphrases quoted code, degrading whole slices to origin=unknown (2/6 PRs in the v3.24 mini)
Found while closing out PR #273's paired mini (`mini-20260901-170531-2cb9104`); out of that PR's scope, filing per standing practice.

## Problem

The verify stage's slice-input document is still written to disk by the `artifact-writer` subagent (`materializeVerifySlices`), the one artifact channel that did not move to the harness-serialized workflow return when #53 closed. The writer is a model asked to copy escape-dense JSON verbatim, and it does not: the slice-input content proof (#25/#69) caught it on 2 of 6 PRs in this run, and on 1 of 6 in the previous mini (`mini-20260818-120540-b423885`, same PR: sentry#93824).

Each hit degrades the whole slice: every finding is stamped `origin=unknown` and surfaced-classification is skipped (30/30 on cal.com#11059, 18/18 on sentry#93824). The findings still ship, so recall is unaffected, but the pre-existing-vs-new classification is lost for the whole PR.

## Evidence (on-disk slice-input vs the pipeline's in-memory copy, all other fields byte-equal)

- cal.com#11059, `bug-2.description` @ char 404: disk has `a meaningless "[object Object]" optional field`; dispatched has `` a meaningless `"[object Object]"` optional field `` (writer dropped the backticks around a quoted string).
- sentry#93824, `bug-1.evidence` @ char 53: disk has real newlines where the dispatched value carried literal `\n` escapes inside a code quote (writer un-escaped the backslash sequences).

Key order on disk matches `VERIFY_SLICE_FIELDS` exactly in all six slices, so the #172 sorted-key mechanism is not the cause; that issue's revisit trigger was evaluated and not met (mismatch is content, not ordering).

The two failing PRs are the two with the most quote/escape-dense evidence (62 and 36 escape sequences in the slice document). The four clean PRs range 24–88, so density alone does not predict it; the writer paraphrasing quoted code is the failure class.

## Fix shape

Move the slice-input write onto a channel that does not transcribe: the workflow cannot write to disk, but the executor already receives the dispatched content in its prompt and computes the receipt over what it parsed. Options, in order of preference:

1. Have the executor (Bash-capable) write the slice-input itself from a payload the workflow hands it inline, then verify against that write. Removes the writer from the verify path entirely.
2. Dispatch the slice content inline to `verify_findings.py` via stdin/argv-safe encoding (base64 of the JSON) so no model ever re-serializes it.

Either way the content proof stays as the belt.

## Watch item

`stats.inputProof.mismatched > 0` in any run envelope. Expect it on sentry#93824 in every mini until fixed.

leehopper:
Closes #275. Wave 5 item on #101; suites-only, no paired mini on this PR.

## Problem

The verify stage's slice-input document was the one artifact still written to disk by the `artifact-writer` subagent. A model asked to copy escape-dense JSON verbatim paraphrases it (dropped backticks, un-escaped `\n`, invalid `\escape`), and the content proof then degrades the whole slice to `origin=unknown`. Two of six PRs in the v3.24 mini and one of six in the v3.12 mini hit it.

The 2026-09-03 mini (`mini-20260903-204424-e8b3af7`) was reported clean on `inputProof.mismatched == 0`, but it was not clean: discourse-graphite-4's on-disk slice input is invalid JSON (`Invalid \escape` at char 2578), both attempts re-read the same file, and 22/22 findings went `origin=unknown` under `status=failed`. The `mismatched` counter never sees the parse-failure class. With the default `verifySliceSize` of 200, every real run was one slice, so one bad copy cost the whole PR its classification.

## Why this shape

There is no non-transcribing channel mid-workflow. The sandbox has no disk, so every byte that reaches disk before the workflow return passes through one model tool call. The fix therefore removes the paraphrase class from what the model copies, makes the retry a fresh transcription, and lets code write the file from what the model carried.

Measured this session:

- Base64 is dead on cost. A 56,588-char base64 blob tokenized at 54,441 tokens, about one token per character and five times the JSON it encodes.
- Percent-encoding string values and keys outside a shell-inert alphabet (`A-Za-z0-9 .,:/_-`) inflates the ten retained mini slice documents by 2 to 4 percent. The result carries no backslash, quote, backtick, control character or non-ASCII byte, so nothing is left for a model to normalize.
- Four sonnet executors copied encoded payloads into a single-quoted argv token in one Bash call: two at 43,636 chars (38 findings, cal.com#11059) and two at 51,744 chars (48 findings). All four decoded documents were byte-identical to the source. Above 51,744 chars is unmeasured: the Read tool caps at 25k tokens, so a 94k probe could not be fed to the executor by file, and the one attempt that stitched partial reads is discarded as invalid.

The earlier "re-encoding was measured and does not work" note in `workflows/AGENTS.md` covered PR #92's respelling of JSON backslash runs on the persist wire, still inside JSON escape semantics. It is now scoped to that measurement, with the verify exception stated next to it.

## Mechanism

**JS (`workflows/src/stages.js`, `args.js`).** `encodeInlineString` / `encodeSliceInline` percent-encode every string value and object key (uppercase `%XX` over hand-rolled UTF-8, `%uXXXX` for a lone surrogate so the value round-trips exactly), then `JSON.stringify` compactly; a postcondition pins the output to printable ASCII without an apostrophe, so `shellWord` emits it as one POSIX `raw_string` token. `planVerifySlices` is the only slicer: greedy by the count cap (`verifySliceSize`, default 200 to 25) and by `VERIFY_INLINE_CHAR_BUDGET` = 50,000 chars, pinned just under the measurement; a single finding over the budget is degraded alone. `worstCaseAgentCount` and `coarsenLimits` take the findings array post-merge and plan with the same planner and the same base branch, so the agent guard counts exactly what verify dispatches; a coarsening term whose doubling stops reducing it leaves the candidate set, so the loop terminates when verify is size-bound. `trustSlice`'s input-proof mismatch is no longer "deterministic, not retried": a re-dispatch is a fresh transcription. `materializeVerifySlices`, the writer prompt, `chunkBySerializedSize`, `PROMPT_SEGMENT_CHAR_BUDGET`, the verify "(no write proof)" reason, `input_recovery`, the RECOVERED gap and the `recovered` counter are deleted from the verify path; `inputProof` gains `oversize`, `retried`, `retriedMismatch` and `retriedMissing`.

**Python (`scripts/verify_findings.py`).** Receipt mode has one live source: `--input-inline` is required with `--input`, and `--input` is now the code-written destination. `decode_inline_slice` parses strictly (no backslash anywhere, no duplicate keys, no `NaN`/`Infinity`), then decodes keys and values in canonical form only: uppercase hex, no `%41`-style escape of a safe byte, strict UTF-8, `%uXXXX` only in the surrogate range. Rejections name the code point and the JSON path. The decoded document is checksummed before numeric coercion (the unchanged value proof, same parity fixtures), written atomically to `--input` as the canonical pretty form so the file's fnv1a32 equals the receipt's `input_checksum`, and verified as before. The receipt envelope and the unspellable-numeric fallback serialize lone surrogates safely. The positional CLI path and its trailing-bytes recovery are unchanged.

**Contracts and docs.** `agents/executor.md` describes the quoted `--input-inline` token; `agents/artifact-writer.md` loses the slice-input payload shape; `skills/code-gauntlet/SKILL.md`, its references, `workflows/AGENTS.md` (and its twin), `docs/machine-parsed-strings.md`, `bench/MEASUREMENT.md` and `bench/PROFILING.md` describe the inline channel. `bench/profile_run.py` and `bench/runner/check.py` are untouched by design: the profiler reads historical journals that carry `verify-input-writer-*` labels, and the G3 sentinel still matches the persist path.

**Tests.** New parity group `tests/fixtures/parity/slice_inline` (10 cases, JS-recorded goldens: astral, controls, leading/embedded/trailing lone surrogates, surrogate pair, non-ASCII keys, `%` lookalikes, safe punctuation, three findings) checked by both runtimes; encoder class pins; no writer dispatch in `verifyStage`; retry-on-mismatch; oversize isolation; planner-in-guard incl. a 1,000-char base branch with 800 findings; coarsening termination; decoder rejection matrix; receipt round trips incl. lone surrogates with `1e21`; argparse pair enforcement; atomic destination write. Every mechanism test was verified red by whole-mechanism mutation with byte-exact restore (ledgers in the review record).

## Watch item for the wave-end paired mini

`inputProof.proven == inputProof.slices` on all six PRs, zero `verify: UNVERIFIED` gaps, and `inputProof.retried == 0`. Not `mismatched == 0`, which is blind to the parse-failure class above.

## Review record

Per the engine-allocation rule. Recon, red-team, build, and review rounds 1 and 2 ran on Codex only (the owner paused Anthropic subagent spend mid-PR); once the usage window reset, a Claude opus lens and four sonnet copy probes were added. The session adjudicated every verdict, and the owner's code-gauntlet run is the cross-family pass.

- Recon: Codex luna, read-only inventory of every coupling (326 lines).
- Design red-team: Codex sol at `max`, verdict REVISE, 1 blocker + 11 major. Adopted: keys encoded too, shell-inert alphabet (no `;()`), `--input-inline` mandatory, budget pinned to the measurement (43,000, not the memo's 60,000), canonical-form decoder, transport parity goldens, atomic write, `retried` counter, planner feeding the guard with a termination rule. Rejected: "checkpoint replay bypasses the transport" (a completed verify phase is replayed as results and needs no transport), and cross-slice duplicate-id concern (nothing is dropped; the join is per slice by design). The blocker (strict UTF-8 vs lone surrogates) was already covered by `%uXXXX` in the spec.
- Build: Codex luna `xhigh` from the spec; 6-entry mutation ledger.
- Round 1: two Codex luna lenses (mutation ledger, spec conformance), both FIX. Blockers: lone surrogate crashed receipt-mode output serialization; the guard planned with the default branch, so a 1,000-char base branch undercounted 800 slices as 400. Plus a missing red test for mandatory `--input-inline`, safe-alphabet and diagnostic-path test gaps, two stale writer sentences, a fixture gap, a stale comment; the session added the atomic write and a control-flow cleanup. All eleven fixed in commit c567274.
- Round 2: conformance lens PASS (all eleven fixes present; one nit, a 997-char test branch labelled 1,000). Mutation lens FIX: four surviving mutations, all test gaps (raw duplicate keys, `NaN`/`Infinity`, exact destination bytes, checksum-before-coercion). Its pre-commit "blocker" was the sandbox's read-only cache; the hook run passes in the session.
- Claude opus lens (the Claude-family read, added once the usage window reset): FIX with 0 blockers, 3 majors, 8 minors. Majors: the `verify_fanout:` disclosure fired on ordinary runs at the new default and advised raising a knob the budget overrides; the on-disk destination bytes were not pinned to the receipt checksum; the receipt path's shape validation call site had no test. Minors were test gaps (backslash guard, planner flush before an oversize finding, `slices` excluding oversize, encoder postcondition via `toJSON`, first-attempt faults invisible in the ledger), one stale comment, one unscoped re-encoding sentence in `phase3-dispatch.md`, and an empty-string base-branch normalization mismatch. Its 25-class encoder/decoder round trip through real Node and Python found zero rejections.
- Round 3 (fix round 2): all of the above plus the budget raise, in commit adfe664. A Codex luna confirmation lens re-verified every fix and re-ran all 11 previously surviving mutations: 11 red, 11 restored byte-exact (PASS, see comment).
- Cursor Bugbot (adfe664): one medium finding, the planner charged the slice envelope once per finding and once more per slice, splitting a fitting slice early. Fixed in c0ed3ad with an exact-fit test and a predicted-equals-encoded property test; 38b2729 drops a leftover local Biome flagged.
- Owner's code-gauntlet run (38b2729): 4 MEDIUM + 4 LOW (doc misattribution, a surrogate-in-deltas test gap, two readability items, oversize-boundary misclassification in the fan-out disclosure, an unmeasured comment figure, duplicated validator branches, a stale literal count in check.py). All fixed in 2e97281 with replies on each thread; the planner now tags each slice's close reason.

## Verification

Local at 2e97281: `node --test` 1244 pass, `pytest tests/` 2005 pass, `pytest bench/tests/` 557 pass, bundle fresh, `record_parity.py --check` fresh, `sync_agent_rules.py --check` current, `pre-commit run --all-files` clean. Coverage measured locally by the fix round: scripts 94.62 (floor 93.7), bench 88.63 (floor 87.5), JS 98.78 / 90.02 / 98.20 (floors 98.2 / 89.3 / 97.6). CI at adfe664 (run 33928618212): scripts 94.58, bench 88.31, JS 98.78 / 90.06 / 98.20. Every gate sits under 1.0 pp above its floor, so no floor is ratcheted in this PR. All 17 checks green at 2e97281.

## Not done

- No probe above 43,636 chars; the budget is pinned to what was measured.
- No live shell-mutation probe of a "restored" apostrophe; the alphabet is designed so a broken quote yields no operator or expansion, and the script rejects the mangled argv, but the sandbox's current auto-approval behaviour on that AST is unmeasured.
- Profiler code untouched (historical journals); only its docs changed.

https://claude.ai/code/session_01PQH5zkqsNswbjXKHkAsC6e

---

> [!NOTE]
> **Medium Risk**
> Changes the Verify-stage trust boundary and how findings get `origin` classification; mis-encoded or transcribed inline tokens can still degrade whole slices, though behavior is heavily tested and documented.
> 
> **Overview**
> **Verify no longer materializes slice JSON through the artifact-writer.** Each projected slice is percent-encoded and dispatched as a single quoted `--input-inline` token (under `VERIFY_INLINE_CHAR_BUDGET`), which the executor must reproduce verbatim in one Bash call; `verify_findings.py` strictly decodes it, checksums the document, atomically writes `--input`, then runs receipt-mode verification as before.
> 
> **Contract and envelope changes:** Agent docs drop the `{ path, content }` writer shape and `input_recovery`; the executor echo is `status`, `receipt` (including `input_checksum`), and `result.deltas` only. Skill, workflow, and bench docs now describe inline transport and retire the verify-input-writer stage from profiling expectations.
> 
> **Python receipt path:** New `decode_inline_slice` with canonical percent-encoding rules; receipt envelopes no longer emit `input_recovery` (trailing-byte recovery stays on the legacy positional CLI). Tests add JS/Python parity fixtures under `slice_inline` and expand receipt/decoder coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e97281ef2d1c3610a30cd635e8514d460b4fea1. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:
## Confirmation lens on adfe664 (Codex luna): PASS

All 16 round-2 fix items present with their named tests. Every previously surviving mutation was re-run: 11 requested, 11 red, 11 restored byte-exact (the four round-2 mutation-lens survivors and the seven the opus lens verified). Gates green in the worktree; the live-read scan found no writer-era slice-input sentence and no stale 43k budget spelling.

CI at adfe664: scripts 94.58, bench 88.31, JS 98.78 / 90.06 / 98.20. No gate has 1.0 pp of headroom over its floor, so no ratchet. Bugbot skipped this run.

Review-ready for the code-gauntlet pass and owner sign-off.

workflows/src/stages.js
```diff
@@ -157,25 +239,45 @@ export function sharedContextLine(inp) {
...
+      oversize.push(finding);
+      continue;
     }
-    cur.push(it);
-    curSize += size;
+    if (current.length > 0 && (current.length >= maxFindings || currentCost + cost + envelopeAllowance > maxChars)) {
+      flush();
+    }
+    current.push(finding);
+    currentCost += cost;
```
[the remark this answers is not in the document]
leehopper: Fixed in c0ed3ad: the planner now prices each projected finding alone and predicts the slice as envelope + findings + commas, so the predicted length equals the real encoded length exactly (property test over random inputs and branch lengths, plus an exact-fit test that the old double charge fails).

leehopper:
bugbot run

bench/MEASUREMENT.md
```diff
@@ -127,18 +127,18 @@ The prose identity receipt symptom belongs to **G4**, not this floor: G4 checks
...
 
-**`origin=unknown` from artifact-writer transcription drift.** The writer is a
-sampled agent, not a function (see CLAUDE.md, "The by-value writer is not
-trustworthy"), and a corrupted slice-input file makes `verify_findings.py`
-refuse the slice, degrading every finding in it to `origin=unknown` and
-tripping gate 3. Measured on two consecutive smokes:
+**Current verify input transport.** Verify slices now travel to the executor as a
+percent-encoded `--input-inline` token planned under `VERIFY_INLINE_CHAR_BUDGET`;
+the executor decodes the token and writes the destination before running
+`verify_findings.py`. The content proof remains as a belt. The artifact-writer
```
leehopper: **🟡 [MEDIUM] MEASUREMENT.md misattributes the inline decode/write step to the executor, not the script**

The new prose says 'the executor decodes the token and writes the destination before running verify_findings.py', but every other doc this PR touches (workflows/AGENTS.md, workflows/CLAUDE.md, agents/executor.md, phase2-triage.md) says the opposite: the executor only reproduces the quoted --input-inline token verbatim in one Bash call, and it is verify_findings.py itself (invoked in that same call) that decodes the token and writes the destination before verification runs.

**Suggested fix:**
Reword to match the other docs, e.g. 'the executor reproduces the token exactly in one Bash call; the script decodes it and writes the destination before running verification.'

⚔️ *Code Gauntlet*

scripts/verify_findings.py
```diff
@@ -1181,13 +1327,19 @@ def load_input(findings_json_path):
     return data, recovery, checksum
 
 
-def _write_output(output, output_path):
-    """Write output JSON to file or stdout."""
+def _write_output(output, output_path, *, receipt=False):
```
leehopper, line 1333: **🟡 [MEDIUM] No test exercises the envelope-escaping fix for a lone surrogate reaching result.deltas**

The receipt path now writes its --output envelope with ensure_ascii=receipt (True) specifically because a decoded inline slice may legally carry a lone UTF-16 surrogate anywhere in a finding, including the id field, which flows verbatim into build_deltas()'s delta['id']. Every surrogate-focused test in this diff (test_receipt_round_trips_lone_surrogates_and_escapes_receipt_output, the astral/lone_surrogates/surrogate_pair parity fixtures) puts the surrogate in a finding's description/evidence field, which never reaches the deltas array, so none of them exercise the envelope write at scripts/verify_findings.py:1627 with a surrogate actually present in the written JSON.

**Suggested fix:**
Add a test that runs receipt mode over a finding whose id (or another _DELTA_FIELDS value) contains a lone surrogate via the inline encoder, and assert the --output envelope is written successfully with the surrogate escaped as \uD800-style JSON (not raising UnicodeEncodeError), so a future regression reverting `ensure_ascii=receipt` to False is caught by a failing test.

**Cited rule:**
> AGENTS.md: "A regression test must fail against the bug it names" — mutating the ensure_ascii=receipt behavior back to False should turn a test red, but no test currently reaches that code path with a surrogate-bearing delta value.

⚔️ *Code Gauntlet*

scripts/verify_findings.py
```diff
@@ -1107,12 +1106,164 @@ def _input_checksum(doc):
...
+            object_pairs_hook=_inline_pairs,
+            parse_constant=lambda value: (_ for _ in ()).throw(
+                ValueError(f"non-finite JSON constant {value}")
+            ),
+        )
+    except (json.JSONDecodeError, ValueError) as exc:
+        _inline_reject(f"invalid JSON at $ ({exc})")
+    if not isinstance(parsed, _InlineObject):
+        _inline_reject("root must be an object at $")
+    return _decode_inline_node(parsed)
```
leehopper, line 1247: **🟡 [MEDIUM] Generator-throw idiom for parse_constant is a clever one-liner that obscures intent**

decode_inline_slice passes json.loads a parse_constant lambda written as `lambda value: (_ for _ in ()).throw(ValueError(...))` — the well-known but obscure trick for raising inside a lambda via an empty generator's .throw(). Since a lambda cannot contain a raise statement, this is the workaround, but it forces any future reader to recognize the idiom before understanding that this simply rejects non-finite JSON constants (NaN/Infinity/-Infinity). A small nested named function that does `raise ValueError(...)` says the same thing directly.

**Suggested fix:**
Define a small nested function, e.g. `def _reject_non_finite_constant(value): raise ValueError(f"non-finite JSON constant {value}")`, and pass it as `parse_constant=_reject_non_finite_constant`. Same ValueError, same message, caught by the same except clause below.

```suggestion
def decode_inline_slice(text):
    """Strictly decode the percent-encoded JSON document carried by the executor."""
    if not isinstance(text, str):
        _inline_reject("payload is not text at $")
    if "\\" in text:
        _inline_reject(
            "JSON escape sequences are not canonical at $ (offending U+005C)"
        )

    def _reject_non_finite_constant(value):
        raise ValueError(f"non-finite JSON constant {value}")

    try:
        parsed = json.loads(
            text,
            object_pairs_hook=_inline_pairs,
            parse_constant=_reject_non_finite_constant,
        )
    except (json.JSONDecodeError, ValueError) as exc:
        _inline_reject(f"invalid JSON at $ ({exc})")
    if not isinstance(parsed, _InlineObject):
        _inline_reject("root must be an object at $")
    return _decode_inline_node(parsed)
```

⚔️ *Code Gauntlet*

workflows/src/stages.js
```diff
@@ -889,56 +998,93 @@ export async function verifyStage(ctx, input) {
...
   // Degrade-and-disclose (issue #72), not abort: a small verifySliceSize is a legitimate
   // transcription-fidelity mitigation, not a mistake to reject, but its dispatch cost is
   // real and otherwise invisible until the run runs long or worstCaseAgentCount rejects it
   // outright. Collected here rather than pushed straight into `gaps` so it lands ahead of
   // any per-slice degrade gap, in the order this stage discovers information.
+  const fanoutAdvice = fanoutBounds.budgetBound
+    ? 'The inline character budget bound this split; raising verifySliceSize will not reduce this fan-out.'
+    : fanoutBounds.countBound
+      ? 'Raise verifySliceSize to reduce fan-out.'
+      : 'The split bound could not be classified; inspect the effective slice size and inline budget.';
```
leehopper: **🟡 [MEDIUM] Nested ternary chain for fan-out advice message**

The new fanoutAdvice assignment nests a second ternary inside the alternate branch of the first to pick among three advice strings. Before: fanoutBounds.budgetBound ? A : fanoutBounds.countBound ? B : C. A reader has to parse two levels of ?: to find the three outcomes, which is exactly the readability trap flat if/else chains avoid. After: an if/else-if/else assigning to a `let fanoutAdvice` reads top-to-bottom with no nesting.

**Suggested fix:**
Replace the nested ternary with an if/else-if/else chain assigning to a `let fanoutAdvice`, one branch per case (budgetBound, countBound, unclassified). Same three strings, same selection logic, no nesting.

```suggestion
  let fanoutAdvice;
  if (fanoutBounds.budgetBound) {
    fanoutAdvice = 'The inline character budget bound this split; raising verifySliceSize will not reduce this fan-out.';
  } else if (fanoutBounds.countBound) {
    fanoutAdvice = 'Raise verifySliceSize to reduce fan-out.';
  } else {
    fanoutAdvice = 'The split bound could not be classified; inspect the effective slice size and inline budget.';
  }
```

⚔️ *Code Gauntlet*

workflows/src/stages.js
```diff
@@ -889,56 +998,93 @@ export async function verifyStage(ctx, input) {
...
+    for (let i = 0; i + 1 < slices.length; i += 1) {
+      const current = slices[i];
+      const next = slices[i + 1][0];
+      const currentCost = predictVerifySliceInlineLength(current, baseBranch);
+      const nextCost = projectedVerifyFindingInlineLength(next);
+      if (current.length >= maxFindings) countBound = true;
+      else if (currentCost + nextCost + 1 > maxChars) budgetBound = true;
+    }
+    return { countBound, budgetBound };
+  })();
```
leehopper: **💡 [LOW] Fan-out disclosure misclassifies the split cause when an oversize finding sits between two slices**

fanoutBounds assumes slices[i] and slices[i+1] are adjacent in the original findings order and computes whether adding slices[i+1][0] to slices[i] would have exceeded the count/budget bound. When planVerifySlices closed slices[i] because the NEXT finding was oversize (and therefore skipped into plan.oversize rather than merged), slices[i+1][0] is not the finding that actually caused the flush, so neither countBound nor budgetBound is set for that boundary. This can make the verify_fanout gap wrongly advise 'Raise verifySliceSize to reduce fan-out' (when the real cause is unrelated oversize findings) or report 'could not be classified' instead of naming the true cause.

**Suggested fix:**
Track the actual close reason inside planVerifySlices itself (e.g. tag each closed slice with 'count' | 'budget' | 'oversize') and surface it directly, rather than reconstructing it afterwards from slices[i+1][0], which can be a finding from far later in the array once an oversize entry has been skipped in between.

⚔️ *Code Gauntlet*

workflows/src/args.js
```diff
@@ -140,7 +140,8 @@ export const LIMIT_DEFAULTS = {
   summarizeBucketSize: 20,
   validateBatch: 25,
   challengeCap: 40,
-  verifySliceSize: 200,
+  // Keep a bad inline transcription's blast radius small; a typical slice is about 30k chars.
```
leehopper: **💡 [LOW] Unmeasured 'typical slice' claim added next to the new verifySliceSize default**

REVIEW.md's rules explicitly require flagging 'a comment or doc that asserts a behaviour, frequency, or limit without saying how it was established.' The new comment states 'a typical slice is about 30k chars' with no measurement citation, unlike the neighboring VERIFY_INLINE_CHAR_BUDGET comment in the same PR which cites exact measured figures (43,636/51,744 chars, '2/2 sonnet executors').

**Suggested fix:**
Either cite the measurement/derivation behind the '30k chars' estimate or drop the specific figure and just state the rationale (small blast radius) without an unmeasured number.

**Cited rule:**
> REVIEW.md Rules: "Claims must name their measurement. Flag a comment or doc that asserts a behaviour, frequency, or limit without saying how it was established."

⚔️ *Code Gauntlet*

scripts/verify_findings.py
```diff
@@ -1107,12 +1106,164 @@ def _input_checksum(doc):
...
+        if inline:
+            reject("missing required 'findings' array at $")
+        else:
+            reject("Input JSON is missing required 'findings' array.")
+    if not isinstance(data["findings"], list):
+        if inline:
+            reject("'findings' must be an array at $")
+        else:
+            reject("'findings' must be an array.")
+    return data
```
leehopper, line 1268: **💡 [LOW] Three repeated if-inline/else blocks in _validate_input_shape obscure the actual checks**

_validate_input_shape repeats the same `if inline: reject(A) else: reject(B)` shape three times, once per structural check. The repetition buries the three actual validations (dict root, findings key present, findings is a list) under six lines of branching that all do the same dispatch. A tiny local `fail(inline_msg, legacy_msg)` helper collapses each check to one line while keeping every message byte-identical.

**Suggested fix:**
Introduce a local `fail(inline_msg, legacy_msg)` that calls `reject(inline_msg if inline else legacy_msg)`, and replace each three-line if/else with one `fail(...)` call carrying both message variants.

```suggestion
def _validate_input_shape(data, inline=False):
    reject = _inline_reject if inline else die

    def fail(inline_msg, legacy_msg):
        reject(inline_msg if inline else legacy_msg)

    if not isinstance(data, dict):
        fail("root must be an object with a 'findings' key at $", "Input JSON must be an object with a 'findings' key.")
    if "findings" not in data:
        fail("missing required 'findings' array at $", "Input JSON is missing required 'findings' array.")
    if not isinstance(data["findings"], list):
        fail("'findings' must be an array at $", "'findings' must be an array.")
    return data
```

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet


---

7 inline comment(s) were posted; the following 1 finding(s) reference lines outside this diff and are included here instead: A finding listed here may not have an anchoring problem of its own — a consolidation group whose primary could not be anchored inline is listed here in full, corroborators included.

#### `workflows/src/stages.js:2683`

**💡 [LOW] Removing the verify slice-input writer's failure string drops bench/runner/check.py's cited literal count from 4 to 3**

This diff deletes materializeVerifySlices, which used to contain the third-to-last string literal reading '...(no write proof)'. bench/runner/check.py:89 documents its G3 degrade-sentinel regex by asserting there are exactly 4 string/template literals spelling that phrase in the pipeline bundle; after this diff only 3 remain (writeArtifacts partial-artifacts gap, the primaries write-proof fail, and the assemble receipt write-proof check), so that count comment is now wrong.

**Suggested fix:**
Update the comment in bench/runner/check.py (around line 89) to say 'x3' (or recount and describe the remaining literals precisely) so the audit note stays accurate for anyone re-verifying G3's regex coverage.

---
Generated by code-gauntlet | Reviewed up to: 38b27295de9ac507873b56939b685e729947620f

bench/MEASUREMENT.md
leehopper: Fixed in 2e97281: Reworded: the executor reproduces the token in one Bash call; verify_findings.py decodes it and writes the destination before verification.

scripts/verify_findings.py
leehopper: Fixed in 2e97281: Added a receipt-mode test with a lone surrogate in a finding id that reaches result.deltas; the envelope writes with the surrogate JSON-escaped, and reverting ensure_ascii for the receipt path goes red.

scripts/verify_findings.py
leehopper: Fixed in 2e97281: Replaced the generator-throw lambda with a named nested rejector; same ValueError and message.

workflows/src/stages.js
leehopper: Fixed in 2e97281: Replaced the nested ternary with a flat if / else if / else.

workflows/src/stages.js
leehopper: Fixed in 2e97281: planVerifySlices now tags each closed slice with its close reason (count, budget, oversize) and the disclosure reads the tags; an oversize-caused boundary no longer advises raising verifySliceSize (regression test, mutation-verified).

workflows/src/args.js
leehopper: Fixed in 2e97281: Replaced the unmeasured figure with its derivation: the ten retained mini slice documents ran 15.7k to 42.4k chars for 14 to 38 findings, about 1.1k chars per finding, so 25 findings is about 28k.

scripts/verify_findings.py
leehopper: Fixed in 2e97281: Collapsed the three branches with a local fail(inline_msg, legacy_msg) helper; messages byte-identical.

workflows/src/stages.js
```diff
@@ -889,56 +998,94 @@ export async function verifyStage(ctx, input) {
...
+  const slices = plan.slices;
+
+  // Recompute only the boundaries that closed a planned slice. The public planner result
+  // stays the compact `{ slices, oversize }` contract, while this tells the disclosure which
+  // exact cost bound caused the fan-out. The cost is the empty envelope plus projected
+  // finding-object lengths and one comma per additional finding.
+  const fanoutBounds = (() => {
+    const maxFindings = Math.max(1, sliceSize || findings.length || 1);
+    const maxChars = Math.max(1, VERIFY_INLINE_CHAR_BUDGET);
+    const envelopeAllowance = encodeSliceInline({ findings: [], base_branch: baseBranch }).length;
```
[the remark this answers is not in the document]
leehopper: Fixed in 2e97281: Removed in 38b2729.

workflows/pipeline.js
```diff
@@ -5544,56 +5654,94 @@ async function verifyStage(ctx, input) {
...
+  const slices = plan.slices;
+
+  // Recompute only the boundaries that closed a planned slice. The public planner result
+  // stays the compact `{ slices, oversize }` contract, while this tells the disclosure which
+  // exact cost bound caused the fan-out. The cost is the empty envelope plus projected
+  // finding-object lengths and one comma per additional finding.
+  const fanoutBounds = (() => {
+    const maxFindings = Math.max(1, sliceSize || findings.length || 1);
+    const maxChars = Math.max(1, VERIFY_INLINE_CHAR_BUDGET);
+    const envelopeAllowance = encodeSliceInline({ findings: [], base_branch: baseBranch }).length;
```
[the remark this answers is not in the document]
leehopper: Fixed in 2e97281: Removed in 38b2729 (bundle rebuilt).

leehopper:
Unanchored gauntlet finding (check.py literal count): corrected in 2e97281; the comment now counts the three remaining literals and names them.

leehopper:
bugbot run

PR #288, opened 2026-09-05
feat(filter): scope REVIEW.md thresholds and ignores per subtree (#196)

### LINKED ISSUE #196: resolveReviewConfig applies the merged REVIEW.md config globally; per-subtree scoping from review-md-spec.md's inheritance examples is not implemented
Surfaced by the adversarial review of PR #194 (issue #24 sub-PR 2, in-pipeline REVIEW.md parse).

`references/review-md-spec.md`'s inheritance section describes per-subtree scoping ("For a file in legacy/: confidence_threshold = 70 (root applies)") — thresholds scoped to the directory subtree whose REVIEW.md set them. `resolveReviewConfig` (workflows/src/args.js) instead merges all discovered REVIEW.md entries into ONE flat config (root-first depth order, deeper wins, ignores accumulate) applied to every finding, so a subdirectory REVIEW.md's threshold governs findings in unrelated directories.

This flattening pre-dates #24 (the orchestrator's hand-merge produced one object too) so it is not a regression, but #24/#80 widened the discovered set (changed-file dirs + ancestors), folding more subdirectory files into that one object. PR #194 narrowed the doc claims honestly (SKILL.md states the flat-config behavior; review-md-spec.md carries an implementation note), so this issue tracks the real design question: implement per-subtree scoping (config resolved per finding path), or ratify the flat merge and rewrite the spec's inheritance examples to match.

Recall/noise-sensitive if implemented (changes which findings are filtered) — Wave 5+ material, not a quick fix. Zero REVIEW.md files exist in the bench mirror set today, so the behavior is latent either way.

leehopper:
Closes #196. Wave 5 item on #101; suites-only, no paired mini on this PR (owner ruling).

## Problem

The old `reviewMd` merge sorted discovered entries and folded every parsed threshold and ignore pattern into one flat config. A setting in `api/REVIEW.md` therefore governed findings outside `api/`, and its ignore patterns could suppress unrelated findings. That broke the spec promise that a subdirectory `REVIEW.md` governs only its subtree. It also made a subdirectory config mechanically redundant with the shared context file for free prose.

## Decision

Implement per-subtree scoping rather than ratify the flat merge. The spec, the subdirectory scaffold template, and the `build-review-md` wizard promise subtree-local settings; thresholds and ignore patterns need the same finding-file attribution. Free prose stays global: `## Rules` and other guidance continue into the shared context file for every agent.

## Mechanism

**JS.** `workflows/src/filterFindings.js` adds the shared `REVIEW_SETTING_KEYS`, `buildReviewConfig`, `scopeMatchesFile`, and `configForFile`. The builder keeps a root layer, creates one layer per directory, merges duplicate directories in entry order, appends ignores, and sorts scopes by depth with first appearance as the tie-breaker. `workflows/src/args.js` keeps `resolveReviewConfig` as the resolver, delegating its `reviewMd` branch to the builder. `configForFile` returns a fresh flat view: a scope matches only when `dir + "/"` is a byte-exact, case-sensitive prefix of a string `finding.file`; a file equal to `dir`, a sibling prefix such as `apiary/`, a case mismatch, or a non-string file falls back to root. Matching scopes apply structurally from shallow to deep regardless of array order, so deeper present threshold settings win and ignore lists append.

`applyThresholdFilter` selects that view per finding, retaining the existing security `min()` rule and defaults. `applyExclusions` accepts the same config and applies the total order root REVIEW.md ignores, matching subtree ignores shallow-to-deep, then global external exclusions. `applyFilterPipeline` passes the external list separately, so root patterns are not duplicated.

`validateArgs` rejects unsafe `reviewMd` paths and malformed pre-parsed scopes at the waist: backslashes, trailing slashes, empty or dot segments, wrong `REVIEW.md` basenames, invalid threshold domains or severity enums, malformed scope objects, bad directories, unknown keys, non-string ignores, and duplicate directories. `workflows/src/stages.js` derives `stats.reviewMdSubtrees` at the filter call site as `{ dir, matched }` entries for non-root scopes, using the same match helper and counting findings entering the filter stage.

**Python.** `scripts/filter_findings.py` extracts `parse_review_md_text` from `parse_review_md`, then adds `build_review_config` and `config_for_file` as the twin builder and lookup. `apply_threshold_filter` and `apply_exclusions` use the per-finding view with the same precedence. `apply_filter_pipeline` is extracted from `main()` for composed parity; `main()` retains argparse, file I/O, warnings, and external exclusion loading.

**Contracts and docs.** The scoped contract is reflected in `skills/code-gauntlet/SKILL.md`, `skills/code-gauntlet/references/review-md-spec.md`, `skills/code-gauntlet/references/phase2-triage.md`, `skills/code-gauntlet/references/headless-mode.md`, `README.md`, and `docs/duplication-register.md`. The docs state that `.reviewignore` is global, `default_delivery` is root-only, and free prose is shared.

## Tests

Parity dispatch is added in `workflows/test/tools/record_parity.py`, `workflows/test/parity.test.js`, and `tests/test_parity_fixtures.py`. The `review_md` group covers builders `build_root_only`, `build_root_and_sub`, `build_sub_only`, `build_nested_and_sibling`, `build_same_dir_twice`, `build_reversed_input_order`, `build_no_default_fill`, and `build_two_root_entries`; lookups `lookup_root_when_no_scopes`, `lookup_sub_overrides_root`, `lookup_nested_deeper_wins`, `lookup_sibling_prefix_not_matched`, `lookup_file_named_like_dir`, `lookup_scopes_array_reversed`, and `lookup_non_string_file`.

The `threshold` group covers `scoped_sub_eliminates_root_keeps`, `scoped_severity`, `scoped_security_min`, and `scoped_mixed_security_min`. The `exclusions` group covers `scoped_ignore_in_subtree`, `scoped_ignore_outside_subtree_kept`, `scoped_ignore_with_empty_global`, `global_reviewignore_everywhere`, and `scoped_overlap_root_names_first`. The `pipeline` group covers `root_only_config` and `scoped_config_two_subtrees`, including filtered and eliminated IDs, full stats, and `generated_at`.

JS unit coverage in `workflows/test/filter_unit.test.js` includes `REVIEW.md scoped builder and lookup keep settings per matching subtree`, `configForFile orders matching scopes structurally and never mutates config`, `scoped thresholds and ignores use each finding file`, and `scoped exclusions run with an empty external list and root wins overlaps`. `workflows/test/args.test.js` adds dedicated path, scope-shape, threshold-domain, enum, ignore, and duplicate-directory rejection tests. `tests/test_filter_findings.py` adds `test_parse_review_md_text_matches_file_parser`, `test_scoped_thresholds_are_selected_by_finding_file`, `test_builder_merges_same_directory_and_orders_layers`, `test_scoped_lookup_is_non_mutating_and_depth_ordered`, `test_scoped_patterns_apply_with_empty_external_list`, and `test_scoped_overlap_uses_root_pattern_before_child_and_external`.

The stage-level regression is `filter stage scopes a child threshold to findings under that directory` in `workflows/test/stages_vfc.test.js`. The runWith-level regression, `runWith scopes raw child REVIEW.md settings through filter and persisted delivery`, checks delivered and persisted IDs plus filter stats.

Round 1 added: the `a/`, `b/` and `./` no-strip pins (`lookup_a_prefix_not_stripped`, `lookup_b_prefix_not_stripped`, `lookup_dot_slash_prefix_not_stripped` plus JS and Python unit tests), order-independence and config-immutability tests in both twins, the zero-match echo case (unrelated `docs`, case-mismatched `API`, sibling-prefix `apiary/x.py` against scope `api`), and the checkpoint-free runWith regression `runWith fresh pipeline scopes raw REVIEW.md settings before challenge and delivery`. Every mechanism test was verified red by whole-mechanism mutation with byte-exact restores: the builder's 12-entry ledger, then 36 mutations per Codex lens in round 1, 21 and 16 in round 2, and 5 by the opus lens (ledgers in the review record and the session scratch).

## Not done

- The Python CLI keeps a scalar `--review-md PATH` (root layer only); the workflow is the scoped consumer. No consumer needs a multi-file CLI today, so no issue is filed.
- The report receipt's `review_md=present|absent` line still reflects the root quick-check only, so a subdirectory-only configuration renders as absent: #290.
- Free prose is not subtree-scoped; every discovered REVIEW.md's text reaches every context-reading agent and the docs now say so: #291.
- `phase2-triage.md` 2l still opens with "unless REVIEW.md disables them", which no key can do: #289.
- Bench: zero REVIEW.md files exist in the mirror set and the bench fixture is a root-only file, so nothing bench-scored changes; the behavior is latent until a repository adds a subdirectory REVIEW.md. No paired mini on this PR (owner ruling); it rides the wave-end mini.

## Review record

Per the engine-allocation rule: the session adjudicated every verdict; bulk reads, the build, the fix rounds and two review lenses per round ran on Codex; one Claude opus lens read the whole branch; the owner's code-gauntlet run is the cross-family pass.

- Recon: six Codex luna `xhigh` readers (filter twins and parity harness, finding-path spelling, doc surface, args waist and receipt, Python CLI and coverage constraints, bench and discovery). Measured: 184/184 sampled `finding.file` values in the 20 largest recorded artifacts are plain repo-relative (no `a/`, `b/`, `./`, absolute or empty); the report receipt prints only `review_md=present|absent`, never a threshold.
- Design red-team: Claude opus (effort high) REVISE with 2 blockers, 7 majors, 5 minors; Codex sol at `max` REVISE with 4 blockers, 10 majors. Both rejected ratifying the flat merge. Adopted: the scoped-ignore early-return fix, match-time depth ordering (array order irrelevant), no reason suffix or stamp, prose stays global and the docs say so, one `config` argument for both filters, the Python CLI change dropped, waist rejections (backslash, trailing slash, dot and empty segments, wrong basename) plus value-domain checks at root and scope, `scopes` omitted when empty, the echo derived at the call site as `reviewMdSubtrees: [{dir, matched}]`, a non-mutating lookup, one shared setting-key list per twin, the declared ignore order root -> subtree -> external, the Python `apply_filter_pipeline` extraction with composed parity cases, and a runWith-level regression. Rejected with reasons in the spec of record: re-running replayed filter checkpoints under a policy digest (a replayed checkpoint is the prior run's delivered result by contract), refusing a REVIEW.md whose directory holds no changed file (under scoping it governs only its own subtree), rendering subtree provenance in the report receipt (filed as an adjacent issue), and deep-comparing eliminated objects in the threshold/exclusion parity arms (other free-text fields already differ).
- Build: Codex luna `xhigh` from the spec; 12-entry mutation ledger, all red, restores byte-verified.
- Round 1 (59802cc): Codex luna mutation lens FIX (6 majors) and conformance lens FIX (5 majors), 36 mutations each. The one real defect: the exclusions parity arms never passed the fixture config, so the scoped exclusion goldens had recorded the unscoped outcome. Also adopted: distinct overlap patterns so root-first precedence is provable, `a/` and `b/` no-strip pins, order-independence and immutability tests in both twins, zero-match echo cases, a checkpoint-free runWith regression, and the prose-is-global wording across the spec, the wizard docs and the README. Rejected: "JS parity stays green on a Python-only builder mutation" (goldens are recorded from Python; freshness is the pytest gate by harness design, observed STALE). Fixed in 9659337.
- Round 2 (9659337): conformance lens FIX (headless note still said root config applies: fixed; a second item claimed a new regex in args.js that predates this PR: rejected). Mutation lens FIX (a leading `./` strip in both matchers survived every suite: pinned). Claude opus lens FIX with 0 blockers, 1 major, 4 minors: the stats echo's match rule had no discriminating test (a slash-less second implementation survived 1307 tests: pinned with a sibling-prefix echo case); JS `Array.isArray` guards diverged from Python's `.get` default on shapes the waist rejects (aligned); an unused `parseReviewMd` import in args.js; a dead arm in `reviewScopeDepth`; "reaches every agent" corrected to every context-reading agent because the challenger is structurally blind. Its 576-case JS/Python differential over the lookup and builder found 514 byte-identical and 62 divergences, all on ill-typed shapes the waist refuses; its 7-combination Python CLI differential against the base script was byte-identical on stdout and stderr. Fixed in ba45b36.
- Round 3 (ba45b36): one Codex luna confirmation lens, PASS with zero findings: every round-2 fix verified at path:line, the two mutations that had survived round 2 (leading `./` strip in both matchers; slash-less second implementation of the echo predicate) now red, four round-1 whole-mechanism mutations re-run red, repo-wide stale-wording grep clean, all gates green. 4522759 then ratchets the two coverage floors from the CI measurement.
- Owner's code-gauntlet run (4522759): 4 LOW (a Python comment citing the JS-only args waist, duplicated path-segment validation in args.js, a per-finding config-presence recompute in apply_exclusions, and the subtree echo untested for the pre-parsed reviewConfig provenance) plus two CodeQL unused-import notes. All fixed in dd1c9b8 with replies on each thread, after rebasing the branch onto main so the bundle is built by #293's size-gated build (258,058 bytes, 49% of the cap).

## Verification

Local at dd1c9b8 (rebased on main after #293): `node --test` 1318 pass, `pytest tests/` 2014 pass, `pytest bench/tests/` 557 pass, bundle byte-stable under the size gate, parity goldens fresh, twins current, pre-commit clean; coverage scripts 94.78, bench 88.44, JS 98.77 / 90.46 / 98.07 against the ratcheted floors 93.8 / 87.5 / 98.2 / 89.4 / 97.6. Earlier, local at ba45b36: `node --test` 1308 pass, `pytest tests/` 2013 pass, `pytest bench/tests/` 557 pass, bundle byte-stable across two rebuilds, `record_parity.py --check` fresh, `sync_agent_rules.py --check` current, `pre-commit run --all-files` clean. CI at ba45b36 (17 checks green): scripts 94.74, bench 88.31, JS 98.78 / 90.33 / 98.26. Scripts and JS branches sat more than 1.0 pp above their floors, so 4522759 ratchets them to 93.8 and 89.4 per the AGENTS.md policy; bench (0.81 pp), JS lines (0.58) and functions (0.66) stay.

https://claude.ai/code/session_01YR32bHxMf61HstnGEkNxEz


leehopper:
Blocked for the owner's code-gauntlet pass, not by this PR's content: the plugin's `workflows/pipeline.js` has exceeded the Workflow tool's 512 KiB script cap since v3.28.0, so every gauntlet run (including the 2026-09-05 attempt on this PR) dies at the Phase 3 dispatch with `Workflow script file ... exceeds 524288 bytes`. Tracked in #292 with the size history and a build-time fix (drop full-line comments from the generated bundle, add a size gate). This branch's bundle is 541,048 bytes and needs the same fix; after #292 lands on main this PR rebases and the gauntlet pass runs.

scripts/filter_findings.py
```diff
@@ -482,14 +470,96 @@ def parse_review_md(path):
...
+        and scope["dir"] != ""
+        and file.startswith(scope["dir"] + "/")
+    )
+
+
+def config_for_file(config, file):
+    """Return a fresh flat view from root plus matching subtree layers."""
+    source = config or {}
+    # The args waist guarantees the ignore shape is a list at this boundary.
+    view = {"ignore": list(source.get("ignore", []))}
```
leehopper, line 535: **💡 [LOW] config_for_file cites the JS-only "args waist" for a Python-script guarantee**

The new comment says "The args waist guarantees the ignore shape is a list at this boundary," but "args waist" is a term specific to the JS pipeline's argument validator (workflows/src/args.js validateArgs) per workflows/AGENTS.md and AGENTS.md; this Python CLI script has no args waist at all. main() builds config via parse_review_md/parse_review_md_text, which is what actually guarantees ignore is a list here, not any args-waist validation.

**Suggested fix:**
Reword the Python-side comment to cite the actual source of the guarantee (parse_review_md_text always initializing config = {"ignore": []}) instead of referencing the JS-only "args waist" concept.

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet

⚔️ Code Gauntlet review: 4 low-severity improvement suggestions (comment accuracy, simplification x2, test coverage). No bugs, security, cross-file, or type-design findings. See inline comments.

---
Generated by code-gauntlet | Reviewed up to: 45227591c413656dbbcfb26994861ecd6dea17c0

workflows/src/args.js
```diff
@@ -786,21 +786,73 @@ export function validateArgs(args) {
...
+        }
+        if (layer.dir.startsWith('/') || layer.dir.endsWith('/')) {
+          errors.push(`${label}.dir must not have a leading or trailing /`);
+        }
+        if (layer.dir.includes('\\')) errors.push(`${label}.dir must use / separators, not backslashes`);
+        if (REVIEW_MD_PATH_CONTROL_RE.test(layer.dir)) errors.push(`${label}.dir must not contain a control character`);
+        const segments = layer.dir.split('/');
+        if (segments.some((segment) => segment === '')) errors.push(`${label}.dir must not contain an empty path segment`);
+        if (segments.some((segment) => segment === '.' || segment === '..')) errors.push(`${label}.dir must not contain . or .. path segments`);
+      };
```
leehopper, line 845: **💡 [LOW] New reviewConfig.scope-dir and reviewMd path validation duplicate the same segment-checking logic**

The new `validateLayer` closure's dir-validation branch (leading/trailing slash, backslash, control character, empty segment, dot/dot-dot segment) and the new `reviewMd[i].path` validation block a hundred lines later run the same five checks over a `/`-delimited string, just with different label text and one extra basename check on the path side. Before: each block independently calls `.split('/')`, checks `segment === ''`, checks `segment === '.' || segment === '..'`, checks `.includes('\\')`, and runs `REVIEW_MD_PATH_CONTROL_RE.test(...)`. After: a shared `validatePathSegments(value, label, { requireBasename })` helper returns the applicable error strings once, called from both sites.

**Suggested fix:**
Extract the shared segment-validation steps (backslash, control character, empty segment, dot/dot-dot segment, plus an optional basename check) into one helper function used by both the scope-dir validator and the reviewMd-path validator, so a future rule change (e.g. widening allowed characters) is one edit instead of two that can silently diverge.

⚔️ *Code Gauntlet*

scripts/filter_findings.py
```diff
@@ -2202,21 +2274,27 @@ def apply_exclusions(findings, exclusion_patterns):
...
     passed = []
     eliminated = []
 
     for finding in findings:
+        patterns = (
+            config_for_file(config, finding.get("file")).get("ignore", [])
+            + list(exclusion_patterns or [])
+            if config is not None
+            else list(exclusion_patterns or [])
+        )
```
leehopper, line 2291: **💡 [LOW] apply_exclusions recomputes the config-presence branch and re-defaults a guaranteed key on every finding**

Inside the per-finding loop, `patterns = (... if config is not None else ...)` re-evaluates the same `config is not None` test for every finding even though it never changes across the loop, and it calls `.get("ignore", [])` on the dict returned by `config_for_file`, which always sets `ignore` (see `config_for_file`'s `view = {"ignore": list(source.get("ignore", []))}`), so the default is dead. Before: `patterns = (config_for_file(config, finding.get("file")).get("ignore", []) + list(exclusion_patterns or []) if config is not None else list(exclusion_patterns or []))` computed per finding. After: hoist `has_config = config is not None` and `external_patterns = list(exclusion_patterns or [])` once before the loop, then index `config_for_file(...)["ignore"]` directly.

**Suggested fix:**
Hoist the config-presence check and the external-patterns list construction above the loop (mirroring the JS twin's `hasConfig` hoist), and drop the redundant `.get("ignore", [])` default since `config_for_file` always populates that key.

```suggestion
    if config is None and not exclusion_patterns:
        return findings, []

    has_config = config is not None
    external_patterns = list(exclusion_patterns or [])
    passed = []
    eliminated = []

    for finding in findings:
        patterns = (
            config_for_file(config, finding.get("file"))["ignore"] + external_patterns
            if has_config
            else external_patterns
        )
```

⚔️ *Code Gauntlet*

workflows/src/stages.js
```diff
@@ -4070,6 +4070,13 @@ export async function runWith(ctx, rawArgs) {
     // ({findings, reviewConfig, exclusionPatterns, generatedAt}) stays untouched (parity
     // seam, issue #24 req 9).
     const resolvedReview = resolveReviewConfig(A);
+    const reviewMdSubtrees = (Array.isArray(resolvedReview.reviewConfig.scopes)
+      ? resolvedReview.reviewConfig.scopes
+      : []).map((scope) => ({
+        dir: scope.dir,
+        matched: (validateOut.findings || []).filter((finding) =>
+          scopeMatchesFile(scope, finding.file)).length,
+      }));
```
leehopper, line 4079: **💡 [LOW] reviewMdSubtrees stat untested when config arrives via legacy pre-parsed args.reviewConfig**

runWith computes stats.reviewMdSubtrees by mapping resolvedReview.reviewConfig.scopes regardless of whether that config came from raw args.reviewMd (built by buildReviewConfig) or from the legacy pre-parsed args.reviewConfig back-compat path; every test exercising reviewMdSubtrees uses args.reviewMd, none supplies args.reviewConfig with a scopes array to confirm the stat is computed correctly for that provenance too.

**Suggested fix:**
Add a stages_delivery.test.js case that calls runWith with `args.reviewConfig` (not `args.reviewMd`) containing a `scopes` array with matching findings, and assert `out.stats.reviewMdSubtrees` reports the expected dir/matched pairs for that provenance.

⚔️ *Code Gauntlet*

workflows/src/args.js
```diff
@@ -1,4 +1,4 @@
-import { parseReviewMd, loadExclusions } from './filterFindings.js';
+import { parseReviewMd, loadExclusions, buildReviewConfig, REVIEW_SETTING_KEYS } from './filterFindings.js';
```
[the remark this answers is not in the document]
leehopper: Removed in ba45b36 (the unused import was dropped in review round 2); dd1c9b8 keeps the import list to the three names args.js uses.

workflows/src/stages.js
```diff
@@ -19,7 +19,7 @@
 import { DIMENSIONS, AGENTS, AGENT_LABELS, resolvePolicy, FINDING_PROP_TYPES, FINDING_REQUIRED, conditionalSchemaActive } from './registry.js';
 import { merge } from './mergeFindings.js';
 import { applyValidations, pyIntStrict, REACHABILITY_VALUES } from './applyValidations.js';
-import { applyFilterPipeline, SEVERITY_ORDER, applyInjectedProseStrip, applyReplayInjectionScan, normalizeFieldNames } from './filterFindings.js';
+import { applyFilterPipeline, SEVERITY_ORDER, applyInjectedProseStrip, applyReplayInjectionScan, normalizeFieldNames, scopeMatchesFile } from './filterFindings.js';
```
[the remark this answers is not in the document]
leehopper: Fixed in dd1c9b8: SEVERITY_ORDER dropped from the stages.js import; nothing in stages.js referenced it.

scripts/filter_findings.py
leehopper: Fixed in dd1c9b8: the comment now cites parse_review_md_text's config = {"ignore": []} initialization for the CLI path and the JS args waist for the pipeline path.

workflows/src/args.js
leehopper: Fixed in dd1c9b8: one validatePathSegments(value, label, { separateSlashErrors, requireBasename }) helper serves both the scope-dir and reviewMd-path validators; the basename rule stays path-only and every existing rejection message and test is unchanged.

scripts/filter_findings.py
leehopper: Fixed in dd1c9b8 with the suggested shape: has_config and external_patterns hoisted above the loop, config_for_file(...)["ignore"] indexed directly; parity goldens unchanged (record_parity.py --check fresh).

workflows/src/stages.js
leehopper: Fixed in dd1c9b8: stages_delivery.test.js gains a runWith case that supplies args.reviewConfig with a scopes array (no args.reviewMd) and asserts reviewMdSubtrees reports { dir: 'api', matched: 1 }; mutating the echo to depend on A.reviewMd alone turns it red.

Issue #173, opened 2026-08-10
bench/runner/check.py sentinel-count comment (x4/x6) does not match measured occurrences (7/9)


leehopper:
Surfaced by the #69/PR #171 final whole-branch review (adjacent, pre-existing on main — no code on that branch touches it).

`bench/runner/check.py`'s comment describes the degrade sentinels ('no write proof', 'partial-artifacts') as occurring "x4"/"x6" in the bundle; measured occurrences in `workflows/pipeline.js` are 7 and 9 on both main and the PR branch. The comment may be counting emission literals only (vs. comments/prompt text that also contain the phrase) — unverified. Either recount with the rule stated explicitly in the comment, or drop the counts; a wrong count in the comment that gates G3 reads as authority it doesn't have.

leehopper:
Record (2026-09-05): the x4/x6-vs-7/9 premise is stale. Measured today on main (v3.29.0), `workflows/pipeline.js` carries `no write proof` 3 times and `partial-artifacts` twice (PR #293 now strips full-line comments from the bundle, and the comment had been rewritten to x3/x2 by PR #287). PR #294 removes every count from `bench/runner/check.py` and `bench/MEASUREMENT.md`; the load-bearing fact (the bundle carries both sentinels as substrings) is pinned by `docs/machine-parsed-strings.md` rows naming `workflows/pipeline.js` as a producer plus `tests/test_machine_parsed_strings.py`, and a new bench test asserts a wf record echoing the real bundle with `gaps: []` passes G3 clean. Closes with PR #294.

leehopper:
Closed by PR #294 (merged 084b7b5, v3.29.1).

Issue #164, opened 2026-08-10
check.py comments assert a structured workflow-timeout gaps[] signal no code produces


leehopper:
## Summary

Two documentation corrections around the `workflow-timeout` literal, both surfaced by the #146 investigation. They matter because one of them is how #146 came to be filed on a false premise.

## 1. `bench/runner/check.py` comments assert a mechanism no code produces

`bench/runner/check.py:51` and `:81` state that "the structural signal is `workflow-timeout` in `gaps[]` when present." Verified false: nothing under `workflows/src` (the sole producer of every `gaps[]` array that reaches a structured carrier) ever pushes that token. `gaps[]` is assembled inside the workflow sandbox; the timeout is observed *outside* it by `scripts/await_workflow.py`, precisely because the workflow never returned. The literal only ever exists as model-authored prose (per `skills/code-gauntlet/SKILL.md` 417/418/425 and `references/phase3-dispatch.md:38`).

Fix: a ~2-line comment correction so `check.py` stops documenting a structural signal that cannot exist.

## 2. Optional: registry row documenting the token as deliberately unparsed

If a `docs/machine-parsed-strings.md` row is added for `workflow-timeout`, its producer column must list **all four** prose sites — `scripts/await_workflow.py` (docstrings, emission sites, argparse help), `skills/code-gauntlet/SKILL.md:417/418/425`, and `skills/code-gauntlet/references/phase3-dispatch.md:38` — with parser column `—` and a note that it is deliberately unparsed prose (exit-4 and exit-5 are labeled with the identical literal, so no text parser can separate them). An incomplete producer list would be the "looks cross-checked but isn't" failure the registry header warns about; omitting the row entirely is also acceptable.

Suites-only.

leehopper:
Record (2026-09-05): PR #294 rewrites both `check.py` passages (no structural `workflow-timeout` signal exists in any artifact; the awaiter reports it on its own stdout marker; the report can carry it only as prose; the new G6 gate from #165 is what catches a timeout that lost a deliverable), adds the optional `docs/machine-parsed-strings.md` row with all three prose producers (`scripts/await_workflow.py`, `skills/code-gauntlet/SKILL.md`, `skills/code-gauntlet/references/phase3-dispatch.md`, parser `—`), and adds a source test asserting the literal never appears under `workflows/src` or in the bundle (the #146 revisit clause goes red if that changes). Closes with PR #294.

leehopper:
[...]

Issue #146, opened 2026-08-06
Decide whether workflow-timeout deserves its own --check sentinel/gate


leehopper:
## Decision

Should Phase 8 timeout partial delivery (`workflow-timeout` in `gaps[]`) get its own mechanical `--check` sentinel/gate?

## Context

Filed adjacent to #57 / #85 (Wave 2 bench/runner PR). The #57 narrowing of `_DEGRADE_RE` dropped the accidental `partial.artifacts` regex alternative that sometimes matched ordinary English like "partial artifacts" in TEXT carriers. That accidental match could also trip on timeout-degraded delivery prose (`scripts/await_workflow.py` / Phase 8 skill: "deliver whatever partial artifacts exist").

Under the narrowed regex that prose no longer trips G3. That is **correct**: G3 is defined as the **writer** degrade gate (`no write proof` / `partial-artifacts`). Timeout degradation is a different failure with its own structural signal — `workflow-timeout` in `gaps[]` — not prose.

## Options

1. **Add a deliberate `workflow-timeout` sentinel/gate** to `--check` (or a sibling gate), keyed off structured `gaps[]` (and possibly TEXT carriers only if emission is pinned).
2. **Close as not-needed** — timeout already surfaces via G1 (non-`ok` checkpoint / incomplete run) and operator reading of gaps; a separate mechanical gate may not earn its keep.

`close-as-not-needed` is a valid outcome (same shape as #124's conditional framing).

## Out of scope

Implementing either option. This issue is the decision surface only.

## Related

- #57 (G3 regex narrowing)
- #85 (`--retry-failed` / `--check` interaction)
- #101 Wave 2

leehopper:
## Closing as not-needed — the proposed gate is structurally impossible as specified, and the workable variants are worse than no gate

Decision reached after two independent verification passes (each claim below re-checked against main at v3.6.2). Recording the full reasoning because this issue's own body carries the premise that fails.

### The premise in this issue's body is false

[...]

PR #294, opened 2026-09-06
fix(bench): add the G6 artifact-completeness gate and archive prior attempts

### LINKED ISSUE #165: Bench has no artifact-completeness gate: a timeout that loses only the report passes G1-G5 green
## Summary

The one genuine mechanical hole surfaced by the #146 investigation: **no `check.py` gate requires the report artifact to exist**, so a partial delivery can pass G1–G5 green.

## The gap, traced

- `scripts/await_workflow.py` exit 4 fires when **any one** of the four `ARTIFACT_BASENAMES` is missing or stale.
- A child that times out at the awaiter but still reaches Phase 8 and writes findings + post-review payload has returncode 0, no `is_error`, a valid echo receipt, and a found payload — `bench/runner/invoke.py` returns `InvokeResult("ok")`, so G1 passes (G1 only compares per-PR state status against `"ok"`).
- The lossy variants currently caught are `no_payload` (G1/G2/G5 collectively) and `workflow_backgrounded` (labeled `invalid` → G1 fails). A run that lost only `code-gauntlet-report-*.md` passes everything.

## Proposed gate

An artifact-completeness gate keyed off **file presence** — a real structural fact — requiring the report artifact (and any other `ARTIFACT_BASENAMES` member the bench contract considers deliverable) to exist and be fresh for the measured run. This is strictly stronger than the gate #146 originally proposed, which was keyed off a `gaps[]` token that no code ever produces (see the #146 close for why that premise fails structurally).

Close-as-not-needed remains a valid outcome if the bench owners judge report-loss acceptable for scoring purposes (scoring reads the payload, not the report) — but the decision should be recorded either way, which is why this is filed rather than folded into the #146 close.

Suites-only if implemented.

leehopper:
## Why?

Closes #165, #164, #173 (Wave 2 of #101).

A child that times out at the awaiter but still reaches Phase 8 and writes findings plus the dry-run payload returns `ok`. If it lost only the report or the checkpoint, `bench/run.py --check` passed G1-G5 green because no gate required those artifacts to exist (#165). Two `check.py` comments also asserted a structural `workflow-timeout` signal in `gaps[]` that no code produces (#164, the false premise behind #146), and one comment stated sentinel occurrence counts for the bundle that drift with every rebuild (#173).

## What Changed?

- **G6 artifact completeness.** `check.py` imports `ARTIFACT_BASENAMES` from `scripts/await_workflow.py` (the awaiter's own definition of a complete set, already pinned in lockstep with `stages.js`) and requires the four members to form one coherent set: exactly one regular-file match each, one shared SHA, non-empty, JSON members parseable (invalid UTF-8 reports as a parse failure, never an exception). G2 still parses every findings artifact it finds. Decision of record: all four members are deliverables, not only the report, because the awaiter's `artifacts_only` fallback already requires all four and a partial persist raises a G3 gap, so G6 adds no false failures on healthy runs.
- **Prior attempts are archived before re-invocation.** `--resume` and `--retry-failed` now move the prior attempt's deliverables into `pr_dir/superseded/` beside the #85 wf-record archive, so presence is a fact about the measured attempt. Every consumer that walks `pr_dir` recursively (echo receipt, identity receipt, scorer payload discovery, `_find_payload`) skips the `superseded` path component through one predicate; before this, a retried PR that produced no payload would have been scored from the archived one. The archiver propagates move errors so a retry can never read a mixed set.
- **Comment corrections.** `check.py` no longer claims a `gaps[]` timeout token (the awaiter reports `workflow-timeout` on its own stdout; a report can carry it only as prose; G6 is what catches a lost deliverable) and no longer states occurrence counts. `docs/machine-parsed-strings.md` registers `workflow-timeout` as deliberately unparsed with all three prose producers, so a rename on either side goes red. A source test pins that nothing under `workflows/src` or the bundle emits the token (the #146 revisit clause).
- **Bench coverage floor 87.5 to 87.7**, pinned from this PR's CI measurement (88.66) per the AGENTS.md ratchet policy (headroom exceeded 1.0 pp).
- Docs: MEASUREMENT gate list item 6, README `--check` row, `--check` summary line prints `deliverable_artifacts`.

## Additional Notes

- #173's x4/x6-vs-7/9 premise is stale: the bundle today carries `no write proof` 3 times and `partial-artifacts` twice, and the comment no longer states any count; `tests/test_machine_parsed_strings.py` pins their presence instead.
- Built by a Codex luna implementer from a plan red-teamed by one Claude opus agent and one Codex sol pass (the red-team caught the scorer rglob and the G4 identity rglob before any code was written). Three Codex luna review rounds (mutation ledger and spec conformance lenses in isolated worktrees): 26 mutations run, every one red at the final commit; the majors they found (a `UnicodeDecodeError` escaping `check_run`, a G4 regression test whose fixture never exercised its mutation, and two archiver boundaries without fixtures) are fixed in-PR.
- Local gates at the final commit: scripts coverage 94.78%, bench coverage 88.76%; CI bench measurement 88.66%.

- [x] Bench self-tests pass: `python -m pytest bench/tests/ -q`
- [x] Workflow tests pass: `node --test workflows/test/*.test.js` (no JS changed)
- [x] Bundle rebuilt and byte-exact: `node workflows/build.js` then `git diff --exit-code workflows/pipeline.js` (no JS changed)
- [x] Docs updated

https://claude.ai/code/session_01UDMiJiiohYeKM4tzKwxHfG

---

> [!NOTE]
> **Medium Risk**
> Changes functional-smoke gating and resume/retry artifact handling; mis-archiving or overly strict G6 could fail legitimate runs or affect scoring on retried PRs, though extensive tests cover the new paths.
> 
> **Overview**
> Adds **G6** to the mechanical smoke checker so `--check` fails when a PR is missing any of the four canonical deliverables (findings, report, post-review JSON, checkpoint), when duplicates or mixed SHAs suggest stale leftovers, or when JSON artifacts are empty/unparseable—closing the gap where a run could pass with only partial artifacts (#165).
> 
> **Resume/retry** now archives prior attempt deliverables into `pr_dir/superseded/` (alongside existing workflow-record archiving), and echo/identity receipt lookup, payload discovery, scoring, and G6 globs all **ignore** paths under `superseded/` so retries gate the current attempt only.
> 
> Docs and comments are updated for G1–G6, `workflow-timeout` is registered as deliberately unparsed (G6 catches lost deliverables instead), and the bench coverage floor is ratcheted **87.5 → 87.7** in CI and `AGENTS.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0834f8889c977637f6f24c7d9ee4e498600fd010. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bench/MEASUREMENT.md
```diff
@@ -95,13 +95,16 @@ python3 bench/run.py --check <RUN_ID>
...
 
-`--retry-failed` moves each retried PR's prior `workflows/wf_*.json` into
-`workflows/superseded/` before re-invoking, so `--check` gates the current
-attempt only; forensic records stay on disk. Pre-fix run dirs that still hold
+`--retry-failed` moves each retried PR's prior `workflows/wf_*.json` and prior
+attempt deliverables into their respective `superseded/` directories before
+re-invoking, so `--check` gates the current attempt only; forensic records stay on disk.
+Pre-fix run dirs that still hold
 a failed attempt beside a success keep FAILing `--check` until retried under
 this behavior.
```
leehopper: **💡 [LOW] MEASUREMENT.md still attributes deliverable-superseding to --retry-failed only, but the code now does it for plain --resume too**

This PR's own change to bench/run.py's _resume() removes the prior `if retry:` guard so the wf-record and attempt-artifact supersede calls run unconditionally for every PR in `todo`, which for plain `--resume` is `cp.pending(urls)` -- meaning a plain resume now also archives prior deliverables before re-invoking. The doc text edited in this same diff still says only `--retry-failed` does this, so a reader of MEASUREMENT.md is told a narrower scope than the shipped behavior.

**Suggested fix:**
Reword the MEASUREMENT.md passage to say both `--resume` and `--retry-failed` archive the todo set's prior wf records and deliverables before re-invoking, since the guard that used to scope this to retry-only was removed in this PR.

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet


---

1 inline comment(s) were posted; the following 1 finding(s) reference lines outside this diff and are included here instead: A finding listed here may not have an anchoring problem of its own — a consolidation group whose primary could not be anchored inline is listed here in full, corroborators included.

#### `bench/runner/invoke.py:379`

**💡 [LOW] supersede_workflow_records docstring says it's called at --retry-failed start, but its only caller now invokes it unconditionally from _resume**

The docstring for supersede_workflow_records was not touched by this diff, but the diff changed its sole call site (bench/run.py's _resume) from `if retry: ... supersede_workflow_records(...)` to an unconditional call for every URL in `todo`, which includes plain `--resume`'s pending set. The docstring's claim that it is 'Called at `--retry-failed` start' is therefore now inaccurate about when the function actually runs.

**Suggested fix:**
Update the docstring to say it is called at the start of every `_resume` invocation (both plain `--resume` and `--retry-failed`), not only `--retry-failed`.

---
Generated by code-gauntlet | Reviewed up to: 0834f8889c977637f6f24c7d9ee4e498600fd010

bench/MEASUREMENT.md
leehopper: Fixed in 71a0116: MEASUREMENT.md now says both `--resume` and `--retry-failed` archive each todo PR's prior wf records and deliverables; the `supersede_workflow_records` docstring (the unanchored finding at invoke.py:379) says the same.

Issue #296, opened 2026-09-06
Phase 8: render FIX-task payloads in code (scripts/render_fix_tasks.py) instead of model composition


leehopper:
## Problem

After a "Yes" on the Phase 8 task-board question, the orchestrator hand-composes every FIX task from the template in `skills/code-gauntlet/references/fix-task-metadata.md`: it detects the toolchain from a table, picks one or two sibling files as `patterns_to_follow`, and transcribes each finding's title, description, location, evidence, suggestion, severity, and dimension into a `TaskCreate` description and a `TaskUpdate` metadata block. The target set ("every finding in the delivered payload") and the mechanism ("TaskCreate only") live in prose. The #207 drift substituted both (GitHub issues from leftover items). The question surface is now pinned by #207; the post-"Yes" payload is still model-composed.

## Fix shape

A stdlib script `scripts/render_fix_tasks.py` that renders the task payloads in code, so the model's only judgement is the Yes/No answer:

- Input: the path of `artifactPaths.postReview` as a required CLI argument (never a reconstructed filename: `code-gauntlet-post-review-input-<sha>.json` sits beside it with only v2 keys). Accept both live shapes: the `{owner, repo, pr_number, sha, review_body, findings}` wrapper (delivery.prIdentity present) and the bare findings array (local-diff review).
- `--repo-root` required; every finding path is resolved beneath it with realpath, and absolute, traversal, and symlink escapes are rejected (the same confinement `scripts/AGENTS.md` requires for model-derived paths).
- Output: one JSON document on stdout through `script_io.write_result`, an array of `{subject, description, metadata}` objects in delivered order, one per finding; status lines on stderr. The skill then instructs: parse the array, assert its length equals the delivered count, call `TaskCreate(subject, description)` per object, feed the returned id into `TaskUpdate(taskId, metadata)`, stop on any parse, count, or call failure, and never fall back to hand composition.
- Optional finding fields are not reliably present in delivered artifacts (measured over 63 local run artifacts, 765 findings: `suggestion` absent 147 times, `cross_file_refs` 45, `origin` 32, `evidence` 20, `line_end` 7; only the eight `FINDING_REQUIRED` fields are guaranteed). Define the rendering for each absent field (omit the section or a fixed literal) and add a golden case for a finding carrying only the required fields.
[...]

PR #297, opened 2026-09-06
fix(skills): ask the Task Board question at the Phase 8 gate and name TaskCreate as the only creation call (#207)

### LINKED ISSUE #207: Phase 8 Task Board question drifts at runtime: agent substituted 'create GitHub issues' for the canonical task-board question
## Problem

The Phase 8 question surface (#35 / PR #199) specifies Stage 2 verbatim: question "Create fix tasks on the task board from these findings?", header "Task Board", options "Yes — create tasks" / "No — done". In a live run on plugin v3.11.4 (session 6759914f, 2026-08-17, reviewing PR #206), the orchestrating agent instead asked:

> "Want me to create GitHub issues from the report's leftover items — e.g. the two unverified/pipeline-degraded findings (stale comment in verify_findings.py, stale test-class banner) and any lower-priority items not posted inline?" — header "Task board", options "No, skip it" / "Yes, file issues"

Both the mechanism (GitHub issues vs task-board FIX tasks) and the target set (leftovers only vs every delivered finding) drifted from spec. The Delivery question (Stage 1) fired verbatim; only Stage 2 drifted. Nothing was created (user declined), but a "Yes" would have produced repo issues the owner didn't ask the pipeline to be able to create — an outward-facing action outside the sanctioned surface.

`tests/test_question_surface.py` pins the 9 question sites and the AskUserQuestion schema in the prompt files, but cannot pin what the model actually asks at runtime — this is the residual prose-fidelity gap.

## Candidate directions (non-binding)

- Tighten phase8-delivery.md's Stage 2 block to state the question text is verbatim-mandatory (same device as other registered echoes), and/or
- Make the mechanism structural: the skill instructs TaskCreate only; explicitly state GitHub/GitLab issue creation is out of scope for Phase 8.

Per the build-the-mechanism rule, prefer whatever removes the paraphrase surface over adding more prose.

leehopper:
## Why?

Closes #207 (Wave 5 of #101).

In a live run the orchestrator replaced the canonical Phase 8 task-board question with "Want me to create GitHub issues from the report's leftover items?" (header "Task board", options "No, skip it" / "Yes, file issues"). Both the mechanism (GitHub issues instead of task-board FIX tasks through TaskCreate) and the target set (leftovers instead of every delivered finding) drifted. The verbatim block lived only in `references/phase8-delivery.md`, and `SKILL.md` carried a one-line gate that pointed at it. `tests/test_question_surface.py` pinned the site count, schema, and uniqueness of the nine question sites but not where a site lives or what its options say. The Yes option also described a task-id shape (`F-01`) that exists nowhere in the product; delivered ids look like `bug-1` and `conv-2`, and the task template emits `FIX-bug-1`.

## What Changed?

- **The Stage 2 question is single-sourced at its point of use.** The `AskUserQuestion` block moves from the reference into `SKILL.md` directly under the Phase 8 MANDATORY GATE, as a top-level fence (a fence nested inside the `>` gate quote is invisible to the question-surface extractor, which would have silently dropped the site count to eight). The reference keeps the post-"Yes" flow and points back at the block.
- **The gate states the action boundary.** On "Yes" the only creation call is TaskCreate, one task per finding in the delivered payload; Phase 8 never creates issues, pull requests, or branches. The Yes option names the mechanism and the real id shape: "One FIX task per delivered finding via TaskCreate (FIX-bug-1, FIX-conv-2, ...)".
- **Four pins in `tests/test_question_surface.py`.** The Task Board site must live in `SKILL.md` and the Delivery site in the reference; the Task Board tuple (question, header, multiSelect, ordered options) is hand-typed in the test and compared whole; the gate paragraph must carry the boundary clauses; the block must sit between its gate and the next section. Each was verified red under its rewrite (block moved back, `via TaskCreate` dropped, block inside the quote, site count set to eight, an option label paraphrased, the old duplicate restored, the boundary clause removed, the block moved below the next section). Site count stays at nine.

## Additional Notes

- **Residual, stated plainly.** No runtime hook can pin what the host model asks. In the drifted run the Delivery question fired verbatim from the same reference file, so the file was in context and the model paraphrased anyway. This change removes the pointer indirection and puts the mechanism and the target set into the bytes the model reads when it asks; it does not prove the cause is fixed. `SKILL.md` is loader-injected, so the block's position past the single-Read window measured in `complete-read-contract.md` does not apply to it.
- **Split out.** Rendering the post-"Yes" FIX-task payload in code (toolchain detection, sibling selection, field mapping) is filed as #296; the design red-team found six open contract questions there that do not belong in this change.
- **Dropped on review.** Banning issue-creation vocabulary in `FORBIDDEN_PATTERNS` was rejected: the phrases have never been in the tree, so the guard could not go red.
- Suites only; no recall or noise surface is touched (the delivered set is fixed upstream in `stages.js`).

- [x] Pipeline tests pass: `python -m pytest tests/ -q` (2041 passed)
- [x] Bench self-tests pass: `python -m pytest bench/tests/ -q` (untouched; CI runs them)
- [x] Workflow tests pass: `node --test workflows/test/*.test.js`
- [ ] JS lint passes: `python3 workflows/test/tools/biome_check.py` (no JS touched; CI runs it)

---

> [!NOTE]
> **Low Risk**
> Documentation and test-only contract pins for interactive delivery UX; no pipeline, posting, or runtime workflow code changes.
> 
> **Overview**
> **Phase 8 Task Board prompt is now single-sourced where the orchestrator actually finishes the run.** The verbatim `AskUserQuestion` block moves from `references/phase8-delivery.md` into `SKILL.md` immediately under a strengthened MANDATORY GATE (as a top-level fence, not inside the gate quote, so the question-surface extractor still counts nine sites). The reference keeps Stage 2 behavior after “Yes” and only points back to that block.
> 
> **The gate documents the creation boundary:** accepting tasks means **TaskCreate only**—one FIX task per delivered finding—with explicit language that Phase 8 must not create GitHub issues, PRs, or branches. The “Yes” option text now names `TaskCreate` and realistic id shapes (`FIX-bug-1`, `FIX-conv-2`, …) instead of `F-01`.
> 
> **`tests/test_question_surface.py` adds regression pins:** Task Board must live in `SKILL.md` and Delivery in the phase-8 reference; the Task Board tuple (question, header, options) is asserted verbatim; the gate must contain the TaskCreate/boundary clauses; and the question block must sit between the gate and `### Print methodology`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 963a7057e8a7b2ed36610548809634942be0cf48. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/test_question_surface.py
+++ b/tests/test_question_surface.py
@@ -174,30 +174,21 @@ class TestQuestionSurface(unittest.TestCase):
                 self.assertEqual(matches[0]["path"], expected_path)
 
     def test_task_board_site_is_pinned_verbatim(self):
-        expected = (
-            "skills/code-gauntlet/SKILL.md",
-            "Create fix tasks on the task board from these findings?",
-            "Task Board",
-            "false",
-            [
+        expected = {
+            "path": "skills/code-gauntlet/SKILL.md",
+            "question": "Create fix tasks on the task board from these findings?",
+            "header": "Task Board",
+            "multiSelect": "false",
+            "options": [
                 (
                     "Yes — create tasks",
                     "One FIX task per delivered finding via TaskCreate (FIX-bug-1, FIX-conv-2, ...)",
                 ),
                 ("No — done", "Finish the review without creating tasks"),
             ],
-        )
+        }
         task_board = next(site for site in sites() if site["header"] == "Task Board")
-        self.assertEqual(
-            (
-                task_board["path"],
-                task_board["question"],
-                task_board["header"],
-                task_board["multiSelect"],
-                task_board["options"],
-            ),
-            expected,
-        )
+        self.assertEqual(task_board, expected)
 
     def test_task_board_gate_states_the_action_boundary(self):
         # The gate paragraph is what the orchestrator reads when it asks; the
--- a/tests/test_question_surface.py
+++ b/tests/test_question_surface.py
@@ -199,6 +199,31 @@ class TestQuestionSurface(unittest.TestCase):
             expected,
         )
 
+    def test_task_board_gate_states_the_action_boundary(self):
+        # The gate paragraph is what the orchestrator reads when it asks; the
+        # boundary it states (TaskCreate only, never an outward-facing artifact)
+        # is the mechanism issue #207 pins, so a rewrite that drops it goes red.
+        skill = (SKILLS / "code-gauntlet" / "SKILL.md").read_text(encoding="utf-8")
+        gate = next(
+            line
+            for line in skill.splitlines()
+            if "MANDATORY GATE" in line and "task-board question" in line
+        )
+        self.assertIn("the only creation call is TaskCreate", gate)
+        self.assertIn("never creates issues, pull requests, or branches", gate)
+
+    def test_task_board_block_sits_at_its_gate(self):
+        # Point of use: the verbatim block follows its gate and precedes the next
+        # section, so the model meets the question where the gate fires.
+        skill = (SKILLS / "code-gauntlet" / "SKILL.md").read_text(encoding="utf-8")
+        gate = skill.index("task-board question below")
+        block = skill.index(
+            'question: "Create fix tasks on the task board from these findings?"'
+        )
+        next_section = skill.index("### Print methodology")
+        self.assertLess(gate, block)
+        self.assertLess(block, next_section)
+
     def test_every_site_conforms_to_the_schema(self):
         offenders = {}
         for s in sites():
--- a/tests/test_question_surface.py
+++ b/tests/test_question_surface.py
@@ -22,6 +22,11 @@ SKILLS = REPO / "skills"
 # that changes the surface, with the reason in the body.
 EXPECTED_QUESTION_SITES = 9
 
+PINNED_SITES = {
+    "Task Board": "skills/code-gauntlet/SKILL.md",
+    "Delivery": "skills/code-gauntlet/references/phase8-delivery.md",
+}
+
 MAX_HEADER_CHARS = 12
 MIN_OPTIONS = 2
 MAX_OPTIONS = 4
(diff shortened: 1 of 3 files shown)


tests/test_question_surface.py
```diff
@@ -160,6 +165,65 @@ def test_site_count_is_pinned(self):
...
+                task_board["options"],
+            ),
+            expected,
+        )
+
+    def test_task_board_gate_states_the_action_boundary(self):
+        # The gate paragraph is what the orchestrator reads when it asks; the
+        # boundary it states (TaskCreate only, never an outward-facing artifact)
+        # is the mechanism issue #207 pins, so a rewrite that drops it goes red.
+        skill = (SKILLS / "code-gauntlet" / "SKILL.md").read_text(encoding="utf-8")
```
leehopper: **💡 [LOW] test_task_board_site_is_pinned_verbatim rebuilds a dict as a positional tuple to compare**

sites() already returns a dict keyed path/question/header/multiSelect/options, but the new test builds a 5-element expected tuple and re-selects the same five fields from task_board in matching order just to compare tuples; a direct dict-to-dict assertEqual says the same thing without requiring the reader to mentally align tuple position with field name. Before: expected = (path, question, header, multiSelect, options) then assertEqual((task_board["path"], task_board["question"], ...), expected). After: expected = {"path": ..., "question": ..., "header": ..., "multiSelect": "false", "options": [...]} then assertEqual(task_board, expected).

**Suggested fix:**
Replace the positional-tuple expected value and the five-field tuple built from task_board with a single dict literal compared directly against task_board via assertEqual(task_board, expected), since dict equality does not depend on key order and removes the need to keep the tuple's field order in sync with the dict's key order by inspection.

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet

⚔️ Code Gauntlet review of `963a7057`: 1 low-severity finding (improvement suggestion), 0 blockers. All 7 dimensions (correctness, security, cross-file impact, tests, conventions, type design) came back clean; simplification flagged one readability nit in the new test.

---
Generated by code-gauntlet | Reviewed up to: 963a7057e8a7b2ed36610548809634942be0cf48

tests/test_question_surface.py
leehopper: Fixed in 6cebfbf: the test compares the site dict directly against a dict literal.

PR #298, opened 2026-09-06
fix(bench): classify the pipeline's ok:false return envelope explicitly (#190)

### LINKED ISSUE #190: bench: classify the all-degraded ok:false envelope (#178) explicitly in invoke.py
Adjacent issue surfaced by the #178 fix (PR #189).

PR #189 makes an all-degraded run return `ok:false` with `error` prefixed `all-degraded:` instead of `ok:true` with an empty report. Verified during that work: `bench/runner/invoke.py` never reads the pipeline's own `ok`/`gaps`/`stats` — it classifies purely from the outer CLI exit code, `is_error`, the config echo, plugin identity, and (when `pr_comments` delivery is configured, which bench's default env does) presence of a post-review payload file. `bench/runner/check.py`'s G3 `_DEGRADE_RE` is scoped to the writer-persist degrade class and deliberately does not own this new token.

Consequence today: an all-degraded bench child most likely lands as `failed` via the step-5 no-payload check (the skill's Phase 8 takes the ok:false path and never posts), which makes it `--retry-failed`-eligible. That is at least loud rather than silently `ok` (the pre-#189 behavior), but it is an incidental classification, not a deliberate one.

Decision to make: should invoke.py gain an explicit branch parsing the workflow envelope's `error`/`gaps` for the `all-degraded:` sentinel (registered in docs/machine-parsed-strings.md with parser `—`, reserved for exactly this) and classify it distinctly — e.g. as `invalid` (non-retryable, mirroring config_echo_mismatch, on the argument that retrying won't fix a broken model/provider config) vs leaving it `failed` (retry-eligible, on the argument that transient dispatch failures also produce this shape)? The machine-parsed-strings row for `all-degraded:` names this issue as the tracker; if the decision is "incidental `failed` is fine", record that and update the row's note.

Suites-only either way; no measurement coupling.

leehopper:
## Why?

Closes #190 (Wave 4 of #101, adjacent to #178).

`bench/runner/invoke.py` classified a child from the CLI exit code, `is_error`, the config echo, the plugin-identity receipt, and payload presence. It never read the pipeline's own return. An all-degraded child (every discovery dimension degraded, nothing persisted) therefore landed `failed/no_payload` only because the bench pins `pr_comments` delivery; with any other delivery it landed `ok` with no payload, which the scorer assembles as a zero-candidate review. And because that branch writes no report, the `Headless config:` receipt reaches `raw.json` only if the model repeated it in its final message, so the config-echo gate could fire first and mislabel the child `invalid/config_echo_mismatch`, a status `--retry-failed` never touches.

## What Changed?

- **Gate 2c reads the pipeline envelope.** Each child leaves Workflow records whose `result` field is the pipeline return. Right after the `workflow_backgrounded` gate and before the config-echo gate, `invoke_review` parses the records new since the pre-launch snapshot (once; the identity gate shares the parse), keeps those whose `scriptPath` resolves to this checkout and whose `args.outputDir` is this run's output directory (`args` is a dict on current records and a JSON string on older ones), and classifies: any record with `ok: true` supersedes (a headless auto-resume that succeeded); otherwise the first `ok: false` record marks the child `failed`, reason `all_degraded` when `error` or a `gaps` entry starts with `all-degraded:`, else `pipeline_failed` (the catch path, the `checkpoint-shape:` refusal). A stderr line names the reason, mirroring the plugin-mutation and identity gates.
- **Decision of record: `failed`, not `invalid`.** Both are unscored; the only behavioral difference is `--retry-failed` eligibility. `invalid` stays reserved for runs whose bytes cannot be trusted (plugin mutated, identity or echo mismatch, AskUserQuestion). An all-degraded envelope is trustworthy and self-describing, its causes include transient dispatch failures that a retry recovers, and the systematic case (a model or provider mismatch) retries cheaply because no discovery agent ran.
- **The run summary prints a reason histogram** for non-ok statuses (`reasons: all_degraded=1, no_payload=2`) through a new `Checkpoint.detail()` accessor, so the classification is visible without opening state files.
- **Registry and docs.** The `all-degraded:` row in `docs/machine-parsed-strings.md` now names `invoke.py` as its parser (the literal lives in `ALL_DEGRADED_PREFIX`); G3's `_DEGRADE_RE` still deliberately does not own it. `bench/README.md` and `bench/MEASUREMENT.md` describe both reasons and their retry semantics.

## Additional Notes

- **Tests.** Six fake-CLI modes write realistic records (`runId`, `taskId`, `startTime`, `args`, `result`; one writes `args` as a JSON string). Fourteen mutations were verified red, each by a named test: gate deleted (`no_payload` instead of `all_degraded`), gate moved after the echo gate (`invalid` instead of `failed`), gaps conjunct dropped, helper narrowed to the all-degraded prefix, ok-supersede dropped, scriptPath filter dropped (a stale record must still fall to the identity gate), outputDir correlation dropped, unparseable and non-dict records not skipped, string `args` not parsed, sorted-order tie, `Checkpoint.detail` deleted, histogram line deleted, and the identity gate re-walking the records. The two-lens Codex review re-ran the ledger and found the double parse; the second commit fixes it with its own pin.
- Bench coverage: CI measured 88.93% on this branch, so the floor is ratcheted 87.7 → 87.9 here (AGENTS.md and ci.yml, twin regenerated).
- Suites only; no measurement coupling.

- [x] Bench self-tests pass: `python -m pytest bench/tests/ -q` (602 passed)
- [x] Workflow tests pass: no `workflows/` change; CI runs them
- [ ] JS lint passes: no JS touched; CI runs it


AGENTS.md
```diff
@@ -74,9 +74,9 @@ LCOV="$(mktemp -d)/js-coverage.lcov" && node --test --experimental-test-coverage
   && node workflows/test/tools/check_coverage_presence.mjs "$LCOV"
 ```
 
-Floors: Python 93.8 / 87.7, JS 98.2 / 89.4 / 97.6. Each floor is pinned from a PR's CI
+Floors: Python 93.8 / 87.9, JS 98.2 / 89.4 / 97.6. Each floor is pinned from a PR's CI
 measurement, most recently 2026-09-05 from #288 (scripts 94.74; JS branches 90.33); bench
```
leehopper, line 78: **💡 [LOW] New bench coverage floor exceeds the stated 1.0pp tolerance**

The diff raises the bench coverage floor to 87.9 citing PR #298's measurement of 88.93, but 88.93-87.9=1.03 percentage points, which violates the same paragraph's explicit policy that a floor must sit no more than 1.0 pp below the measurement that justified it (every other floor in this table keeps that gap at 0.93-0.96). This looks like an off-by-a-hair rounding choice (87.93 truncated down to 87.9 instead of rounded to 88.0) rather than the 0.9-0.96 pattern used elsewhere.

**Suggested fix:**
Round the new bench floor up to 88.0 (or 87.93) so the CI-fail-under value in .github/workflows/ci.yml and AGENTS.md stays within the documented 1.0pp tolerance, consistent with how the other floors in the same table were set.

**Cited rule:**
> AGENTS.md: 'Policy: a floor sits no more than 1.0 pp below the CI measurement for that gate; ... raise when measured headroom exceeds 1.0 pp.'

⚔️ *Code Gauntlet*

bench/runner/invoke.py
```diff
@@ -963,23 +978,87 @@ def _script_path_matches_repo(script_path, repo_root):
...
+        )
+        reason = (
+            "all_degraded"
+            if (isinstance(error, str) and error.startswith(ALL_DEGRADED_PREFIX))
+            or gap_has_prefix
+            else "pipeline_failed"
+        )
+        failure = (reason, error_text)
+
+    return None if success else failure
```
leehopper, line 1061: **💡 [LOW] _workflow_failure ignores the nested Workflow-record wrapper that scriptpath_from_record already defends against**

_workflow_failure extracts scriptPath via scriptpath_from_record, which unwraps input/toolInput/parameters wrapper keys (a real CLI-version variance the codebase already tests for elsewhere), but then reads args = data.get("args") straight off the top-level record instead of the same wrapper the scriptPath came from; for a wrapped record this returns None, args fails the isinstance(dict) check, and the record is skipped before result.get("ok") is ever inspected, so the new all_degraded/pipeline_failed classification silently never fires for that CLI shape and the PR falls through to a generic failed/no_payload reason instead.

**Suggested fix:**
Factor a shared helper that returns the unwrapped tool-input object (the same dict scriptpath_from_record drills into) and have _workflow_failure read args from that object rather than re-deriving args from the raw top-level record; add a WorkflowFailureTest case with a nested input/args record to pin the behavior.

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet

⚔️ Code Gauntlet reviewed this PR: 3 low-severity findings (2 main, 1 improvement suggestion). See inline comments.

---
Generated by code-gauntlet | Reviewed up to: fe61cb229fafbc6c4f98faadc78a7e401045727b

bench/runner/invoke.py
```diff
@@ -963,23 +978,87 @@ def _script_path_matches_repo(script_path, repo_root):
...
+        gaps = result.get("gaps") or []
+        gap_has_prefix = isinstance(gaps, (list, tuple)) and any(
+            isinstance(gap, str) and gap.startswith(ALL_DEGRADED_PREFIX) for gap in gaps
+        )
+        reason = (
+            "all_degraded"
+            if (isinstance(error, str) and error.startswith(ALL_DEGRADED_PREFIX))
+            or gap_has_prefix
+            else "pipeline_failed"
+        )
```
leehopper, line 1058: **💡 [LOW] Redundant isinstance(error, str) check duplicated in _workflow_failure**

The new `_workflow_failure` helper checks `isinstance(error, str)` twice on the same `error` value within a few lines — once to compute `error_text` (line 1041) and again to compute the `all_degraded` reason (line 1051). Before: two separate isinstance guards on the same variable. After: compute `error_str = error if isinstance(error, str) else ""` once, then reuse it for both `error_text = error_str or str(result.get("failingPhase") or "")` and `reason = "all_degraded" if error_str.startswith(ALL_DEGRADED_PREFIX) or gap_has_prefix else "pipeline_failed"`.

**Suggested fix:**
Normalize `error` to a string once (`error_str = error if isinstance(error, str) else ""`) and reuse it for both the error_text fallback and the all_degraded prefix check, removing the duplicate isinstance test.

```suggestion
        error = result.get("error")
        error_str = error if isinstance(error, str) else ""
        error_text = error_str or str(result.get("failingPhase") or "")
        gaps = result.get("gaps") or []
        gap_has_prefix = isinstance(gaps, (list, tuple)) and any(
            isinstance(gap, str) and gap.startswith(ALL_DEGRADED_PREFIX) for gap in gaps
        )
        reason = (
            "all_degraded"
            if error_str.startswith(ALL_DEGRADED_PREFIX) or gap_has_prefix
            else "pipeline_failed"
        )
```

⚔️ *Code Gauntlet*

AGENTS.md
leehopper: Fixed in b20b588: floor rounded to 88.0 (88.93 measured, 0.93 pp headroom) in ci.yml and AGENTS.md, twin regenerated.

bench/runner/invoke.py
leehopper: Fixed in b20b588: a shared _record_tool_input helper returns the object that carries scriptPath, and the failure gate reads args from that same object; a wrapped-record test pins it.

bench/runner/invoke.py
leehopper: Fixed in b20b588: error is normalized to one string and reused for both the fallback text and the prefix check.

Issue #265, opened 2026-08-31
claude_md_rule carries model meta-prose boilerplate when no rules file exists, poisoning every citation-text consumer


leehopper:
## Problem

`claude_md_rule` is schema-required on the convention dimension (`requiredWhenDimension` in
`workflows/src/registry.js`), so when the reviewed repo has no CLAUDE.md/AGENTS.md the agent
must still fill it — and it fills it with meta-prose: `"No CLAUDE.md/REVIEW.md exists in this
repository; flagged per the case-sensitivity-mismatch check..."`. Measured on the local bench
corpus (2026-08-31, the #247 measurement): ~46% of populated `claude_md_rule` values are such
boilerplate, not quoted repo text.

## Costs, measured

- It renders into posted comments as a `**Cited rule:**` blockquote quoting a rule that does
  not exist (`post_review.py`).
- It was the direct cause of the #247 collateral band: the natural user ignore pattern
  `CLAUDE.md` would have eliminated 12 real correctness findings purely via this boilerplate
  (0 today, 12 widened, all collateral). #247 was declined on that measurement and its
  reopening condition is this issue.

## Fix shape

Make "no rules file" representable without prose: either drop the conditional requirement
when Phase 2 stamps a no-rules-file signal (the `project_rules_absent` gap already exists),
or give the field a structural absent marker the renderer and filters skip. Contract +
registry + render-site change; recall/noise-sensitive only insofar as conventions findings
change shape — decide the measurement tier when picked up.

Surfaced by the #247 measurement (PR #264 red-team).

leehopper:
## Re-scoped 2026-08-31: the causal claim in this issue's body is corrected; fix deferred pending a post-#218 measurement

A design pass targeting this issue was red-teamed before implementation, and the red-team refuted the premise this issue states (which I filed, from the #247 measurement round). Corrections of record:

[...]

PR #299, opened 2026-09-06
feat(bench): measure claude_md_rule absence preambles as a non-gating watch axis (#265)

leehopper:
## Why?

The #265 line in #101 is a measurement: read the post-#218 `claude_md_rule` boilerplate rate at Smoke B. Two paired minis already on disk ran after the conditional schema shipped (`mini-20260901-170531-2cb9104`, `mini-20260903-204424-e8b3af7`; every child's `resolvedPolicy.conditionalSchema` is `true`), so the rate could be read now at zero marginal cost. Nothing in the bench could read it. This PR adds the instrument and records the measurement; the fix stays open on #265 as a Wave 5 item (its red-teamed design is on the issue).

## What Changed?

- **`bench/runner/citations.py`** (stdlib). A populated `claude_md_rule` carries an absence preamble when it states that no rules file or documented rule exists. The pattern is a tuple of five alternatives joined at compile time (never VERBOSE, which silently changes what matches). `measure_run` reports two denominators, the persisted findings artifact (every survivor) and the post-review artifact (the delivered set), each with a per-dimension split, a per-PR split, and a `rule_source` histogram for the structural companion field once it exists. Direct PR directories and direct artifact files only, like `check.py`.
- **`check_run` attaches the block as non-gating `stats.citation_boilerplate`; `bench/run.py --check` prints one watch line;** `python3 -m bench.runner.citations RUN_DIR...` prints the table for historical runs (`--json` for the dicts).
- **Fixtures.** Forty hand-labeled corpus rows (every populated value in every local run, in-sample: the pattern was designed against them) and twenty-one held-out probes, including legitimate citations that contain "no", "not", "missing", or a rules-file name in a non-absence sense. A test rebuilds the pattern without each alternative and asserts a fixture row flips, so a redundant alternative cannot hide.
- **Measurement of record** in the `bench/MEASUREMENT.md` watch ledger, with the read-out instruction for Smoke B and the wave-end paired mini:

| era | absence preamble / populated (findings artifact) | rate |
|---|---|---|
| pre-#218 (runs to 2026-08-18) | 12 / 26 | 46.2% |
| post-#218 (the two September minis) | 7 / 14 | 50.0% |

## Additional Notes

- **Caveats, recorded as such.** The two eras draw on different PR mixes and per-PR rates range from 0/4 to 6/6, so the aggregate cannot separate the schema change from composition; on the PRs common to both eras the paired rate is flat or worse. The classifier is a lower bound whose sensitivity depends on how the agent phrases absence. The delivered-set denominator reads 6/13 on the same two minis.
- **Root cause, verified in the tree** (recorded on #265): the typo/naming sweep prompt extra and the model-authored "judge conventions against the surrounding code" paragraph both sanction rule-less convention findings, while the contract forbids them and the schema requires a non-empty `claude_md_rule`; the agent writes the disclaimer into the field. Sixteen of the nineteen preambles wrap a real precedent citation.
- Bench coverage: CI measured 88.91% on this branch. The floor ratchet to 87.9 rides PR #298 (this branch touches neither AGENTS.md nor ci.yml, so merge order does not matter).
- Suites only; the product is untouched.

- [x] Pipeline tests pass: `python -m pytest tests/ -q` (untouched; CI runs them)
- [x] Bench self-tests pass: `python -m pytest bench/tests/ -q` (592 passed)
- [x] Workflow tests pass: no `workflows/` change; CI runs them
- [ ] JS lint passes: no JS touched; CI runs it


bench/runner/citations.py
```diff
@@ -0,0 +1,236 @@
...
+
+def compile_absence_pattern(alternatives=ABSENCE_PREAMBLE_ALTERNATIVES):
+    """Join *alternatives* into one case-insensitive pattern (never VERBOSE)."""
+    return re.compile("|".join(alternatives), re.IGNORECASE)
+
+
+ABSENCE_PREAMBLE_RE = compile_absence_pattern()
+
+_FINDINGS_GLOB = "code-gauntlet-findings-*.json"
+_POST_REVIEW_GLOB = "code-gauntlet-post-review-*.json"
```
leehopper: **💡 [LOW] citations.py hardcodes the artifact glob instead of deriving it from ARTIFACT_BASENAMES**

citations.py defines its own literal glob strings for the findings and post-review artifact basenames, duplicating scripts/await_workflow.py's ARTIFACT_BASENAMES template that check.py deliberately derives its own _FINDINGS_GLOB from (with a dedicated regression test) specifically to avoid silent drift if the JS persist stage ever renames these artifacts; citations.py's copy has no such derivation or test, so a rename would leave it silently reporting zero citations everywhere instead of failing loudly.

**Suggested fix:**
Derive citations.py's globs from scripts.await_workflow.ARTIFACT_BASENAMES the same way check.py does, and extend the canonical-template test to cover citations.py's copy too.

**Cited rule:**
> AGENTS.md: 'Cross-runtime lists must change together... A change to one side only is a defect even though both files still pass their own tests.'

⚔️ *Code Gauntlet*

bench/run.py
```diff
@@ -1075,6 +1075,23 @@ def _check_only(run_id):
...
+    rate = "n/a" if not populated else f"{preamble / populated:.3f}"
+    print(
+        "  citation_boilerplate: preamble={}/{} (rate {}), convention={}/{}".format(
+            preamble,
+            populated,
+            rate,
+            convention["absence_preamble"],
+            convention["populated"],
+        )
+    )
```
leehopper, line 1094: **💡 [LOW] _check_only's citation watch-line "n/a" rate branch and missing-stat defaults are untested**

bench/run.py's _check_only computes rate = "n/a" if not populated else f"..." and falls back to citation.get("findings") or {} plus a default convention dict when keys are absent, but the only test of this print block (CheckWatchLineTest) supplies populated=2, so the zero-population and missing-key fallback branches never execute under test.

**Suggested fix:**
Add a CheckWatchLineTest variant where stats['citation_boilerplate']['findings'] has populated=0 (and one where 'citation_boilerplate' key is entirely absent from stats), asserting the printed line reads 'rate n/a' and does not raise ZeroDivisionError or KeyError.

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet

⚔️ Code Gauntlet reviewed this PR (2 findings: 1 main, 1 improvement suggestion). Details inline.

---
Generated by code-gauntlet | Reviewed up to: 5da27fe0f3eb75e6834cb5b44bbc881ad138d1aa

bench/runner/citations.py
leehopper: Fixed in 7d57e97: both globs derive from ARTIFACT_BASENAMES and the canonical-template test in test_check.py covers the citations copies.

bench/run.py
leehopper: Fixed in 7d57e97: two CheckWatchLineTest cases cover populated=0 (rate n/a) and a missing citation_boilerplate block.

Issue #303, opened 2026-09-06
configEcho.pr_comment_cap.value must be the string "null", SKILL.md never says so


leehopper:
## What happened

Running `/code-gauntlet PR 300` on 2026-09-06 (plugin 3.30.0), the first Workflow launch was refused at `validateArgs` before any dispatch:

```
invalid args: configEcho.pr_comment_cap.value must be a string
```

I had stamped `"pr_comment_cap": {"value": null, "source": "default"}`. The registry in `workflows/src/args.js` (`valueRule` on the `pr_comment_cap` row) accepts only the string `"null"` or digits in interactive mode. SKILL.md shows the echo as `pr_comment_cap=null (default)` and the waist skeleton says `limits.deliveryCap: null` is a meaningful null, so the natural reading is a JSON null in the echo too.

## Ask

One sentence in SKILL.md at the configEcho stamp: every `configEcho` value is a string, and an unset cap is the string `"null"`. Or relax the rule to accept JSON null for that one key, since `limits.deliveryCap` already carries the real null.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #290, opened 2026-09-05
Resolved-config receipt review_md=present|absent reflects the root REVIEW.md only; subdirectory-only configs render as absent


leehopper:
The resolved-config receipt line `review_md=present|absent (discovery)` (SKILL.md Phase 1 echo, `KNOB_REGISTRY` in `workflows/src/args.js`, rendered into the report's Review Methodology by `renderReport.js`) is sourced from the Phase 1 composite's `review_md_root` section, i.e. the repo-root REVIEW.md only. `validateArgs` enforces it against `reviewConfigPath` (root path). A repository whose only REVIEW.md files live in subdirectories therefore renders `review_md=absent` while those files govern thresholds and ignores for their subtrees (per-subtree scoping shipped in the #196 PR; before it, they governed the whole run and the receipt was equally wrong).

The machine-verifiable channel is right: the pipeline return carries `stats.reviewConfigSource`, `stats.reviewMdEntryCount` and (from #196) `stats.reviewMdSubtrees: [{dir, matched}]`. Only the human receipt is root-only.

Decision to make: (a) redefine the knob's value as "any REVIEW.md discovered" (source stays `discovery`; the skill stamps it after the 2d walk instead of the Phase 1 root check, and `validateArgs`'s `reviewConfigPath` cross-check changes), or (b) add a second registry row (e.g. `review_md_subtrees=<n>`) and keep the root row as is. Either touches `KNOB_REGISTRY`, `renderReport`, `render_report.test.js`, `test_boundary_parity.py` and the bench receipt parser, so it is its own PR. Suites-only.

Surfaced by the #196 red-team (2026-09-04); deliberately left out of that PR.

PR #307, opened 2026-09-07
feat: one code-owned resolver for the resolved config; the workflow derives the typed waist from the receipt (#305)

### LINKED ISSUE #305: One code-owned resolver for the resolved config: printed block, configEcho, and limits from one value
## Problem

Phase 1 prints the `Resolved config:` block as text, Phase 2 hand-assembles `configEcho` (the receipt of that block) and `limits.deliveryCap` (the typed cap) as two more copies of the same resolution, and `validateArgs` checks the last two against each other but never receives the printed text. Three model-typed copies of one value is how #303 happened (a JSON null stamped where the printed token `null` was meant), and the lockstep checks in `workflows/src/args.js` (the `configEcho.pr_comment_cap` vs `limits.deliveryCap` arm) exist only to police that duplication.

## Ask

One code-owned resolver (a stdlib script under `scripts/`, the same shape as `ensure_output_dir.py`) that reads the env pins, the REVIEW.md-derived values and the mode, and emits ONE structured result from which Phase 1 prints the block, Phase 2 copies `configEcho` verbatim, and `limits.deliveryCap` / `delivery.tier` are taken as typed values. The printed block becomes a rendering of the same object, so the receipt is mechanically a receipt. The registry examples (#303) then become the resolver's own fixtures.

Out of scope for #303 (both red-team lenses named it as the root cause but a larger change than that issue justifies). Adjacent, from the #303 design red-team.

leehopper:
## Summary

Configuration resolution now has one code-owned path. `scripts/resolve_config.py` reads the env pins, the root REVIEW.md `## Default Delivery` section and the registry defaults, validates every pin with the same fail-closed rule interpreter the workflow uses, and emits one JSON object with the rendered `Resolved config:` / `Headless config:` block on stderr. Phase 1 calls it; Phase 2 copies `configEcho` verbatim and stamps nothing else about configuration. The workflow derives `limits.deliveryCap`, `delivery.tier`, a light-eligible headless `scopeAnswer` and `configEcho.review_md` from that receipt, so the waist carries one spelling per resolved value and the #303 class of mis-stamp has no place to occur. The knob registry in `workflows/src/args.js` is the single source: it is generated into the resolver, into the SKILL.md receipt fence and into the headless env table.

## What changed

| Area | Files | What |
| --- | --- | --- |
| JS registry and derivation | `workflows/src/args.js`, `workflows/pipeline.js`, `workflows/test/args.test.js` | Rows are JSON-safe data (`rule`, `defaults`, `env`, `reviewMdKey`, `type`, `waistPath`, `derivedFrom`, `deriveWhen`, `nullReceipt`); `matchesRule` is the one fail-closed interpreter; `deriveConfigWaist` fills absent typed waist fields from the receipt; the lockstep arms stay for a caller that stamps both. |
| Resolver | `scripts/resolve_config.py`, `tests/test_resolve_config.py` | stdlib CLI: precedence env > REVIEW.md (`delivery` only) > default, fail-loud `HEADLESS CONFIG ERROR:` line in both modes, `--target` optional, `--plugin-root` mismatch is a setup error, empty stdout on every non-zero exit, block renderer that is a byte twin of the JS renderer. |
| Generator and generated fences | `scripts/generate_contract_requirements.py`, `skills/code-gauntlet/SKILL.md` (receipt fence), `skills/code-gauntlet/references/headless-mode.md` (env table fence) | The node loader projects every row field; the resolver module is loaded by path from `--repo-root`; new fences `knob_registry` (ruff-normal Python literal) and `headless_env_table`; the receipt fence renders both blocks and both receipts through the resolver's registry-parameterised functions, so the examples are the resolver's own fixtures. |
| Skill prose | `SKILL.md`, `references/phase1-preflight.md`, `phase2-triage.md`, `headless-mode.md`, `review-md-spec.md`, `phase8-delivery.md`, `delivery-guide.md` | Phase 1 composite gains the `=== config ===` section and loses the root REVIEW.md cat; later gates branch on `configResult.resolved.<knob>`; the Default Delivery grammar is stated; the duplicate block in phase1-preflight.md is a pointer to the generated fence. |
| Parity and docs tests | `tests/test_boundary_parity.py`, `tests/test_generate_contract_requirements.py`, `tests/test_docs_registry.py`, `tests/test_review_marker.py` | Python block equals the JS-rendered block for every key and both identity lines under lures; the resolver's receipt passes `validateArgs` across the env and cap matrix; rule verdicts agree across runtimes; exact generator projection, fence independence and formatter fixed point; structural pins on the resolver call sites. |
| Docs registries | `docs/machine-parsed-strings.md`, `docs/duplication-register.md` | The resolver is a producer of the two block headers and the error line; the three new twin pairs are registered. |

## Design decisions of record

Two Codex sol max lenses and two Claude opus lenses red-teamed the draft plan; the rulings below came out of that review.

1. One spelling per value on the wire (R1). `configEcho` is the only model-stamped copy; the JS derives the typed fields when absent. Both families named the remaining `limits.deliveryCap` and `delivery.tier` stamps as the same failure class #305 removes.
2. `review_md` is discovery output, not configuration (R2). The receipt entry is filled from `reviewConfigPath`, because a pre-checkout root probe cannot agree with the post-checkout discovery walk that `validateArgs` pins it to.
3. Shape-gated Default Delivery grammar (R3). The body ends at any ATX heading, a fence opener or EOF; HTML comments are removed; only a line shaped `^[a-z_]+(,[a-z_]+)*$` is a candidate. All four lenses showed the draft grammar hard-failed on the plugin's own shipped scaffold.
4. Canonical digits with the safe-integer bound and a fail-closed interpreter in both runtimes (R4), so a receipt always round-trips through `String(Number(v))`.
5. The JS registry stays the source and generates the Python fence (R5, R6): Python `#` identity markers already exist, JS-safe markers do not, and HTML markers would ship in the bundle. The fence is a ruff-normal exploded literal because `ruff format` is an auto-writer in the pre-commit gate.
6. Every function the generator calls takes the registry as a parameter and is total over synthetic rows (R5), so the generator's hand-typed oracle keeps proving it hardcodes nothing.
7. `--target` is optional (R8): the PR-not-found gate runs before the target is final and needs `pr_not_found_policy`; the composite's later call with the final target is the run of record.
8. The Phase 2 checkout guard keeps its shell env read and the provider, gateway and subagent-model rules stay in prose; a shell test is code, not a model-resolved value.

## Contract for skill authors

Phase 1 runs, inside the composite:

```bash
echo "=== config ==="
if ! CONFIG_JSON=$(python3 "{plugin_root}/scripts/resolve_config.py" --target {target_type} --plugin-root "{plugin_root}"); then
  echo "config: FAILED"
  exit 1
fi
echo "$CONFIG_JSON"
```

The stderr block is the Phase 2 gate; the JSON is retained as `configResult` with `mode`, `target`, `block`, `waist.configEcho`, `resolved` and `identity`. Phase 2 copies `waist.configEcho` verbatim and never stamps `limits.deliveryCap`, `delivery.tier`, `configEcho.review_md` or a headless `scopeAnswer`. Every later gate branches on `resolved.<knob>`. Exit 1 is an invalid pin or REVIEW.md value (one `HEADLESS CONFIG ERROR: <VAR>=<value> not in {...}` line); exit 2 is usage or setup; stdout is empty on both.

## Review record

- Recon: five Codex luna read-only items mapped the JS consumers, docs, tests, generators and bench surface.
- Plan: a luna draft, amended by the orchestrator, then red-teamed by two sol max and two opus lenses (verdict revise on all four); the rulings R1-R11 are the adjudication.
- Build A (JS), Build B (resolver, generator, parity) and Build C (prose) ran on luna xhigh, each followed by two luna lenses (mutation ledger, spec conformance) in detached worktrees.
- Build A lenses found two mechanisms without a red test (unknown rule kind, safe-integer bound); the orchestrator added the R5 violation that the derivation keyed on a knob name.
- Build B lenses found nine test gaps (grammar termination, three-value precedence, target guard, plugin-root, renderer lures, cap matrix, rule parity, generator projection and fence independence) and two CLI nits.
- Build C lenses and the orchestrator found two prose regressions (the checkout env guard replaced by a placeholder, the provider paragraph deleted), a dropped `previously_reviewed` condition, a weakened headless no-prompt rule and an inexact grammar paragraph.
- Fix round 1 (code) and 1b (prose) closed every accepted item; round-2 lenses on each reported every mutation red and no accepted blocker or major.
- The orchestrator adjudicated each lens, committed each stage, and reviewed the skill diff by hand. Lee's own code-gauntlet run is the cross-family pass still to come.

## Verification

All three suites, the coverage floors, the generator check, the bundle fixed point and `pre-commit run --all-files` are green on the final commit.

| Gate | Measured | Floor |
| --- | --- | --- |
| `pytest tests/` coverage | 94.26 | 93.8 |
| `pytest bench/tests/` coverage | 89.27 | 88.2 |
| JS lines / branches / functions | 98.84 / 90.59 / 98.28 | 98.2 / 89.6 / 97.6 |

```bash
python -m pytest tests/ -q            # 2129 passed
python -m pytest bench/tests/ -q      # 637 passed
node --test workflows/test/*.test.js  # 1334 passed (Node 24)
python3 scripts/generate_contract_requirements.py --check
node workflows/build.js && git diff --exit-code workflows/pipeline.js
pre-commit run --all-files
```

Resolver smoke from a clean environment at the repo root (`--target pr`, interactive and headless) exits 0 with the block on stderr and the JSON on stdout; `CODE_GAUNTLET_HEADLESS=1 CODE_GAUNTLET_DELIVERY=bogus` exits 1 with empty stdout and one `HEADLESS CONFIG ERROR:` line.

## Out of scope, filed as adjacent issues

- Bench: `build_env` copies the parent environment before overlaying `BENCH_ENV`, which never sets `CODE_GAUNTLET_DELIVERY_TIER`, so an ambient pin can flip a child to `main_only` against `EXPECTED_ECHO`; and `BENCH_ENV` / `EXPECTED_ECHO` are a hand copy of the registry defaults (#308).
- `reviewScope.requested` is still model-derived from `reviewed_policy` (#309).
- #290 (subtree-only REVIEW.md receipt) is unchanged.

Closes #305

https://claude.ai/code/session_014QnnRvizsT6d29d7Rg1fN5


skills/code-gauntlet/references/headless-mode.md
```diff
@@ -59,40 +64,40 @@ The **orchestrator** (the session running this skill and the workflow's own reas
...
+| Closed / merged PR (eligibility) | Proceed — do not stop. Review the pinned head as resolved; posting follows `resolved.post_mode` and delivery follows `resolved.delivery` regardless of PR state. (Interactive mode stops here; headless does not.) |
+| Draft PR | `resolved.draft_policy`: `review` proceeds; `skip` stops the run. |
+| Previously reviewed (Phase 2 2b-post step 3, after checkout) | `resolved.reviewed_policy`: `incremental` scopes the diff to new commits only when `detect_prior_review.py`'s `incremental_safe` is true; when the head has not advanced, the recorded SHA is unresolvable, history was rewritten, or detection errored, it degrades to `full` and discloses the degradation. `skip` stops the run only when `previously_reviewed` is true and `sha_is_ancestor` is true; on rewritten history it proceeds as a full review with the degradation disclosed. Detection is read-only and exits successfully for every outcome. |
+| Trivial / light-scope (all low-risk, <50 lines) | `resolved.trivial_scope` controls the workflow's derived `scopeAnswer`: `light` dispatches only the two core agents (`bug-detector`, `security-reviewer`), and `full` runs all seven dimensions. |
```
leehopper: **💡 [LOW] Per-gate table cites `resolved.trivial_scope`, a field the resolver never emits**

The headless per-gate resolution table says the trivial/light-scope gate is controlled by `resolved.trivial_scope`, but `scripts/resolve_config.py`'s `resolve()` only populates the `resolved` dict for rows whose `waistPath` is `None`; `trivial_scope`'s row has `waistPath: "scopeAnswer"`, so `resolve()` never adds a `trivial_scope` key to `resolved` and the JSON payload's `resolved.trivial_scope` is always absent/undefined, unlike every sibling row in the same table (`pr_not_found_policy`, `draft_policy`, `reviewed_policy`, `delivery`, `post_mode`) whose rows lack a `waistPath` and are genuinely present in `resolved`.

**Suggested fix:**
Either add `trivial_scope` to the `resolved` dict in `resolve_config.py` regardless of `waistPath` (so the documented field actually exists), or reword the table row to point at the real source of truth (e.g. `waist.configEcho.trivial_scope.value`, mirroring how the `delivery.tier`/`pr_comment_cap` rows in SKILL.md and phase1-preflight.md correctly say 'copy its receipt entry' instead of naming a nonexistent `resolved.*` path).

⚔️ *Code Gauntlet*

scripts/resolve_config.py
```diff
@@ -0,0 +1,853 @@
...
+def matches_rule(
+    rule: Any,
+    value: Any,
+    mode: str,
+    *,
+    registry: Sequence[Mapping[str, Any]] | None = None,
+) -> bool:
+    """Return whether a string satisfies a registry rule for ``mode``."""
+    _registry_rows(registry)
+    selected = _selected_rule(rule, mode)
```
leehopper: **💡 [LOW] matches_rule's registry parameter is accepted, threaded through, and then silently discarded**

matches_rule(rule, value, mode, *, registry=None) calls `_registry_rows(registry)` on line 427 and never uses the return value; the rest of the function only consults the `rule`/`mode` arguments already passed in. Both call sites in resolve() (lines 582 and 593) pass `registry=rows` for no effect, which misleads a reader into thinking rule matching consults the registry. Before: `_registry_rows(registry)` then `selected = _selected_rule(rule, mode)`. After: drop the dead call (and, once unused, drop the `registry` parameter and the two `registry=rows` call-site arguments).

**Suggested fix:**
Remove the dead `_registry_rows(registry)` statement from matches_rule, and once the parameter has no remaining use, drop `registry` from the signature and the `registry=rows` keyword arguments at the two call sites in resolve().

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet



---
Generated by code-gauntlet | Reviewed up to: de5d8f385dda6c4e4d79f826cb893b61c4fdafb2

skills/code-gauntlet/references/headless-mode.md
leehopper: Fixed in 7db7491.

scripts/resolve_config.py
leehopper: Fixed in 7db7491.

PR #319, opened 2026-09-09
feat(args): registry-driven waist lockstep and scopeAnswer presence rule (#315, #316)

### LINKED ISSUE #315: Registry-driven validateArgs lockstep for every waistPath row
## Problem

`validateArgs` (`workflows/src/args.js`) checks a stamped waist value against its receipt with one hand-coded arm per knob: `delivery.tier` vs `configEcho.delivery_tier`, `limits.deliveryCap` vs `configEcho.pr_comment_cap`, `reviewScope.requested` vs `configEcho.reviewed_policy` (behind the detector predicate), `scopeAnswer` vs `configEcho.trivial_scope`. `deriveConfigWaist` is registry-driven: a new row with a `waistPath` derives generically, but a caller that stamps a disagreeing value for that row passes validation, because no arm exists for it. The generated derived-waist fence therefore states nothing about `validateArgs`.

## Ask

A registry-driven lockstep pass in `validateArgs` that walks every row with a non-null `waistPath` in the current mode, evaluates the same `DERIVE_WHEN` predicate, `waistMap` and `type` conversion that `deriveConfigWaist` uses, and rejects a present waist value that differs from the mapped receipt value. Keep the genuinely special arms (headless cap must be a number; interactive absent cap needs a `null` receipt; `delivery.tier` absent means `all`) as explicit exceptions with their own tests. Acceptance: a test-injected registry row with a `waistPath` and a stamped conflicting value fails validation; removing the generic pass turns that test green.

Adjacent, from the #311 red-team (PR #314).

### LINKED ISSUE #316: Decide whether validateArgs enforces omission of ungated or derived scopeAnswer values
## Problem

The skill prose says an interactive run stamps `scopeAnswer` only when the trivial-scope question fired, and a headless run never stamps it. `validateArgs` enforces neither: it rejects an interactive `light` on an ineligible `riskTable`, but accepts an interactive `full` when the gate never fired, and accepts a headless-stamped `scopeAnswer` that happens to equal `configEcho.trivial_scope`. Both over-stamps are harmless today (`full` is the default outcome, and a matching headless value derives identically), so the instruction is prose-only.

## Ask

Decide whether omission of ungated or derived fields is enforced. If yes: reject an interactive `scopeAnswer` when `computeLightEligible` is false, and reject a headless-stamped `scopeAnswer` (record caller-supplied paths before derivation so the check sees the stamped value, not the derived one), with tests that go red when each arm is removed. If no: record the decision and leave the prose as guidance.

Adjacent, from the #311 red-team (PR #314).

leehopper:
This PR makes `validateArgs` registry-driven for waist lockstep and aligns `scopeAnswer` presence with light eligibility. It also records derived paths so the null-tolerance disclosure stops naming a field derivation refilled, and regenerates the derived-waist fence.

Closes #315
Closes #316

## Mechanism

- `workflows/src/args.js`: Adds shared, fail-closed `deriveWhenHolds` logic.
- `workflows/src/args.js`: Adds a final registry lockstep pass with mapped receipt values, CSV equality, mode filtering, and focused malformed-receipt handling.
- `workflows/src/args.js`: Retains the explicit presence and absence rules (headless cap must be a number, interactive absent cap needs a `null` receipt, absent tier means `all`, receipt `light` on an ineligible table); each records its path in `reported` so the lockstep pass yields and one fault is one error. An ill-typed cap keeps only its shape error.
- `workflows/src/args.js`: Replaces one-way scope checks with a biconditional presence rule.
- `workflows/src/args.js` and `workflows/src/stages.js`: `deriveConfigWaist` returns `{ args, derivedPaths }`; `normalizeArgsReport` reports `derivedPaths` and the null-tolerance probe skips a dropped key that derivation refilled.
- `workflows/pipeline.js`: Regenerates the workflow bundle.
- `scripts/generate_contract_requirements.py`, `skills/code-gauntlet/*`, and Python tests: Regenerate the waist fence and update its refusal guidance, expected bodies, and census.

## Decisions of record

- R1-R3: The shared predicate fails closed for unknown or inherited names. Special arms remain because undefined values need presence rules.
- R4: `scopeAnswer` must be present exactly when the waist is light-eligible. Either `light` or `full` is accepted when eligible.
- R5: Equal headless caller-stamped values remain accepted. Rejecting correct values adds friction without observable protection.
- R6: Ineligible `full` with a light receipt uses the scope gate and trivial-scope receipt errors, not a deleted equality arm.
- R7: Null disclosures are suppressed only for paths re-derived during normalization. Other nullable keys remain disclosed.
- R8-R9: Generated prose describes conditional receipt disagreement. Refusal guidance lists the three validator refusals.

## Tests

- T1-T2: Probe and real registry lockstep pairs go red when the pass, `waistMap`, row filtering, or old equality arms are changed.
- T3: `derivedFrom` validation goes red when its loop is deleted.
- T4: Unknown and inherited `deriveWhen` names go red when the ownership guard is removed.
- T5-T7: Unhandled types, CSV ordering, and stamped nulls go red when their skips or comparisons change.
- T8-T11: Malformed receipts, duplicate cap errors, mode filtering, and missing `configEcho` go red under their listed guard mutations.
- T12-T15: Scope presence, accepted values, caller stamps, and the combined case go red when the biconditional or special arms change.
- T16 and pipeline R7: Derived null filtering goes red when derived paths are omitted.
- T18: Direct presence and effective-tier fixtures go red when named special arms are deleted.
- T17: Python followability, generated-body, census, and bundle checks cover the contract updates.

## Gates

- `node --test workflows/test/*.test.js` - PASS, 1365 tests, 0 failures (Node 24.20.0).
- `node workflows/build.js && git diff --exit-code workflows/pipeline.js` - PASS after the bundle was committed (the builder's run saw the uncommitted bundle).
- `python3 scripts/generate_contract_requirements.py && python3 scripts/generate_contract_requirements.py --check` - PASS. Last output: `generated registry blocks are current`.
- `/Users/lee/personal/claude-deep-review/.venv/bin/python -m pytest tests/ -q` - PASS. `2140 passed, 1602 subtests passed in 29.15s`.
- `/Users/lee/personal/claude-deep-review/.venv/bin/python -m pytest bench/tests/ -q` - PASS. `645 passed, 106 subtests passed in 22.93s`.
- `python3 workflows/test/tools/biome_check.py` - PASS (run by the orchestrator; the builder's sandbox could not execute the cached binary): `Checked 39 files in 19ms. No fixes applied.`
- JS coverage command from `AGENTS.md` - PASS. Lines `98.86`, branches `90.76`, functions `98.31` (floors 98.2 / 89.6 / 97.6, unchanged).

## Review record

Three luna review rounds (mutation ledger and spec conformance lenses in detached worktrees): round 1 found one major (the headless cap presence arm spliced an earlier shape error out of the list) and two minors (a misplaced omit hint on the receipt refusal, an out-parameter); round 2 found one major in the tests (T4 never reached the derivation caller of the shared predicate) and one minor; round 3 passed both lenses with 20 and 18 mutations red. CI is green on every commit.

## Adjacent

none

https://claude.ai/code/session_01BrPvPqRF4KPVvodPKdVuYr

---

> [!NOTE]
> **Medium Risk**
> Changes pre-dispatch args validation and scope gating in the review pipeline; incorrect logic could block valid runs or accept incoherent waist args, but behavior is heavily covered by new unit and pipeline tests.
> 
> **Overview**
> **Adds registry-driven lockstep validation** so `validateArgs` rejects any stamped derived waist field that disagrees with its `configEcho` receipt (plus `derivedFrom` receipt rows), replacing scattered one-off equality checks with a final `KNOB_REGISTRY` pass that honors `waistMap`, `deriveWhen`, and ordered `csv_list` comparison.
> 
> **Tightens `scopeAnswer` rules** to a biconditional: the field must be present exactly when `riskTable`/`changedLines` are light-eligible (any value when ineligible is refused with an omit hint), while still accepting either `light` or `full` when eligible.
> 
> **Tracks re-derived paths during normalization** — `deriveConfigWaist` / `normalizeArgsReport` now expose `derivedPaths`, and `nullToleranceRejectedKeys` skips null-disclosure gaps for keys immediately refilled from the receipt (e.g. `delivery.tier`, headless `scopeAnswer`).
> 
> **Regenerates operator docs** (derived-waist fence + Phase 3 prose) to state that receipt disagreements are refused pre-dispatch, with matching Python generator/tests and a mirrored `workflows/pipeline.js` bundle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b61100f753df511909f727be82bb65928295307. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/workflows/src/args.js
+++ b/workflows/src/args.js
@@ -334,6 +334,11 @@ function receiptEntryFor(args, descriptor) {
   return entry;
 }
 
+function receiptEntryForKey(args, key) {
+  const descriptor = KNOB_REGISTRY.find((candidate) => candidate.key === key);
+  return descriptor ? receiptEntryFor(args, descriptor) : null;
+}
+
 function dottedValue(object, path) {
   let current = object;
   for (const part of path.split('.')) {
@@ -439,7 +444,8 @@ function deriveConfigWaist(args) {
 }
 
 // normalizeArgsReport(raw) -> { args, dropped, respelled, derivedPaths }
-// `derivedPaths` identifies dropped nulls whose fields were immediately re-derived.
+// `derivedPaths` identifies waist paths the receipt filled because they were absent after null-stripping,
+// whether originally missing or dropped as an explicit null.
 export function normalizeArgsReport(raw) {
   const parsed = typeof raw === 'string' ? JSON.parse(raw) : raw;
   const report = stripNullOptionalsReport(parsed);
@@ -1278,8 +1284,7 @@ export function validateArgs(args) {
   // refusals from also producing the generic waist-path mismatch below.
   const reported = new Set();
   if (isPlainObject(args.configEcho)) {
-    const capDescriptor = KNOB_REGISTRY.find((descriptor) => descriptor.key === 'pr_comment_cap');
-    const capEntry = capDescriptor ? receiptEntryFor(args, capDescriptor) : null;
+    const capEntry = receiptEntryForKey(args, 'pr_comment_cap');
     if (isPlainObject(args.limits) && capEntry !== null) {
       const capValue = args.limits.deliveryCap;
       const capShapeInvalid = capValue !== undefined && capValue !== null
@@ -1297,8 +1302,7 @@ export function validateArgs(args) {
       }
     }
 
-    const tierDescriptor = KNOB_REGISTRY.find((descriptor) => descriptor.key === 'delivery_tier');
-    const tierEntry = tierDescriptor ? receiptEntryFor(args, tierDescriptor) : null;
+    const tierEntry = receiptEntryForKey(args, 'delivery_tier');
     const stampedTier = dottedValue(args, 'delivery.tier');
     if (tierEntry !== null
       && (args.delivery === undefined || args.delivery === null || isPlainObject(args.delivery))
@@ -1308,8 +1312,7 @@ export function validateArgs(args) {
     }
 
     if (args.mode === 'headless') {
-      const trivialDescriptor = KNOB_REGISTRY.find((descriptor) => descriptor.key === 'trivial_scope');
-      const trivialEntry = trivialDescriptor ? receiptEntryFor(args, trivialDescriptor) : null;
+      const trivialEntry = receiptEntryForKey(args, 'trivial_scope');
       if (trivialEntry !== null && trivialEntry.value === 'light'
         && !computeLightEligible(args.riskTable, args.changedLines)) {
         errors.push(`configEcho.trivial_scope is "light" but the riskTable/changedLines are not light-eligible (not every file is low risk, or changedLines >= ${LIGHT_SCOPE_MAX_CHANGED_LINES}) — the orchestrator answered a light/full question the gate never asked`);
--- a/workflows/src/args.js
+++ b/workflows/src/args.js
@@ -402,10 +402,13 @@ function deriveWhenHolds(descriptor, args) {
     || (Object.hasOwn(DERIVE_WHEN, descriptor.deriveWhen) && DERIVE_WHEN[descriptor.deriveWhen].holds(args));
 }
 
-function deriveConfigWaist(args, derivedPaths = []) {
-  if (!isPlainObject(args) || !isPlainObject(args.configEcho) || typeof args.mode !== 'string') return args;
+function deriveConfigWaist(args) {
+  if (!isPlainObject(args) || !isPlainObject(args.configEcho) || typeof args.mode !== 'string') {
+    return { args, derivedPaths: [] };
+  }
   const out = { ...args };
   const configEcho = { ...args.configEcho };
+  const derivedPaths = [];
   let configEchoChanged = false;
   for (const descriptor of KNOB_REGISTRY) {
     if (!descriptor.modes.includes(args.mode)) continue;
@@ -432,7 +435,7 @@ function deriveConfigWaist(args, derivedPaths = []) {
     }
   }
   if (configEchoChanged) out.configEcho = configEcho;
-  return out;
+  return { args: out, derivedPaths };
 }
 
 // normalizeArgsReport(raw) -> { args, dropped, respelled, derivedPaths }
@@ -440,8 +443,9 @@ function deriveConfigWaist(args, derivedPaths = []) {
 export function normalizeArgsReport(raw) {
   const parsed = typeof raw === 'string' ? JSON.parse(raw) : raw;
   const report = stripNullOptionalsReport(parsed);
-  report.derivedPaths = [];
-  report.args = deriveConfigWaist(report.args, report.derivedPaths);
+  const { args, derivedPaths } = deriveConfigWaist(report.args);
+  report.args = args;
+  report.derivedPaths = derivedPaths;
   return report;
 }
 
@@ -1277,14 +1281,16 @@ export function validateArgs(args) {
     const capDescriptor = KNOB_REGISTRY.find((descriptor) => descriptor.key === 'pr_comment_cap');
     const capEntry = capDescriptor ? receiptEntryFor(args, capDescriptor) : null;
     if (isPlainObject(args.limits) && capEntry !== null) {
-      if (args.mode === 'headless' && typeof args.limits.deliveryCap !== 'number') {
-        const capShapeError = 'limits.deliveryCap must be null, absent, or a non-negative safe integer when present';
-        const shapeErrorIndex = errors.indexOf(capShapeError);
-        if (shapeErrorIndex !== -1) errors.splice(shapeErrorIndex, 1);
+      const capValue = args.limits.deliveryCap;
+      const capShapeInvalid = capValue !== undefined && capValue !== null
+        && (!Number.isSafeInteger(capValue) || capValue < 0);
+      if (capShapeInvalid) {
+        reported.add('limits.deliveryCap');
+      } else if (args.mode === 'headless' && (capValue === undefined || capValue === null)) {
         errors.push('headless limits.deliveryCap must be a number matching configEcho.pr_comment_cap');
         reported.add('limits.deliveryCap');
       } else if (args.mode === 'interactive'
-        && (args.limits.deliveryCap === null || args.limits.deliveryCap === undefined)
+        && (capValue === null || capValue === undefined)
         && capEntry.value !== 'null') {
         errors.push('configEcho.pr_comment_cap must be null when limits.deliveryCap is absent or null');
         reported.add('limits.deliveryCap');
@@ -1306,7 +1312,7 @@ export function validateArgs(args) {
       const trivialEntry = trivialDescriptor ? receiptEntryFor(args, trivialDescriptor) : null;
       if (trivialEntry !== null && trivialEntry.value === 'light'
         && !computeLightEligible(args.riskTable, args.changedLines)) {
-        errors.push(`configEcho.trivial_scope is "light" but the riskTable/changedLines are not light-eligible (not every file is low risk, or changedLines >= ${LIGHT_SCOPE_MAX_CHANGED_LINES}) — the orchestrator answered a light/full question the gate never asked; omit scopeAnswer`);
+        errors.push(`configEcho.trivial_scope is "light" but the riskTable/changedLines are not light-eligible (not every file is low risk, or changedLines >= ${LIGHT_SCOPE_MAX_CHANGED_LINES}) — the orchestrator answered a light/full question the gate never asked`);
         reported.add('scopeAnswer');
       }
       const deliveryEcho = configEchoValue(args, 'delivery');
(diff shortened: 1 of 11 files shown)


workflows/src/args.js
```diff
@@ -873,11 +889,12 @@ export function validateArgs(args) {
...
-    }
-    if (lightEligible && args.scopeAnswer === undefined) {
-      errors.push('riskTable/changedLines are light-eligible but scopeAnswer is missing — the light/full gate must have asked, and its answer must be stamped as scopeAnswer');
+    if ((args.scopeAnswer !== undefined) !== lightEligible) {
+      if (args.scopeAnswer !== undefined) {
+        errors.push(`scopeAnswer is "${args.scopeAnswer}" but the riskTable/changedLines are not light-eligible (not every file is low risk, or changedLines >= ${LIGHT_SCOPE_MAX_CHANGED_LINES}) — the orchestrator answered a light/full question the gate never asked; omit scopeAnswer`);
+      } else {
+        errors.push('riskTable/changedLines are light-eligible but scopeAnswer is missing — the light/full gate must have asked, and its answer must be stamped as scopeAnswer');
+      }
     }
```
leehopper, line 904: **💡 [LOW] scopeAnswer refusal now rejects any answer (not just "light") on ineligible riskTable, but phase2-triage.md still describes only the old "light"-only case**

validateArgs's biconditional now refuses ANY stamped scopeAnswer (light or full) when riskTable/changedLines are not light-eligible, per the new error message and the T12 test explicitly asserting this widened behavior, but skills/code-gauntlet/references/phase2-triage.md:397 was not updated and still tells the orchestrator the refusal only fires for "a `\"light\"` answer to a riskTable that wasn't light-eligible", which is the old, narrower condition.

**Suggested fix:**
Update phase2-triage.md:397's parenthetical to describe the biconditional generically (any present scopeAnswer when the riskTable isn't light-eligible, or a light-eligible riskTable with none), matching the wording already used in the diff's SKILL.md:521-524 update.

⚔️ *Code Gauntlet*

workflows/src/args.js
```diff
@@ -1256,59 +1273,84 @@ export function validateArgs(args) {
...
+      if (tierEntry.value !== 'all') errors.push('configEcho.delivery_tier does not match delivery.tier');
     }
-    const deliveryEcho = configEchoValue(args, 'delivery');
-    if (deliveryEcho && deliveryEcho.split(',').includes('pr_comments')
-      && (!isPlainObject(args.delivery) || args.delivery.prIdentity === undefined || args.delivery.prIdentity === null)) {
-      errors.push('delivery.prIdentity is required when configEcho.delivery contains pr_comments');
+
+    if (args.mode === 'headless') {
+      const trivialDescriptor = KNOB_REGISTRY.find((descriptor) => descriptor.key === 'trivial_scope');
+      const trivialEntry = trivialDescriptor ? receiptEntryFor(args, trivialDescriptor) : null;
```
leehopper: **💡 [LOW] Registry find-then-lookup pattern repeated verbatim three times in validateArgs**

The new coherence-arms block in validateArgs introduces the exact same two-line pattern three times — find a KNOB_REGISTRY descriptor by key, then null-safe call receiptEntryFor — once each for pr_comment_cap, delivery_tier and trivial_scope, with the intermediate *Descriptor variable used only to build the *Entry variable and never referenced again. Before (x3): `const capDescriptor = KNOB_REGISTRY.find((d) => d.key === 'pr_comment_cap'); const capEntry = capDescriptor ? receiptEntryFor(args, capDescriptor) : null;`. After: a single helper `function receiptEntryForKey(args, key) { const descriptor = KNOB_REGISTRY.find((d) => d.key === key); return descriptor ? receiptEntryFor(args, descriptor) : null; }` called as `const capEntry = receiptEntryForKey(args, 'pr_comment_cap');` (and similarly for the other two), removing three now-unused descriptor locals and the risk of the three copies drifting.

**Suggested fix:**
Extract a small helper `receiptEntryForKey(args, key)` (find descriptor by key, then delegate to the existing receiptEntryFor, returning null when no descriptor matches) and call it at each of the three sites instead of duplicating the find+null-check inline.

---

**Corroborating finding — type-design-analyzer (type_design, confidence 80):**

**Special-case waist-path exclusions in validateArgs are hardcoded string literals, not read from the KNOB_REGISTRY descriptor they must match**

Encapsulation: 6/10, Expression: 5/10, Usefulness: 8/10, Enforcement: 6/10. The new generic registry-driven lockstep pass is excluded from paths already handled by a special-case arm via a `reported` Set populated with literal strings ('limits.deliveryCap', 'delivery.tier', 'scopeAnswer') rather than `capDescriptor.waistPath`/`tierDescriptor.waistPath`/`trivialDescriptor.waistPath`. If a future edit renames a descriptor's waistPath in KNOB_REGISTRY, the special arm and the generic pass silently desynchronize (both would validate the same field, producing duplicate/contradictory errors) unless the accompanying test fixture (T2) is also updated by hand.

⚔️ *Code Gauntlet*

workflows/src/args.js
```diff
@@ -416,20 +428,24 @@ function deriveConfigWaist(args) {
...
+      derivedPaths.push(descriptor.waistPath);
+    }
   }
   if (configEchoChanged) out.configEcho = configEcho;
-  return out;
+  return { args: out, derivedPaths };
 }
 
+// normalizeArgsReport(raw) -> { args, dropped, respelled, derivedPaths }
+// `derivedPaths` identifies dropped nulls whose fields were immediately re-derived.
```
leehopper: **💡 [LOW] derivedPaths doc comment is narrower than what the field actually contains**

The new JSDoc-style comment says derivedPaths identifies 'dropped nulls whose fields were immediately re-derived', but deriveConfigWaist pushes any waistPath onto derivedPaths whenever dottedValue(out, waistPath) is undefined and a receipt fills it — this includes fields that were never stamped at all, not only fields that were nulled and stripped by stripNullOptionalsReport.

**Suggested fix:**
Reword the comment to describe derivedPaths as every waist path the receipt filled in because it was absent after null-stripping (whether originally missing or dropped-null), e.g. 'derivedPaths lists every waist path the receipt derivation filled because it was absent (originally missing, or dropped as an explicit null).'

⚔️ *Code Gauntlet*

workflows/test/args.test.js
```diff
@@ -684,16 +684,351 @@ test('T4: unknown deriveWhen names fail closed without throwing', () => {
...
+    assert.deepEqual(validateArgs({ ...base, probeWaist: { field: 'mapped' } }), { ok: true, errors: [] });
+  } finally {
+    KNOB_REGISTRY.length = originalLength;
+  }
+});
+
+test('T2: every real waist row and mode has one exact lockstep oracle pair', () => {
+  const cases = {
+    'headless:limits.deliveryCap': {
+      path: 'limits.deliveryCap', key: 'pr_comment_cap', receipt: '6',
```
leehopper: **💡 [LOW] New bare Tn test names collide with unrelated pre-existing tests in the same file**

This diff adds a new local cluster of bare-numbered tests (T1-T3, T5-T16, T18) for the registry-driven lockstep refactor, but four of those numbers (T2, T5, T10, T11) already name unrelated, pre-existing tests earlier/later in the same file, so the same identifier now maps to two different scenarios.

**Suggested fix:**
Rename the new cluster's colliding identifiers (T2, T5, T10, T11) to numbers or suffixed labels not already used in this file, so a failing test name in CI output maps unambiguously to one assertion block.

**Repo precedent:**
> workflows/test/args.test.js establishes disambiguated test naming via issue/PR-suffixed identifiers (e.g. 'T182-ARGS: ...', 'T305-ARGS: ...', 'T309-ARGS: ...') specifically to keep test names traceable and unique within the file.

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet



---
Generated by code-gauntlet | Reviewed up to: 0b61100f753df511909f727be82bb65928295307

workflows/src/args.js
leehopper: Fixed in 644e4fa: the phase2-triage.md:397 parenthetical now reads "any present `scopeAnswer` when the riskTable is not light-eligible, or a light-eligible riskTable with no `scopeAnswer` at all", kept on one line so the census count is unchanged.

workflows/src/args.js
leehopper: Fixed in 644e4fa: `receiptEntryForKey(args, key)` replaces the three find-then-lookup copies.

workflows/src/args.js
leehopper: Fixed in 644e4fa: the comment now says derivedPaths lists every waist path the receipt filled because it was absent after null-stripping, whether originally missing or dropped as an explicit null.

workflows/test/args.test.js
leehopper: Fixed in 644e4fa: the new cluster is renamed to `T315-ARGS:` (lockstep) and `T316-ARGS:` (scopeAnswer presence and derivedPaths disclosure); pre-existing tests untouched.

PR #321, opened 2026-09-11
fix(bench): drop the report G3 carrier and make profiler stage rules data (#280, #282)

### LINKED ISSUE #280: Separate code-owned report text from raw G3 carrier scanning
Surfaced by the #36 pass (PR #276, merged as d617c56), filed as out-of-scope-but-real per the standing practice.

**Fact:** `bench/runner/check.py:112` scans `report.md` as a raw G3 text carrier. Uniform evidence rendering increases the chance that finding evidence containing `partial-artifacts` reaches the report.

**Why it matters:** A finding quoting a G3 sentinel can be classified as a degradation even when the code-owned report text is clean.

**Decision or fix shape:** Separate code-owned report text from finding prose before G3 classification, or replace raw scanning with a structured carrier. Keep T-G3 focused on code-owned text until the classifier decision is made.

**Verification tier:** Suites-only for the parser and guard shape, then paired-mini before changing scoring or classification.

### LINKED ISSUE #282: Label the historical-only `report-writer` profiler bucket
Surfaced by the #36 pass (PR #276, merged as d617c56), filed as out-of-scope-but-real per the standing practice.

**Fact:** `bench/profile_run.py:79` still has a `report-writer` bucket that applies only to pre-v3.25 runs.

**Why it matters:** Current profile readers can mistake an archived bucket for a live pipeline stage after the report-writer deletion.

**Decision or fix shape:** Mark the bucket historical-only in profiler output and documentation, or remove it after historical data no longer needs the label. Add a fixture covering both pre-v3.25 and current profiles.

**Verification tier:** Suites-only, using profiler fixtures and documentation checks.

leehopper:
Closes #280. Closes #282.

## #280: the report leaves the G3 degrade-carrier policy

`bench/runner/check.py` raw-scanned `code-gauntlet-report-*.md` for the writer-degrade sentinels (`no write proof`, `partial-artifacts`). Since the report became a pure code render, that scan could only false-degrade:

- The report is rendered before persistence from `reportInput`, which carries a gap **count**, never gap strings, so code-owned report text cannot hold the sentinel (`render_report.test.js` T-G3 pins it).
- Every persistence path (legacy by-value writer, derived writer, RETURN fallback, replay) pushes the writer-degrade gap through `partial()` into `writeOut.gaps` and the structured Workflow-return carrier, which G3 already judges from its parsed `gaps` array.
- Measured over the retained corpus before the change: 248 reports, 17 sentinel hits; every one of the 6 renderer-era hits sat in model prose (the change summary or a finding body of a review of this repository, whose code names the sentinels). Zero hits in code-owned text.

Mechanism: the report pattern is removed from `_DEGRADE_CARRIER_POLICY`; `raw.json` stays a text carrier; the policy dict is pinned by a hand-typed equality. Tests: a renderer-era report carrying both literals in finding prose with clean structured carriers passes G3 (`ok` true, no `writer degrade` failure); the missing-checkpoint and benign-spacing fixtures move their sentinels into `raw.json`'s `.result` string; one path test per structured branch (legacy writer no-proof gap, derived writer gap, RETURN oversize with a failed fallback, a genuine workflow gap) proves the structured carrier still fails while the report carries nothing. `bench/MEASUREMENT.md` and the policy comment describe the carrier rules.

Recorded loss: archived pre-v3.25 runs are no longer G3-checkable through the report, the only carrier the deleted writer agent's degrade prose ever used; those runs also carry `wf_*.json` and checkpoint records. An era-gated report carrier was considered and declined (it keeps a false-degrade surface alive for an archive nobody re-checks).

## #282: profiler stage rules are data, historical buckets are labelled

`bench/profile_run.py` kept `report-writer` (removed in v3.25) and `verify-input-writer` (slice input moved inline) as live-looking buckets that rendered `0 UNAVAILABLE` rows on every current run, while three live labels (`summarize-bucket-N`, `summarize-merge`, `assemble-artifacts`) classified as `other`.

Mechanism: one ordered `STAGE_RULES` table (name, matcher, label or agent types, optional historical note) derives `STAGE_ORDER`, the predicates and `HISTORICAL_STAGES`. `summarize` matches its bucketed labels; `assemble-artifacts` gains a bucket. A historical stage's row carries `historical: true` and a note and is omitted when it has no agents; a live stage with zero agents keeps its UNAVAILABLE row (that is the anomaly worth seeing); a historical stage that did run (archived runs) renders with a ` (historical)` suffix. Transform gaps (merge, filter) resolve against the nearest earlier and later rows that carry a span, not positional neighbours, so omitted or spanless rows cannot change the answer.

Tests: hand-typed live labels classify to their stages and nothing classifies as `other`; a source pin derived from the table checks each live label literal appears in `workflows/src/stages.js` and each historical label appears nowhere in `workflows/src`; `SyntheticRunBuilder` gains an `era` flag (legacy keeps the old agent list; current drops the historical agents and adds bucketed summarize plus the assembler); both eras are asserted in JSON and Markdown, the current era field-for-field against the legacy rows that are not era-specific, plus the live-zero-agent row and the nearest-span transform rule. `bench/PROFILING.md` describes the table and the rules.

## Verification

Suites only, per both issues. Local gates at HEAD: Python 2146 passed, bench 653 passed, JS 1365 passed, bundle fixed point, generator `--check` current, Biome clean, pre-commit clean, bench coverage 89.56 (floor 88.2). Review: two Codex luna lenses (mutation ledger 10 of 10 red, spec conformance), one fix round (row equality was `agent_count` only; docs reduced to rules), one confirmation round (11 of 11 red).

https://claude.ai/code/session_019tGgdQCJmUhdLwtQgxN4WE

---

> [!NOTE]
> **Low Risk**
> Bench-only smoke checker and profiler behavior; no runtime pipeline or auth changes. Slightly relaxes G3 on report text while structured carriers remain the source of truth for writer degrade.
> 
> **Overview**
> **Smoke G3 (#280):** Removes `code-gauntlet-report-*.md` from the writer-degrade scan policy so finding prose that mentions `no write proof` / `partial-artifacts` no longer fails mechanical smoke. Degrade is still enforced on `workflows/wf_*.json`, `raw.json`, and `code-gauntlet-checkpoint-all-*.json` (structured `gaps` vs text scan). `bench/MEASUREMENT.md` and tests are updated to match, including cases where only structured carriers should trip G3.
> 
> **Profiler (#282):** Replaces ad-hoc stage predicates in `bench/profile_run.py` with an ordered `STAGE_RULES` table (order, matchers, optional `historical_note`). Live labels like `summarize-bucket-*`, `summarize-merge`, and `assemble-artifacts` bucket correctly; legacy `report-writer` and `verify-input-writer` show as **(historical)** when present and are omitted when absent. Merge/filter transform gaps resolve from the nearest earlier/later span-bearing rows. `bench/PROFILING.md` and `test_profile_run.py` cover legacy vs current eras and source pins to `workflows/src/stages.js`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39ea1003c18895542fce0388b4c3afa1c2225f93. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/bench/MEASUREMENT.md
+++ b/bench/MEASUREMENT.md
@@ -74,18 +74,13 @@ python3 bench/run.py --check <RUN_ID>
 2. Payload parse + adapter-required fields + union-schema findings check
    (requires ≥1 `code-gauntlet-findings-*.json` per PR)
 3. Zero `origin=unknown` findings; no writer no-write-proof / partial-artifacts
-   degrade. The report is rendered before persistence from data that carries only a gap
-   count, so code-owned report text is not a G3 writer-degrade carrier. Every persistence
-   path pushes its writer-degrade signal through `partial()` into `writeOut.gaps` and
-   the Workflow-return structured carrier. The legacy renderer-era report hits retained
-   in the corpus were model prose. `raw.json` remains the TEXT carrier and is scanned
-   as-is. Carriers that own a `gaps` array — `workflows/wf_*.json` (the compact
-   Workflow return) and `code-gauntlet-checkpoint-all-*.json` — are judged from that
-   parsed array alone; their raw bytes are never scanned, because a wf record echoes
-   the whole `workflows/pipeline.js` bundle into its `script` field and the bundle's
-   own source contains those sentinels as ordinary substrings. When such a carrier
-   will not parse, or carries no `gaps` at all, it falls back to a raw-text scan with
-   that `script` field blanked first.
+   degrade. The report is not a carrier: it is rendered before persistence from data
+   that carries only a gap count, so code-owned report text cannot hold the signal,
+   although model prose can. G3 scans only the direct `workflows/wf_*.json`, `raw.json`,
+   and `code-gauntlet-checkpoint-all-*.json` carriers. `raw.json` is scanned as text.
+   The other carriers are judged from parsed `gaps` arrays and their string items alone;
+   an empty array is clean. If a structured carrier cannot parse or has no `gaps` array,
+   G3 falls back to a raw scan after blanking its `script` field.
 4. Plugin identity — when the Headless config echo carries `pipeline_version` and
    `plugin_root`, those receipts are validated against the repo's
    `workflows/pipeline.js` version and plugin root (primary). A complete valid
--- a/bench/MEASUREMENT.md
+++ b/bench/MEASUREMENT.md
@@ -74,18 +74,18 @@ python3 bench/run.py --check <RUN_ID>
 2. Payload parse + adapter-required fields + union-schema findings check
    (requires ≥1 `code-gauntlet-findings-*.json` per PR)
 3. Zero `origin=unknown` findings; no writer no-write-proof / partial-artifacts
-   degrade. Carriers that own a `gaps` array — `workflows/wf_*.json` (the
-   compact Workflow return) and `code-gauntlet-checkpoint-all-*.json` — are
-   judged from that parsed array alone; their raw bytes are never scanned,
-   because a wf record echoes the whole `workflows/pipeline.js` bundle into its
-   `script` field and the bundle's own source contains those sentinels as
-   ordinary substrings.
-   When such a
-   carrier will not parse, or carries no `gaps`
-   at all, it falls back to a raw-text scan with that `script` field blanked
-   first. Carriers with no `gaps` structure to parse — `raw.json` (a result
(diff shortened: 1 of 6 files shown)


leehopper (commented):
### ⚔️ Code Gauntlet



---
Generated by code-gauntlet | Reviewed up to: 39ea1003c18895542fce0388b4c3afa1c2225f93

Issue #330, opened 2026-09-12
Verify decoder accepts an escaped surrogate pair the encoder never emits, aliasing one astral character


leehopper:
Found while closing #172 (PR on `fix/172-slice-input-proof-key-order`); out of that issue's scope, filing per standing practice.

## Problem

`decode_inline_slice` accepts a spelling `encodeSliceInline` can never produce. A well-formed surrogate pair is encoded as its UTF-8 bytes (`encodeInlineString` walks UTF-16 and only emits `%uXXXX` for a **lone** surrogate), but the decoder accepts two adjacent `%uXXXX` escapes that form a valid pair.

The two spellings decode to different Python strings:

```
%F0%9F%98%80.js   -> '😀.js'            (len 4)
%uD83D%uDE00.js   -> '\ud83d\ude00.js'          (len 5)
```

They are indistinguishable to the **value** proof, because `escape_lone_surrogates` (`scripts/script_io.py:31`) deliberately re-spells an adjacent pair as one astral character for JS fidelity, so both serialize identically and share one `input_checksum` (`fnv1a32:0xadd66026`).

## Why it is not urgent

The token proof shipped for #172 detects it: the two tokens differ (`0xeda66e71` against `0x754ba2cb`), so a slice carrying the alternative spelling now fails guard (4a) rather than passing silently. This is a decoder-canonicality gap, not a live trust hole.

## Fix shape

The decoder already rejects every other non-canonical spelling by name: lowercase hex, a `%XX` escape of a safe ASCII byte, a non-surrogate `%u` escape. Rejecting an adjacent high+low `%u` pair belongs in the same family, with the same style of named rejection, and would make the accepted alphabet injective **as JS values** rather than only as Python strings.

Note when fixing: `tests/fixtures/parity/slice_inline/lone_surrogates` deliberately carries lone surrogates in sequence (`%uD800leading ... %uDFFF trailing%uD800`), which must keep round-tripping; only a high surrogate immediately followed by a low one is the case to reject.

leehopper:
Closed by PR #345 (squash 72ae060, released in v3.41.1).

[...]

Issue #329, opened 2026-09-12
Duplication register cites line numbers that rot; the checksum row has five stale refs


leehopper:
Found while closing #172 (PR on `fix/172-slice-input-proof-key-order`); out of that issue's scope, filing per standing practice.

## Problem

`docs/duplication-register.md` cites **line numbers**, and nothing resolves them, so every row rots silently on the next refactor. The row for the checksum pair (line 61) now has five stale references out of five:

| cited | actual at 6a881ee |
| --- | --- |
| `stages.js:2343` fnv1a32 | `stages.js:2845` |
| `stages.js:493` DELTA_KEYS | `stages.js:855` |
| `verify_findings.py:1061` `_DELTA_FIELDS` | `verify_findings.py:1394` |
| `assemble_artifacts.py:178` fnv1a32 | `assemble_artifacts.py:185` |
| `workflows/AGENTS.md:43-48,59-60` (the justifying doc) | those lines are now the bundle-generation bullets; the verify-boundary text is further down |

At least one other row is stale the same way: the row citing `workflows/src/stages.js:608-614` against `:1275-1281` as a config-parse duplication no longer resolves to that code.

## Fix shape

Correcting five numbers fixes the symptom and leaves the generator of stale rows in place. The mechanism fix is to cite **symbols** (or heading anchors) instead of line numbers, so a row names something a test can resolve, then add that resolver as a test. The `sliceInputChecksum` / `_input_checksum` pair that #172 just named gives this row its symbols on the JS side.

This is documentation-only and does not affect the pipeline.

leehopper:
Closed by PR #345 (squash 72ae060, released in v3.41.1).

Every location in `docs/duplication-register.md` is now a citation a test resolves at HEAD. The forms are a tracked repo-relative path, a directory with a trailing `/`, a glob, `path::symbol` with nesting joined by `::`, `path::"node:test title"`, and `path#Heading text`. Line numbers, `...`, same-file shorthand and bare basenames are gone. The grammar is one paragraph in the register.

`tests/test_docs_registry.py` enforces it:

- One table parser serves both the row-count test and the new `test_duplication_register_citations_resolve`.
- A classifier decides which code spans are locations.
- An ordered definition-pattern resolver matches Python, JS, YAML hook ids and node:test titles.
- A residual scan covers prose and Markdown link targets outside the tables.
- A subcheck table carries one malformed citation per rejection reason.
- Failures name the register line, the span and the reason.

[...]

PR #331, opened 2026-09-12
fix(verify): prove the slice-input token itself, decline sorted keys (#172)

### LINKED ISSUE #172: Slice-input content proof is key-order sensitive; decide whether a sorted-key canonical form is warranted
## Context

PR #171 (issue #69) added the slice-input content proof: Python hashes `js_stringify_pretty(parsed doc)`, JS hashes `JSON.stringify(content, null, 2)` over the dispatched content. Neither sorts keys, so the proof is insertion-order sensitive: an artifact-writer that re-serializes a slice-input with transposed keys (rather than copying the prompt payload verbatim) produces a legitimate-looking file that fails the proof — a deterministic, disclosed, non-retried degrade of that slice.

## Measured basis for accepting the risk

Both retained captures from `smoke-20260728-144630-a162ecd` are byte-canonical compact JSON — the writer copied the prompt's payload verbatim, key order intact (n=2). The chosen form reuses the plugin's single existing checksum pair and its parity infrastructure (`tests/fixtures/parity/slice_input_proof/`).

## Trigger

Revisit if any smoke/live run shows `stats.inputProof.mismatched > 0` with on-disk content that is value-equal to the dispatch (i.e. the mismatch is ordering, not content). Otherwise this can stay closed-as-accepted.

## Option if triggered

Sorted-key canonicalization on both sides (a second canonical form — the cost is a new cross-runtime serializer contract plus its own parity group, which is why it was not built speculatively). A byte proof over the compact form is the stronger alternative recorded in PR #171's residual-risks section.

leehopper:
Closes #172. Wave 5 item on #101; suites-only, no paired mini on this PR.

## The question and the ruling

#172 asked whether the key-order sensitivity of the verify slice-input content proof warrants a
sorted-key canonical form. **Sorted keys are declined.** The alternative #172 names in the same
breath, a byte proof over the compact token, is shipped instead, because PR #287 removed the one
reason PR #171 had for rejecting it.

PR #171's residual-risks section called the byte proof "the stronger alternative" and rejected it
on one premise: "the writer persists compact JSON, so a byte comparison against the pretty form
would fail every real run." PR #287 deleted that writer. The script now writes `--input` itself,
so the byte the executor carried is available at both ends, and hashing it costs one receipt
field.

## Why not sorted keys

1. Sorting forgives the one thing the proof exists to catch. The token is meant to be copied
   character for character. A value-equal document whose keys arrive in another order is a
   regenerated token, not a copied one.
2. Key order is not the only order the proof covers, so canonicalizing it alone settles nothing.
   Measured over the committed `slice_inline/three_findings` fixture, four different mutations
   each move the value proof:

   | document | `fnv1a32(js_stringify_pretty(doc))` |
   | --- | --- |
   | the fixture as committed | `0x4fd0dedc` |
   | root keys transposed | `0x862110ec` |
   | `findings` array reordered | `0xf10a29dc` |
   | one finding's keys transposed | `0x5624dc64` |
   | an invented extra field | `0x0b2ee1c2` |

3. A sorted form needs a cross-runtime collation contract that does not exist. Python's `sorted`
   orders by code point and JS's `Array.prototype.sort` by UTF-16 code unit: for the keys U+FFFD
   and U+1D11E they disagree. The divergence is solvable (a `utf-16-be` sort key reproduces JS's
   order) and unreachable on today's fixed ASCII projection, so this is a cost rather than a
   defect. It is still a third serializer contract bought for a quarter of the problem.

## What ships instead: the token proof

`sliceTokenChecksum(payload)` in JS and `fnv1a32(args.input_inline)` in Python hash the token
itself, before any decode. `trustSlice` checks it as guard (4a), ahead of the existing value
proof. Both runtimes already had `fnv1a32`, and the parity fixtures already recorded the exact
token per case, so the new golden key is one field in an existing group.

What it buys, measured:

- Cross-runtime agreement on all 10 committed `slice_inline` tokens, Python against Node.
- It is always computable. The value proof returns `None` for a document holding a number the two
  runtimes spell differently (`confidence: 7.5`), and those slices are counted `unprovable` today.
- It separates documents the value proof cannot. `%F0%9F%98%80` and `%uD83D%uDE00` decode to
  different Python strings, but `escape_lone_surrogates` re-spells the surrogate pair as one
  astral character, so both serialize identically and share one value proof (`0xadd66026`). Their
  tokens differ (`0xeda66e71` against `0x754ba2cb`). The JS encoder never emits the second
  spelling, so this is a decoder-language gap the value proof was blind to.
- The token is printable ASCII by construction, so it needs no collation, escaping, or
  number-spelling agreement, which is exactly what makes it cheaper than the sorted form.

## Also in this change

- **The ledger attributes every slice it degrades.** `verifySliceWithRetry` returned only the
  second attempt's `inputFault`, so a slice whose first attempt failed its input proof and whose
  second died another way incremented neither `mismatched` nor `retriedMismatch`. That is the
  counter #172's own revisit trigger reads. Now `second.inputFault ?? first.inputFault`.
- **`sliceInputChecksum` is a named function.** The value proof used to be an inline expression,
  so any test of its order-sensitivity had to re-derive both sides and stayed green when the
  production site was canonicalized. The regression now calls the production function.
- **`_input_checksum`'s docstring was stale.** It still claimed the writer persists compact JSON.
  The rationale is now scoped per path: the value proof exists for the legacy positional path,
  where an external writer chose the on-disk spelling.
- **A test that passed for the wrong reason.** `test_rejects_duplicate_decoded_object_keys` fed
  `%69d`, which dies on the canonical-escape guard because `%69` is the safe ASCII `i`; the
  decoded-duplicate branch it was credited with covering never ran. That branch is deleted along
  with the unused `seen_keys` parameter, and a property test now pins what makes it unnecessary:
  the accepted spelling alphabet is injective, so two distinct raw keys cannot decode equal.

## Verification

Every regression was verified by mutating the implementation and watching it go red.

| mutation | result |
| --- | --- |
| deep key sort inside `sliceInputChecksum` | RED: the transposed-key pin |
| revert to `second.inputFault` | RED: the attribution regression |
| disable the token-proof guard | RED: both token-proof tests |
| accept lowercase `%xx` escapes | RED: the injectivity property |

The fourth needed the whole mechanism mutated. Widening `_INLINE_HEX` alone left
`_INLINE_STRING_RE` demanding uppercase, and the partial mutation passed misleadingly.

Gates: `node --test` 1385 passed, `pytest tests/` 2195 passed, `pytest bench/tests/` 655 passed,
`pre-commit run --all-files` clean, bundle rebuild a no-op. Coverage: scripts 94.08 (floor 93.8),
bench 89.55 (floor 88.2), JS 98.91 / 90.93 / 98.42 (floors 98.2 / 89.6 / 97.6).

Bench and JS-branch headroom now exceed the 1.0 pp policy. Those floors are pinned from a PR's CI
measurement rather than a local one, so the ratchet follows in a second commit here once CI
reports.

## Review

Planned and red-teamed across both model families before implementation: a Codex Astra recon of
the proof path, a four-lens Claude opus panel with an adversarial verifier per lens, then a Codex
Astra and a three-lens Claude opus red-team of the decision memo itself. Both families
independently refused the original memo, for the same reason: it closed #172 without evaluating
the alternative the issue itself names. They also killed three arguments the memo leaned on, and
proved the first draft of the durability test could not fail against the change it existed to
prevent. Every measurement quoted above is fixture-derived and re-runnable.

## Residual risks, disclosed

- The proof stays sensitive to array order and invented fields. That is deliberate and now
  documented rather than incidental.
- `fnv1a32` is a 32-bit non-cryptographic hash. The boundary's threat model is a stale, drifting
  or confused executor, not a Byzantine one, and it says so.
- Above 51,744 characters the inline token's copy fidelity is unmeasured. The budget stays pinned
  below that.
- No live ordering mismatch has ever been recorded, but the corpus cannot prove one never
  happened: a mismatch rescued by the retry emits no UNVERIFIED token by design, and no run
  artifact persists the `inputProof` ledger. Across 19 local runs since the #287 merge, 17 of
  which dispatched a slice, zero slices degraded and two retried.

## Filed separately

- #329 — the duplication register cites line numbers that rot; the checksum row's five references
  are all stale, and the mechanism fix is to cite symbols a test can resolve.
- #330 — the verify decoder accepts an escaped surrogate pair the encoder never emits. The token
  proof shipped here detects it, so it is a canonicality gap rather than a trust hole.

https://claude.ai/code/session_01WuLfVwNCvPBPW3fTWDoWw4

---

> [!NOTE]
> **Medium Risk**
> Changes verify-stage trust gates and executor echo contract; mis-sync between JS and Python would degrade slices to `origin=unknown`, but extensive parity and stage tests pin the behavior.
> 
> **Overview**
> Verify slice trust now uses a **two-part input proof** instead of relying only on a value checksum over the decoded document. The workflow computes **`inline_checksum`** (`fnv1a32` over the percent-encoded `--input-inline` token) and **`trustSlice`** enforces it as guard **(4a)** before the existing **`input_checksum`** value proof **(4b)**. Python **`verify_findings.py`** adds the same field on the receipt (hashed before decode), and the **executor** prompt/schema must echo **`inline_checksum`** verbatim with the other receipt fields.
> 
> This closes **#172** by **rejecting sorted-key canonicalization** for the value proof and documenting that projection key order stays **unsorted**—regenerated tokens with transposed keys must not pass. **`sliceInputChecksum`** / **`sliceTokenChecksum`** centralize the JS side; parity fixtures gain **`token_checksum`**, and tests cover token mismatches, astral vs surrogate spellings, injectivity of inline decoding (removed unreachable duplicate-key branch), and **`inputFault`** attribution when retry fails for a different reason than the first attempt.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f0477096e376c40b52458fb238c8c1df1bcafa8. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:

leehopper:
Addressed in `1db331b`: the injectivity probe's bare `except InputError: pass` is now a named skip with a comment, which is what CodeQL's empty-except rule flagged. Behaviour is unchanged, since a rejected spelling is by definition not an accepted one.

The same commit closes the coverage headroom this PR opened. CI measured bench 89.45 and JS branches 90.93 against floors of 88.2 and 89.6, both beyond the 1.0 pp the floor policy allows, so they rise to 88.5 and 90.0 in both homes (`ci.yml` and the commands quoted in `AGENTS.md`).

leehopper (commented):
### ⚔️ Code Gauntlet

This PR (418 changed lines, resolving issue #172) claims to add a second, stronger half to the verify-stage "slice input proof" in the code-gauntlet review pipeline: alongside the existing `input_checksum` (a value proof computed over the decoded document), it introduces `inline_checksum` (a token proof computed with `fnv1a32` over the exact percent-encoded `--input-inline` bytes handed to the executor, before any decoding). The stated motivation is that the value proof alone cannot distinguish two payloads that decode to different documents but serialize identically (e.g., an astral character spelled as a raw UTF-8 percent-escape vs. an escaped surrogate pair), and that the earlier design explicitly declined to sort JSON keys in the value proof (deliberately preserving order-sensitivity as a regeneration detector) — both codified in new/updated tests and doc comments. It also removes a previously-added decoded-duplicate-key check in `scripts/verify_findings.py`'s inline decoder, arguing the accepted percent-escape spelling alphabet is provably injective, replacing that guard with an injectivity test sweeping code points; and it fixes a bug where `inputFault` from a slice's first verify attempt could be silently dropped (falling back to the first attempt's fault when a retry fails a different way), which affects the pipeline's mismatch/missing counters used for degradation bookkeeping. The touched surface spans the Python verification script (`scripts/verify_findings.py`, high risk), the JS pipeline's twin implementation (`workflows/src/stages.js`, high risk, plus the generated `workflows/pipeline.js` bundle), the executor agent contract and dispatch/validation reference docs, ten frozen parity fixtures gaining a `token_checksum` field, CI/AGENTS.md coverage-floor ratchets (JS branches 89.6%→90.0%, bench 88.2%→88.5%), and an extensive matching test suite (Python unit tests, JS stage/parity tests, and a `verifyDelta.js` test helper) — a cross-runtime change requiring the two checksum functions and their schemas/receipts to be kept in lockstep, which is the primary area of risk reviewers should scrutinize (the JS/Python twin parity, the schema/receipt field additions, and the removed duplicate-key guard's replacement rationale).

3 finding(s) after the gauntlet — 3 low. 3 routed as improvement suggestion(s).

---

### ⚠️ 3 finding(s) could not be anchored inline

0 inline comment(s) were posted; the following 3 finding(s) reference lines outside this diff and are included here instead: A finding listed here may not have an anchoring problem of its own — a consolidation group whose primary could not be anchored inline is listed here in full, corroborators included.

#### `skills/code-gauntlet/references/validation-pipeline.md:27`

**💡 [LOW] Receipt shape summary contradicts the Guard (4) paragraph it sits twelve lines above**

Line 27's compact receipt shape `{ sha, n_in, nonce, deltas_checksum, input_checksum? }` was left unchanged by this diff, while the Guard (4) paragraph on line 39 (rewritten by this same diff) states the receipt must also echo `inline_checksum` as one of two mandatory proof halves — the two lines in the same document now disagree about what fields the receipt carries.

**Suggested fix:**
Add inline_checksum to the compact shape on line 27 so it agrees with the Guard (4) description below it.

#### `scripts/verify_findings.py:1594`

**💡 [LOW] _run_receipt docstring's success-shape claim omits the new inline_checksum field**

_run_receipt's own docstring states the success envelope is `{status:'ok', receipt:{sha, n_in, nonce, deltas_checksum, input_checksum}, ...}`, but the function body this same diff modified now always adds `inline_checksum` to that receipt dict, so the docstring directly under the changed code no longer describes what the function returns.

**Suggested fix:**
Update the docstring's receipt field list to include inline_checksum, ideally noting it (unlike input_checksum) is unconditional.

#### `scripts/verify_findings.py:58`

**💡 [LOW] Module docstring's receipt example is stale after adding inline_checksum**

The module docstring's worked example of the receipt-mode envelope lists only sha, n_in, nonce, deltas_checksum and input_checksum, but this same diff makes _run_receipt() unconditionally add an inline_checksum field to that same dict, so the documented example no longer matches what the script actually emits.

**Suggested fix:**
Add an "inline_checksum": "fnv1a32:0x..." line to the worked example, matching the new unconditional field.

---
Generated by code-gauntlet | Reviewed up to: 1db331bb343da4ca2cb4e9e597509f233f831f4a

leehopper:
Addressed in `1fdb19c`: all three gauntlet findings are one defect seen from three angles — the receipt gained an unconditional `inline_checksum`, and the three summaries of the receipt shape still described the old one. `validation-pipeline.md`'s compact shape, the module docstring's worked example, and `_run_receipt`'s own docstring now all list it, and all three mark `input_checksum` optional, which the code has always made it and none of them said.

Docs only, so no bundle rebuild and no behaviour change. Suites re-run clean: `pytest tests/` 2195 passed, `pre-commit run --all-files` clean.

PR #338, opened 2026-09-15
fix: bound severity labels in rendered fix tasks (#336)

leehopper:
## Why?

`scripts/render_fix_tasks.py` copied agent-controlled `finding["severity"]` into both the Category line and `metadata.severity` without a length bound or closed-set check, while its complexity heuristic read the raw value. This made the FIX-task renderer the third severity surface after the report renderer and review posters without the shared normalization guarantee, affecting any review finding that becomes a FIX task, especially malformed or oversized severity input.

## What Changed?

- The new stdlib-only `scripts/report_severity.py` module defines `JS_TRIM_CHARS` and `normalize_report_severity(raw, labels)` as the shared normalizer.
- The poster keeps a one-argument wrapper that passes its module-level `SEVERITY_EMOJI` at call time, so the generator's placeholder sample still renders correctly.
- The renderer normalizes severity once in `_task` and passes it into `_render_description`, so the Category line, metadata severity, and complexity heuristic use one closed-set label.
- `RenderFixTasksTest.test_normalized_severity_controls_category_complexity_and_model` now pins canonical and unknown labels, Category bytes, complexity, and model selection.
- `RenderFixTasksTest.test_oversized_severity_is_bounded_in_description_and_metadata` pins the low fallback at both output sites, the trivial and haiku selections, the description byte bound, and the absence of the oversized value.
- `RenderFixTasksTest.test_non_string_severities_normalize_to_low` pins low, trivial, and haiku output for non-string severities.
- `RenderFixTasksTest.test_padded_high_severity_normalizes_both_sites` pins trimming, case normalization, the U+3000 edge, and matching Category and metadata values.
- `RenderFixTasksTest.test_empty_and_whitespace_only_severity_normalizes_to_low` pins the low fallback for empty and whitespace-only values.
- `RenderFixTasksTest.test_renderer_reads_the_generated_severity_map` is an in-process test that widens the generated map and pins both output sites to the map read at call time.
- `RenderFixTasksTest.test_renderer_imports_only_stdlib_or_local_helpers` is renamed to pin the renderer's permitted stdlib and local-helper imports.
- The existing severity-twin row in `docs/duplication-register.md` is rewritten in place and the register remains at 51 rows.
- The delivery-guide paragraph documents normalized Category and metadata severity and the complexity and model rule.
- The existing `tool.mypy.overrides` module list in `pyproject.toml` now includes `report_severity`.

## Additional Notes

Rulings of record:

[...]

Issue #336, opened 2026-09-14
Bound the severity field in rendered fix tasks


leehopper:
Surfaced by the #335 pass (branch `fix/334-335-marker-scan-severity-twin`), filed as out-of-scope-but-real per the standing practice.

## Fact

`scripts/render_fix_tasks.py` renders `_one_line(finding["severity"])` into the task category line and into the task metadata (`_render_description` and `_task`). `_one_line` stringifies and collapses to one line but has no length bound and no closed-set check, and `_load_findings` checks only that the field is present.

An agent-controlled severity of 70,000 characters therefore expands the rendered fix task in full. A non-string does not raise there, because `_one_line` stringifies first, so the poster's #335 crash does not reproduce in this consumer.

## Why it matters

The fix-task renderer is the third surface that shows a severity label (after the report renderer, fixed by #327, and the review posters, fixed by #335) and the only one still without the closed set.

## Decision or fix shape

Reuse the closed-set normalization the poster now carries, or add the same twin here, so every rendered severity is one of the generated `SEVERITY_EMOJI` keys with `low` as the fallback. The complexity heuristic (`"trivial" if severity.lower() in ("medium", "low")`) then reads a normalized value.

## Verification tier

Suites-only.

leehopper:
Closed via PR #338 (squash cfc2658). `scripts/report_severity.py` is now the one Python home for the report severity twin (`JS_TRIM_CHARS` and `normalize_report_severity(raw, labels)`); `post_review._normalize_report_severity` is a one-line wrapper passing its generated `SEVERITY_EMOJI` at call time, and `render_fix_tasks.py` normalizes once in `_task` and passes the value into `_render_description`, so the Category line, `metadata.severity` and the complexity heuristic read one closed-set label. Deliberate behavior change: an off-set, oversized, non-string, empty or whitespace-only severity renders `low`, trivial, haiku in the fix task (it rendered as typed with standard, sonnet before); case variants render lowercase; canonical severities and the poster's delivery keys are unchanged. Regressions: oversized, non-string, padded and cased, empty and whitespace severities through the CLI; the closed-set complexity table with hand-typed Category bytes; an in-process pin that the renderer reads the generated map at call time. The register row for the twin names the shared module and both consumers. Adjacent: #339 (consolidate `resolve_config`'s JS-trim alphabet onto the shared constant).

PR #340, opened 2026-09-15
refactor: share resolver JS trim alphabet (#339)

### LINKED ISSUE #339: Consolidate resolve_config's JS-trim alphabet onto report_severity.JS_TRIM_CHARS
Surfaced by the #336 pass (PR #338), filed as out-of-scope-but-real per the standing practice.

## Fact

PR #338 moves the poster's JS `trim()` alphabet into `scripts/report_severity.py::JS_TRIM_CHARS`, the one Python home for the report severity twin, consumed by `scripts/post_review.py` and `scripts/render_fix_tasks.py`. `scripts/resolve_config.py::_JS_TRIM_RE` still carries the same 25-codepoint alphabet as a hand-written regex, spelled twice inside one pattern, and `one_line` uses it for the Phase 1 receipt. The duplication register row for the twin names this copy and records the consolidation as deferred.

## Why it matters

Two hand-written copies of one alphabet drift independently. The poster's live-Node parity test pins the shared copy; nothing pins the resolver's copy against it, so a codepoint dropped from one side is invisible to the other.

## Decision or fix shape

Build `_JS_TRIM_RE` from `report_severity.JS_TRIM_CHARS` (for example `re.compile(f"^[{re.escape(JS_TRIM_CHARS)}]+|[{re.escape(JS_TRIM_CHARS)}]+$")`) so one string carries the alphabet. Keep `one_line`'s operation order (coerce, collapse newlines and spaces, then trim) and the receipt bytes unchanged; `tests/test_boundary_parity.py` pins the receipt against the JS renderer. `resolve_config.py` imports only stdlib today, so the sibling import needs the same `sys.path` bootstrap the other consumers use, or the module keeps its regex and a test asserts the two alphabets are equal. Update the duplication register row either way.

## Verification tier

Suites-only.

leehopper:
## Why?

`scripts/resolve_config.py` hand-wrote the same 25-codepoint JavaScript `trim()` alphabet that `scripts/report_severity.py::JS_TRIM_CHARS` carries, and nothing pinned the resolver's copy, so a codepoint dropped from one side was invisible. The shared constant came from #336 / PR #338, and this change makes the resolver consume that established source of truth.

## What Changed?

- The resolver's `one_line` now strips with `JS_TRIM_CHARS`, and the hand-written `_JS_TRIM_RE` is deleted.
- The resolver imports the sibling helper through the existing `sys.path` bootstrap pattern.
- `tests/test_resolve_config.py::TestResolvePureFunctions::test_one_line_uses_shared_js_trim_chars_at_call_time` pins object identity, proves that a patched `!@` alphabet is read at call time for two-sided trimming, and proves that U+3000 is not trimmed by a leftover local alphabet.
- The `TestCliAgainstRealRegistry` fixture now derives its copied paths from `compute_targets` plus `declared_inputs`, iterated in sorted order.
- The `report_severity.py` docstring and adjacent comment now identify the shared alphabet and its resolver consumer.
- The duplication-register row was edited in place, with the row count unchanged.

## Additional Notes

No behavior change: Receipt bytes are unchanged. The regex character class equals the same 25 codepoints as `JS_TRIM_CHARS`, and newlines collapse to spaces before the trim, so the regex's end anchor before a final newline never applies. The orchestrator compared both pipelines over about 40,000 constructed strings, and the Opus lens compared them over 300,000 random strings plus the non-string values that `_receipt_as_text` coerces; both found no differences.

No JS changed, and the bundle is byte-identical to HEAD. The refactor commit type cuts no release.

Rulings of record:

- R1: Structural consolidation was chosen instead of an equality test because one shared alphabet and an identity and call-time test enforce the dependency directly.
- R4: No new live-Node sweep was added because the poster's existing sweep pins `JS_TRIM_CHARS` against JavaScript `trim()` over all 25 shared codepoints plus five Python-only codepoints.
- A0-3: The fixture is derived from the generator's declared inputs, so no red result is claimed for the fixture repair.
- A0-4: m7 was caught by existing tests, so no fresh-interpreter test was added.
- A0-1: Local Node 24.21.0 was used as a no-regression check; CI's pinned Node 24.18.0 run remains the coverage measurement of record.
- Opus minor: For a damaged install with a missing `report_severity` sibling, exit 1 rather than the documented exit 2 is noted and not changed.

Verification:

- `python -m pytest tests/ -q`: 2246 passed, 2630 subtests passed.
- `python -m pytest bench/tests/ -q`: 657 passed, 106 subtests passed.
- `node --test workflows/test/*.test.js` on Node 24.21.0: 1393 passed, 0 failed.
- Python coverage was 94.27% for `scripts` and `.github` against the 93.8% floor, and 89.55% for `bench` against the 88.5% floor.
- Node coverage was 99.10% lines, 90.92% branches, and 98.62% functions against floors of 98.2%, 90.0%, and 97.7%; the coverage presence check passed.
- `node workflows/build.js && git diff --exit-code workflows/pipeline.js`: exit 0; the bundle was unchanged.
- `python3 scripts/generate_contract_requirements.py --check`: generated registry blocks are current.
- `pre-commit run --all-files`: exit 0; every hook passed.
- `python3 workflows/test/tools/biome_check.py`: exit 0; checked 39 files, no fixes applied.
- `claude plugin validate .`: exit 0; validation passed.

Mutation ledger:

- m1, a local literal regex missing U+3000, was caught red by `tests/test_resolve_config.py::TestResolvePureFunctions::test_one_line_uses_shared_js_trim_chars_at_call_time`.
- m2, default-argument capture of `JS_TRIM_CHARS`, was caught red by the same resolver test.
- m3, removing U+3000 from `report_severity.JS_TRIM_CHARS`, was caught red by `tests/test_post_review.py::TestRenderCommentBody::test_severity_fallback_and_labels_match_js_twin` in its `trim-U+3000` case.
- m4, a separate equal module-local alphabet, was caught red by the resolver test's identity assertion.
- m7, deleting `sys.path.insert` while retaining the bare import, turned 35 bench tests red, including `bench/tests/test_invoke.py::InvokeReviewTest::test_ok` and `bench/tests/test_invoke.py::InvokeReviewTest::test_all_degraded_workflow_record_is_retryable_failure`, and caused fresh-process collection of `tests/test_resolve_config.py` to fail with `ModuleNotFoundError`; existing tests caught it, so no fresh-interpreter test was added.
- The mutation lens also dropped U+FEFF, which was caught by `tests/test_post_review.py::TestRenderCommentBody::test_severity_fallback_and_labels_match_js_twin` in its `trim-U+FEFF` case.
- The mutation lens also replaced `text.strip(JS_TRIM_CHARS)` with plain `strip()`, used `lstrip(JS_TRIM_CHARS)` only, and imported `scripts.report_severity` under a second module identity; the resolver test caught each of these mutations.
- Reverting the fixture to the hand list while omitting `report_severity.py` stayed green in `tests/test_generate_contract_requirements.py::TestCliAgainstRealRegistry` by the A0-3 ruling, so no red was claimed.
- m5 was intentionally not run because production output cannot distinguish its order; source order was preserved by inspection.

Review shape: The plan was drafted by Codex Astra and red-teamed by one Claude Opus lens and one Codex Astra lens. The one-commit branch was reviewed by three Codex luna lenses and one Claude Opus lens; no paired mini, suites-only tier per the issue.

- [x] Bench self-tests pass: `python -m pytest bench/tests/ -q`
- [x] Workflow tests pass: `node --test workflows/test/*.test.js` (Node 24; the glob is required)
- [x] JS lint passes: `python3 workflows/test/tools/biome_check.py` (downloads the pinned Biome once)
- [x] Plugin manifests valid: `claude plugin validate .` (needs the Claude Code CLI; CI runs it regardless)
- [ ] Parity fixtures regenerated with `workflows/test/tools/record_parity.py` if a deterministic transform changed (not applicable: no deterministic JS transform changed)
- [ ] Docs, skill references, and agent contracts updated if behavior changed (not applicable: behavior did not change)

Closes #339

https://claude.ai/code/session_013cebo4Z93HNUUR86s9L89d

---

> [!NOTE]
> **Low Risk**
> Refactor-only: same trim set and receipt behavior; risk is limited to import/bootstrap of `report_severity` when running the resolver.
> 
> **Overview**
> **Consolidates the JavaScript `trim()` alphabet** so `scripts/resolve_config.py` no longer maintains a duplicate regex (`_JS_TRIM_RE`); `one_line` now trims via the shared `JS_TRIM_CHARS` imported from `scripts/report_severity.py` (same 25 codepoints, intended byte-identical receipts).
> 
> `report_severity` docs/comments name the resolver as a second consumer. The duplication register row is updated to record that consolidation and points at a new resolver test. **`TestResolvePureFunctions::test_one_line_uses_shared_js_trim_chars_at_call_time`** asserts object identity with `report_severity`, call-time patching of the alphabet, and that ideographic space (U+3000) is not stripped when the patch excludes it.
> 
> The contract-requirements CLI test fixture now copies files from **`compute_targets` ∪ `declared_inputs`** (sorted) instead of a hand-maintained path list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3cd7d32fee3ce1ce9db5fb85dccba01b7a9f701. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


leehopper:

scripts/report_severity.py
```diff
@@ -1,13 +1,15 @@
...
+used by normalize_report_severity and scripts.resolve_config.one_line.
 It has no CLI and emits no receipt.
 """
 
 from collections.abc import Mapping
 
-# Keep this hand-written JS trim alphabet in sync with normalizeReportSeverity and
-# scripts/resolve_config.py::_JS_TRIM_RE; neither copy is generated.
+# One hand-written alphabet mirrors JS String.prototype.trim.
+# Shared by normalize_report_severity and scripts/resolve_config.py::one_line.
```
leehopper, line 12: **💡 [LOW] Module docstring uses dotted-path reference style inconsistent with the repo's path::symbol convention, and with its own adjacent comment**

The new docstring line references the resolver symbol as "scripts.resolve_config.one_line" (Python dotted-import style), while the comment two lines below it in the same file, added in the same diff, references the identical symbol as "scripts/resolve_config.py::one_line" (the file-path::symbol format used pervasively elsewhere in this repo, including the very same docstring's first sentence referencing renderReport.js::normalizeReportSeverity).

**Suggested fix:**
Use the same path::symbol reference form in the docstring as in the adjacent comment, e.g. "used by normalize_report_severity and scripts/resolve_config.py::one_line."

**Inconsistency:**
> Self-inconsistency within scripts/report_severity.py: line 5 says "scripts.resolve_config.one_line" while line 12 (added in the same diff) says "scripts/resolve_config.py::one_line" for the same symbol.

```suggestion
"""Shared report severity normalization and JavaScript-compatible trim alphabet.

This stdlib-only library mirrors renderReport.js::normalizeReportSeverity.
JS_TRIM_CHARS is the single shipped Python twin of JS String.prototype.trim,
used by normalize_report_severity and scripts/resolve_config.py::one_line.
It has no CLI and emits no receipt.
"""
```

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet

The PR (46 changed lines) claims to consolidate the duplicated JS-trim alphabet that previously lived as two hand-written copies — `JS_TRIM_CHARS` in scripts/report_severity.py and a derived `_JS_TRIM_RE` regex in scripts/resolve_config.py — into a single shipped Python source: resolve_config.py now inserts scripts/ onto sys.path and imports `JS_TRIM_CHARS` from report_severity, and its `one_line` function calls `text.strip(JS_TRIM_CHARS)` instead of applying the local regex, with report_severity.py's docstring and inline comment updated to describe it as the shared alphabet used by both `normalize_report_severity` and `resolve_config.one_line`. The stated motivation is to remove a documented cross-file duplication (tracked in docs/duplication-register.md, which is updated to reflect the new single-source relationship and its consumers) per the "build the mechanism, not the instruction" design rule. tests/test_resolve_config.py adds a test asserting `resolver.JS_TRIM_CHARS is report_severity.JS_TRIM_CHARS` by identity and exercising `mock.patch.object(resolver, "JS_TRIM_CHARS", ...)` to confirm the alphabet is read at call time rather than baked into a compiled regex; tests/test_generate_contract_requirements.py separately changes a contract-fixture-staging test to copy files from `set(gen.compute_targets(...).keys()) | set(gen.declared_inputs(...))` instead of a hand-maintained list of workflows/src paths, which is a change to fixture-staging completeness discovery. Risk classification marks scripts/resolve_config.py high (it is the resolver consulted for delivery-affecting config and now carries a new intra-package import and sys.path mutation), scripts/report_severity.py medium (its docstring/comment semantics change and it becomes an explicit shared-import target), and the two test files plus the doc as low. The touched areas span a cross-runtime/cross-file duplication registry, a config resolver on the delivery path, and two test suites, so downstream reviewers should attend to import-time side effects (sys.path insertion order relative to other imports), whether `declared_inputs` is guaranteed to enumerate the same file set the removed hardcoded list covered, and whether the duplication-register update accurately reflects the new consumer relationship.

1 finding(s) after the gauntlet — 1 low. 1 routed as improvement suggestion(s).

---
Generated by code-gauntlet | Reviewed up to: f3cd7d32fee3ce1ce9db5fb85dccba01b7a9f701

scripts/report_severity.py
leehopper: Fixed in 7edca47: the docstring now names `scripts/resolve_config.py::one_line`, matching the adjacent comment.

PR #345, opened 2026-09-16
fix: reject escaped surrogate pairs and cite register symbols a test resolves (#329, #330)

### LINKED ISSUE #329: Duplication register cites line numbers that rot; the checksum row has five stale refs

### LINKED ISSUE #330: Verify decoder accepts an escaped surrogate pair the encoder never emits, aliasing one astral character

leehopper:
Two adjacent issues from #172 in one branch: the verify decoder now rejects an escaped surrogate pair the encoder never emits (#330), and the duplication register cites symbols a test resolves instead of line numbers (#329). Seven commits: one per issue, four review fix rounds, and one fix for the code-gauntlet finding.

## #330

`decode_inline_slice` now rejects two adjacent `%uXXXX` escapes that form a valid high/low surrogate pair, by name and in the same family as the lowercase-hex, safe-ASCII-escape and non-surrogate-`%u` rejections. `encodeSliceInline` never emits that spelling: a well-formed pair is encoded as its UTF-8 bytes and only a lone surrogate is escaped. The two spellings decoded to different Python strings that the value proof serialised identically, so the accepted alphabet was injective only as Python strings. It is now injective as JS values. This is a canonicality fix, not a live trust hole: the token proof from #172 already caught the alternative spelling.

Lone surrogates in every position keep round-tripping (high then high, low then high, high at the end, high then a safe character, high then a UTF-8 run), and the `lone_surrogates`, `surrogate_pair` and `astral` parity fixtures are unchanged. Tests: `test_rejects_adjacent_escaped_surrogate_pair_by_name`, `test_inline_surrogate_pair_accept_reject_matrix`, the rewritten `test_the_token_proof_separates_documents_the_value_proof_cannot` (asserts the rejection and keeps the record that the value proof cannot separate two hand-built documents), and `test_the_accepted_spelling_alphabet_is_injective`, which now probes the pair spelling for U+1F600 and U+10FFFF and compares decoded values as JS values.

Mutations shown red then green: A1 (delete the whole rejection branch) failed all four tests; A2 (reject every high surrogate) failed the matrix's accepted rows and the lone-surrogates parity round-trip. Four comments and the validation-pipeline reference no longer say the decoder accepts that spelling. `workflows/pipeline.js` is byte-identical after rebuild. `workflows/AGENTS.md` was deliberately left alone: its verify-boundary bullet states nothing false, and the injectivity contract is held by the decoder comment and the test.

## #329

The register's citation grammar is one paragraph: a citation is a tracked repo-relative path, a directory with a trailing `/`, a glob, `path::symbol` (nesting joined with `::`, as rows already used for `_fold_review_body`), `path::"node:test title"`, or `path#Heading text`. Line numbers, `...`, same-file shorthand and bare basenames are forbidden. `tests/test_docs_registry.py` enforces it: a shared table parser (the row-count test now uses it too), a classifier that decides which code spans are locations, an ordered definition-pattern resolver, a residual scan over prose and Markdown link targets outside the tables, and a failure format of register line, span and reason.

Every one of the 58 rows was rewritten from a per-row census of what each cited line range meant at its origin commit (212 citations, all resolved at HEAD). Rows that changed meaning on purpose:

- The checksum row cites `workflows/src/stages.js::fnv1a32`, `::DELTA_KEYS`, `scripts/assemble_artifacts.py::fnv1a32` and `scripts/verify_findings.py::_DELTA_FIELDS`, with `workflows/AGENTS.md#The verify boundary` as its doc ref.
- The M4 counting-asymmetry row moved from the bare-URL pattern (dropped by #252) to `_INJECTION_ENCODED_PATTERNS` / `INJECTION_ENCODED_PATTERNS`, where the same UTF-16-unit versus codepoint asymmetry survives in the `[^\x00]{0,40}` gaps (probe at HEAD: 21 astral characters in the gap match in Python and not in JS).
- Same-file pairs (`candidates.json` records, `report.py::CSS` blocks, `test_check.py` methods) keep both sides distinct by nested symbol or by a parenthetical outside the code span.
- The hypothetical `conftest.py` mention became "a shared pytest `conftest` module", since the file does not exist.

Mutations shown red then green: B1 renamed symbol, B2 appended `:12`, B3 altered heading, B4 bare basename, B5 emptied `DEFINITION_PATTERNS` (every symbol and title citation reported), B6 grouped-table line spec and a glob with no match, C1/C2 classifier returns False/True, C3/C4 plain-text predicate returns False/True, C5 ordered cursor reset to zero.

## Review

Plan drafted by Codex Astra from the orchestrator's rulings, red-teamed by one Codex Astra seat and one Claude Opus seat (every finding folded in; one Opus blocker refuted, since a `😀` escape literal stays two code points in CPython). Built by Codex luna in two parallel stages. Whole-branch review: three Codex luna lenses (mutation ledger, plan conformance, docs) and one Claude Opus lens, over four rounds. The Opus lens caught what the luna lenses normalised: a row with bare symbols beside file citations, a `REVIEW.md` referent reworded away, and two mechanisms (the plain-text predicate and the resolver's ordered cursor) that survived a whole-mechanism mutation; all fixed and pinned in the second fix round, and re-verified by the same lens. The luna conformance lens then caught a method citation without its class and dead bookkeeping in the test; the last round restored prose wording the earlier over-broad scan had forced. Final verdicts: mutation accept, docs accept, conformance accept, Opus accept. Lee's own code-gauntlet run found one LOW (the resolver's rejection branches had no negative cases) with a corroborating simplification; f4ca056 adds one case per rejection reason (each branch mutated to `None` goes red) and removes `_path_reason`'s unused `extensions` parameter.

## Gates

Measured by the orchestrator on 954ced5 (f4ca056 changes one test file; its full Python suite passed again) under Node 24.21.0 and Python 3.12.13:

- `pytest tests/`: 2267 passed, coverage 94.29% (floor 93.8).
- `pytest bench/tests/`: 657 passed (floor 88.5 met).
- `node --test workflows/test/*.test.js`: 1400 passed; coverage lines 99.10 / branches 90.91 / functions 98.63 (floors 98.2 / 90.0 / 97.7).
- `node workflows/build.js`: `workflows/pipeline.js` byte-identical.
- `pre-commit run --all-files`: every hook passed.

Closes #329, closes #330

https://claude.ai/code/session_015272KkidHPufJaTLTj7c4F


tests/test_docs_registry.py
```diff
@@ -86,46 +87,351 @@ def tracked(pathspec):
...
+    if any(char in path for char in "*?["):
+        if not any(fnmatch.fnmatchcase(candidate, path) for candidate in tracked_files):
+            return "glob matches no tracked files"
+        return None
+    if path in tracked_files:
+        return None
+    return "untracked path"
+
+
+def _definition_reason(path, symbols, tracked_files):
```
leehopper, line 295: **💡 [LOW] Negative-path branches of the new citation validator are never asserted to reject bad input**

The new _citation_reason/_definition_reason/_title_reason resolver added in this diff has several rejection branches (heading not found, title not found, invalid symbol chain, no definition patterns for a file extension, dot/empty path component forbidden, glob matches no tracked files, invalid citation syntax) that no test exercises with a case expected to produce that specific reason, so a bug that makes any of these checks too permissive would silently let a broken documentation citation pass validation.

**Suggested fix:**
Add cases to malformed_citations (or a new table) that exercise each currently-unverified reason, e.g. a heading fragment that does not exist in a real .md file ('CLAUDE.md#Nonexistent Heading'), a JS test-title citation pointing at a title string absent from the file, a symbol containing an invalid character ('scripts/verify_findings.py::bad symbol'), a citation into a file extension with no DEFINITION_PATTERNS entry ('README.md::Foo' if .md has no python/js patterns), and a path containing './' or an empty segment ('scripts/./verify_findings.py', 'scripts//verify_findings.py').

---

**Corroborating finding — code-simplifier (simplification, confidence 88):**

**Same tracked-file-extensions set is rebuilt inline on every _path_reason call**

The expression `{Path(p).suffix for p in tracked_files if Path(p).suffix}` is written out identically four times (in _definition_reason, _title_reason, and twice in _citation_reason) purely to satisfy _path_reason's third parameter, even though the caller in test_duplication_register_citations_resolve already computes the same set once as `extensions` at line 619 and just doesn't thread it through. Each call rebuilds the set from the full tracked_files tuple, and a reader has to notice all four spots compute the same thing to realize it's not meaningfully different per call site.

⚔️ *Code Gauntlet*

leehopper (commented):
### ⚔️ Code Gauntlet

This PR (861 changed lines total) claims two entangled changes: (1) it closes a decoding ambiguity in the verify boundary's inline-slice decoder, and (2) it rewrites the duplication register's citation format from fragile line-number ranges to durable, resolvable citations, adding machinery to enforce that format. In scripts/verify_findings.py, `_decode_inline_string` now rejects an adjacent escaped high/low UTF-16 surrogate pair (an astral character re-spelled as `%uD83D%uDE00`-style escapes) as "non-canonical" instead of silently decoding it, closing a case where the token proof and value proof could describe the same decoded JS value two ways; workflows/src/stages.js and skills/code-gauntlet/references/validation-pipeline.md receive matching comment/prose updates describing this new rejection (no logic changes visible in stages.js beyond comments). tests/test_verify_findings.py updates the existing test that documented the old decode-then-distinguish behavior into one asserting the new rejection, adds `test_rejects_adjacent_escaped_surrogate_pair_by_name` and an accept/reject matrix test, and changes `test_the_accepted_spelling_alphabet_is_injective` to compare decoded results by UTF-16/JS-value equality rather than raw Python string equality, adding a 0x10FFFF probe. Separately, docs/duplication-register.md replaces essentially every row's location citations (line-range spans like `file.py:12-34`) with `path::symbol`, `path#Heading`, or `path::"quoted title"` forms and adds a paragraph defining this citation grammar, while tests/test_docs_registry.py adds substantial new parsing/validation infrastructure (`_register_sections`, `_split_table_cells`, `_code_spans`, `_is_location_span`, `_looks_like_location`, `_parse_citation`, `_path_reason`, `_definition_reason`, `_title_reason`, `_citation_reason`, `_tracked_files`) that resolves every citation in the register — in tables and prose — against git-tracked files, symbols, headings, and test titles, plus a new test exercising it. The stated motivation is closing a spelling gap between two independent JS-value-equality proofs in the security-sensitive verify boundary, and making the duplication register's location references resist drift from future renames/line-shifts by validating them structurally; risk concentrates in scripts/verify_findings.py (security-sensitive parsing, classified high risk) and the new citation-resolution logic in tests/test_docs_registry.py, which is large, regex-heavy, and governs whether documentation citations across the whole register are considered valid.</summary>


1 reported issue(s) from 2 finding(s) after the gauntlet — 1 low. 1 routed as improvement suggestion(s).

---
Generated by code-gauntlet | Reviewed up to: 954ced5fcf6b485059ee7aa686161260a4218512

tests/test_docs_registry.py
leehopper: Fixed in f4ca056. The malformed-citation table now has one case per rejection reason in the resolver, 18 new cases in all. I mutated each rejection return to `None` one at a time, and every mutation turned `test_duplication_register_citations_resolve` red. The simplification went one step further: `_path_reason` never read its `extensions` argument, so I removed the parameter and the four rebuilt sets.

leehopper:

